PR #12, opened 2014-12-11
Changed kaazing.org references to links

robinzimmermann:


robinzimmermann:
@vjwang made these changes independently, so I'm retracting this pull request.

PR #38, opened 2015-03-05
Set version at develop-SNAPSHOT.

dpwspoon:
Simplified versions such that versions are only SNAPSHOT or non-SNAPSHOT (no more beta).  Eliminated gateway.distribution.version by having the manifest set in gateway.server and picked up from gateway.service.update.check (This may eventually be better off in gateway.service.update.check all together).


dpwspoon:
One of 2 Travis builds failed.  The failed test intermittently fails and is already assigned to @krismcqueen

gateway.distribution.version/pom.xml
```diff
@@ -1,85 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
krismcqueen: Does removing this affect what the management console displays for the Gateway version?

PR #39, opened 2015-03-05
Update to latest robot to avoid race conditions in integration-tests

krismcqueen:
Updated to 2.0.0-alpha-8 of maven-k3po-plugin, modified some tests to use the new start/finish semantics in K3PO, fixed up a race condition and another WSEB failure.


pom.xml
```diff
@@ -83,6 +94,22 @@
                     <matchWithExisting>false</matchWithExisting>
                 </configuration>
             </plugin>
+            <plugin>
+              <groupId>org.kaazing</groupId>
+              <artifactId>k3po-maven-plugin</artifactId>
```
dpwspoon: What about alpha-9?

PR #66, opened 2015-03-23
Minor clean up work renaming of classes and updating/adding JavaDoc

dpwspoon:
I was trying to find the CLI and it took me a little bit to end up at the CommandProcessors so I renamed it.  If your in opinion that it shouldn't change I'm alright with that.

Changed Gateway Listener to Observer, as I felt it more accurately reflected observing changes in the gateway.  Added Javadoc to the Observer


server/src/main/java/org/kaazing/gateway/server/GatewayCLI.java
```diff
@@ -32,20 +32,18 @@
...
- * value of <configFile> is considered as a path to the gateway configuration file--NOT just a file name. If provided by the
- * caller, the property is converted to a system property and passed to Gateway, rather than requiring Gateway to handle a new
- * input vector.
+ * Supported command line arguments are --config <configFile> and --help. The value of <configFile> is considered as a
+ * path to the gateway configuration file--NOT just a file name. If provided by the caller, the property is converted to
+ * a system property and passed to Gateway, rather than requiring Gateway to handle a new input vector.
  */
-public class GatewayCommandProcessor {
-    private static final Logger LOGGER = LoggerFactory.getLogger(GatewayCommandProcessor.class);
+public class GatewayCLI {
```
krismcqueen: Could this be GatewayCommandLineProcessor or something similar that isn't an abbreviation?  Everything else looks good.

server/src/main/java/org/kaazing/gateway/server/WindowsMain.java
```diff
@@ -64,7 +61,7 @@ public static void main(String... args) throws Exception {
             args[i] = arg;
         }
 
-        GatewayCommandProcessor commandProcessor = new GatewayCommandProcessor(formatter);
+        GatewayCLI commandProcessor = new GatewayCLI(formatter);
```
krismcqueen: maybe rename the variable too based on the final class name

server/src/main/java/org/kaazing/gateway/server/GatewayCLI.java
dpwspoon: Sounds good,  Done!

server/src/main/java/org/kaazing/gateway/server/WindowsMain.java
dpwspoon: Sounds good,  Done!

dpwspoon:
Thanks @krismcqueen

PR #68, opened 2015-03-24
Misc minor code changes

jitsni:


dpwspoon:
Maybe we shouldn't have coverage/coverails be so strict.  I can turn off the coveralls fail if test coverage decreases? 

At the very least I don't think we should hold back this pull request because of this coveralls failure.

PR #77, opened 2015-03-26
Created BOM, refactor folder structure

dpwspoon:


pom.xml
```diff
@@ -35,52 +34,53 @@
...
-        <module>transport.http</module>
-        <module>transport.nio</module>
-        <module>transport.pipe</module>
-        <module>transport.sse</module>
-        <module>transport.ssl</module>
-        <module>transport.ws</module>
-        <module>transport.wseb</module>
-        <module>transport.wsn</module>
-        <module>transport.wsr</module>
+        <module>service/spi</module>
```
jitsni: I am ok with this way of organizing things under transport, service, resource.address. Other way is organize under protocols (for e.g: http, http/resource.address, http/transport, http/service.http.balancer, http/service.http.proxy etc). I wonder whether we thought about that one.

service/http.proxy/pom.xml
```diff
@@ -7,6 +6,7 @@
         <groupId>org.kaazing</groupId>
         <artifactId>gateway</artifactId>
         <version>develop-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
```
jitsni: So, it doesn't reference parent, i.e. service/pom.xml ? Do we even have that one ?

jitsni:
travisci build is failing with 
testTaskIsStartedWithProperConfiguration(org.kaazing.gateway.service.update.check.UpdateCheckServiceTest)  Time elapsed: 0.005 sec  <<< FAILURE!
java.lang.AssertionError: There is a jar in ./src/test/resources/gateway.server-5.0.0.8.jar that needs to be on the class path, fix your test runner settings

dpwspoon:
The jar got excluded because of the .gitignore and a file move.  I've added it back

pom.xml
dpwspoon: I'm up for refactoring it latter

PR #79, opened 2015-03-31
Add some more dependencies to the bom so it includes transitive org.kaaz...

cmebarrow:
...ing dependencies from distribution (mina.netty, etc).
This is needed so I can progress with work on transport.apns and jms.server


cmebarrow:
Closed because I will create a new one based on a branch with additional changes to remove all duplication of version numbers across all poms and include everything in the assembly (generic-bin.xml).

PR #88, opened 2015-04-08
Http.proxy.persistence

jitsni:


jitsni:
@jfallows, @cmebarrow, @krismcqueen, @dpwspoon ,
Please review.

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultConnectOptionsContext.java
```diff
@@ -114,6 +121,19 @@ public DefaultConnectOptionsContext(ServiceConnectOptionsType connectOptions, Se
...
+            if (timeoutValue != null) {
+                long val = Utils.parseTimeInterval(timeoutValue, TimeUnit.SECONDS);
+                if (val > 0) {
+                    tmpHttpKeepaliveTimeout = val;
+                }
+            }
+
         }
 
         // Set default values via service-defaults
```
dpwspoon: I think you are missing the value if it is set in the service defaults and not in a service's connect options

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultConnectOptionsContext.java
```diff
@@ -222,6 +242,10 @@ public DefaultConnectOptionsContext(ServiceConnectOptionsType connectOptions, Se
         }
 
         this.sslEncryptionEnabled = (sslEncryptionEnabled == null) ? true : sslEncryptionEnabled;
```
dpwspoon: Perhaps you could set the default value on the field at the top of the file and then override the value as it is set, thereby no longer needing the tmp variables.  I get whats going on here, but its a slightly different pattern then what other properties are doing in this class, see `sslEncryptionEnabled = true; // default to true`.  Also @krismcqueen is reworking this class in general so you might want to sync with him

service/http.proxy/src/test/java/org/kaazing/gateway/service/http/proxy/HttpProxyPersistenceIT.java
```diff
@@ -0,0 +1,79 @@
...
+    private final GatewayRule gateway = new GatewayRule() {
+        {
+            // @formatter:off
+            GatewayConfiguration configuration =
+                    new GatewayConfigurationBuilder()
+                        .service()
+                            .accept(URI.create("http://localhost:8110"))
+                            .connect(URI.create("http://localhost:8080"))
+                            .type("http.proxy")
+                            // NOTE: even though in the config file it's http.keepalive.timeout
```
dpwspoon: Why the difference?

service/http.proxy/src/test/scripts/org/kaazing/gateway/service/http/proxy/http.proxy.connect.persistence.rpt
```diff
@@ -0,0 +1,123 @@
...
+read "Content-Type: text/html;charset=UTF-8\r\n"
+read /Date: .*/ "\r\n"
+read "Server: Kaazing Gateway\r\n"
+read "Transfer-Encoding: chunked\r\n"
+read "\r\n"
+read "14\r\n"
+read "<html>Helloo1</html>\r\n"
+read "0\r\n"
+read "\r\n"
+
```
dpwspoon: This barrier is implicitly there, writes after reads will always wait for the reads to complete

service/http.proxy/src/test/scripts/org/kaazing/gateway/service/http/proxy/http.proxy.connect.persistence.rpt
```diff
@@ -0,0 +1,123 @@
...
+write "Transfer-Encoding: chunked\r\n"
+write "Date: Tue, 10 Feb 2015 02:17:15 GMT\r\n"
+write "\r\n"
+write "14\r\n"
+write "<html>Helloo2</html>\r\n"
+write "0\r\n"
+write "\r\n"
+
+close
+closed
```
dpwspoon: I believe we've been trying to put newlines at end of files

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultConnectOptionsContext.java
jitsni: I am using tmp variable so that the class fields can be final. I don't like fields getting changed multiple times and that's what happening to other fields. I will wait for kris's changes.

service/http.proxy/src/test/java/org/kaazing/gateway/service/http/proxy/HttpProxyPersistenceIT.java
jitsni: I thought builder invoking xmlbeans use camel case of notation methods and doesn't deal with the dotted config options . It is used like that ws.inactivity.timeout (for e.g. see WsebInactivityTimeoutIT.java). But I really prefer to use http.keepalive.timeout if possible.

service/http.proxy/src/test/scripts/org/kaazing/gateway/service/http/proxy/http.proxy.connect.persistence.rpt
jitsni: OK, will take care of that.

transport/spi/src/main/java/org/kaazing/gateway/transport/BridgeService.java
```diff
@@ -23,7 +23,8 @@
 
 
 public interface BridgeService {
-    
+
     void dispose();
-    
+
```
jitsni: I fixed this formatting.

service/http.proxy/src/test/java/org/kaazing/gateway/service/http/proxy/HttpProxyPersistenceIT.java
dpwspoon: oh.okay.  Seems like a builder issue, perhaps we should file an issue in Gateway to fix that 

Ping @sbadugu

krismcqueen: Don't bother filing another issue since I'm changing some of this with the SOCKS code...

service/http.proxy/src/test/scripts/org/kaazing/gateway/service/http/proxy/http.proxy.connect.persistence.rpt
jitsni: Removed the barrier

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultConnectOptionsContext.java
jitsni: The schema says "This is the element for configuring a set of default accept-options for all services running on the gateway." Are there service-defaults for connect-options also ?

jitsni: Added code and a test for keep alive connect option in <service-defaults>

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultConnectOptionsContext.java
```diff
@@ -40,10 +40,12 @@
 import org.kaazing.gateway.util.Utils;
 import org.kaazing.gateway.util.ssl.SslCipherSuites;
 import org.kaazing.gateway.util.ws.WebSocketWireProtocol;
+import static org.kaazing.gateway.service.TransportOptionNames.*;
 
 public class DefaultConnectOptionsContext implements ConnectOptionsContext {
 
     private static final long DEFAULT_WS_INACTIVITY_TIMEOUT_MILLIS = 0L;
+    private static int DEFAULT_HTTP_KEEPALIVE_TIMEOUT = 30; //seconds
```
cmebarrow: Are we changing the externally perceived behavior of the product by altering the default value of one of the connect options?

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultConnectOptionsContext.java
cmebarrow: David recently made a change to allow service-defaults to include connect-options
I do see why you are using the tmpXxx variables: to allow the members to be final.

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultConnectOptionsContext.java
cmebarrow: Yeah, my vote if for final too. Might be worth changing the others to final to for consistency

server/src/main/xsd/gateway-config-201409.xsd
```diff
@@ -1225,6 +1229,18 @@
             <element name="ws.inactivity.timeout" maxOccurs="1" minOccurs="0" type="gateway:TimeIntervalString">
                 <annotation/>
             </element>
+            <element name="http.keepalive" maxOccurs="1" minOccurs="0">
```
cmebarrow: Default is enabled, right? I expect you are already on to this but please ensure this gets documented

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultConnectOptionsContext.java
jitsni: I forgot the final keyword, fixing it. The default is 30 secs (or could be lower) and can be configured via connect/accept option. We already have this option on the accept side. I see that some products (like nginx etc) offer it for configuration. But some (like JDK) don't offer it and manage behind the scenes.

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultConnectOptionsContext.java
jitsni: Yes, I took care of it. Also, updated the ServiceDefaultsTest

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultConnectOptionsContext.java
jitsni: Kris has some changes here. Will do it later.

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpConnector.java
```diff
@@ -140,9 +161,88 @@ protected boolean canConnect(String transportName) {
...
+        final ResourceAddress transportAddress = address.getTransport();
+
+        if (transportAddress != null) {
+            Executor ioExecutor = null;
+            boolean ioAligned = isIoAligned();
+            if (!ioAligned) {
+                ioExecutor = getIoExecutor();
+            }
+
+            if (ioAligned || ioExecutor == null) {
```
cmebarrow: How could ioExecutor be null? Wouldn't that be a serious issue?

jitsni: There are two cases:
1) acceptor+proxy+connector. In this case, the executor won't be null. This applies to http proxy 
2) connector. In this case, the executor can be null and will get associated with an executor later.

John want both the cases to work.

server/src/main/xsd/gateway-config-201409.xsd
jitsni: Updated schema with docmentation for http.keepalive, http.keepalive.timeout. Will check with doc team.

dpwspoon:
LGTM

PR #89, opened 2015-04-08
Added test for service defaults getting parsed from xml, and generified ...

dpwspoon:
...test so it runs on java 8


krismcqueen:
Looks good!  Thanks for adding the parsing sanity check.

dpwspoon:
I'm not sure why travis doesn't show green.  Both logs show that they built successfully

PR #90, opened 2015-04-08
barriers are not required in these scripts

jitsni:


dpwspoon:
LGTM, will merge when travis completes

PR #94, opened 2015-04-14
Update Accept/Connect Options to be generic such that they can be extended

krismcqueen:
The Accept and Connect Options are updated to be generic bags of properties.  This allows them configuration to be extended without the need to update the schema for the new configuration elements.

The configuration builder is changed to map accept/connect options more like the XSD used to, and then the AcceptOptionsContext and ConnectOptionsContext exposes an asOptionsMap() to apply those options to the ResourceAddresses.


krismcqueen:
ping @jitsni @dpwspoon 

If you guys can both take a look that would be great.

management/src/main/java/org/kaazing/gateway/management/config/ServiceConfigurationBeanImpl.java
```diff
@@ -150,39 +156,39 @@ public String getAcceptOptions() {
 //                  }
 
 
-                    jsonOptions.put("ws.maximum.message.size", context.getWsMaxMessageSize());
+                    jsonOptions.put("ws.maximum.message.size", (Long) acceptOptions.get("ws.maximum.message.size"));
```
jitsni: Long cast is not required.

management/src/main/java/org/kaazing/gateway/management/config/ServiceConfigurationBeanImpl.java
```diff
@@ -216,14 +222,16 @@ public String getConnects() {
     public String getConnectOptions() {
         if (serviceContext.supportsConnects()) {
             ConnectOptionsContext context = serviceContext.getConnectOptionsContext();
+            Map<String, Object> connectOptions = context.asOptionsMap();
 
             JSONObject jsonOptions = new JSONObject();
 
             try {
                 if (context != null) {
```
jitsni: context will not be null at this line (as it already dereferenced at line 225), so you could eliminate null check

management/src/main/java/org/kaazing/gateway/management/config/ServiceDefaultsConfigurationBeanImpl.java
```diff
@@ -62,6 +64,7 @@ public String getAcceptOptions() {
         JSONObject jsonObj;
 
         AcceptOptionsContext context = serviceDefaultsContext.getAcceptOptionsContext();
+        Map<String, Object> acceptOptions = context.asOptionsMap();
         try {
             if (context != null) {
```
jitsni: same comment, context cannot be null at this line.

server/src/main/java/org/kaazing/gateway/server/config/parse/translate/March2015Translator.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.kaazing.gateway.server.config.parse.GatewayConfigNamespace;
+import org.kaazing.gateway.server.config.parse.translate.mar2015.ReorderAcceptAndConnectOptionsVisitor;
+
+public class March2015Translator extends GatewayConfigTranslatorPipeline {
+    public March2015Translator() {
+        super();
+
+        // Make sure accept-options and connect-options are properly ordered
+        addTranslator(new ReorderAcceptAndConnectOptionsVisitor());
```
jitsni: Why do we need the options to be reordered in DOM ?

server/src/main/java/org/kaazing/gateway/server/config/parse/translate/NamespaceVisitor.java
```diff
@@ -0,0 +1,72 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * Set the namespace on the given DOM to be the given {@ GatewayConfigNamespace} .
+ */
+public class NamespaceVisitor extends AbstractVisitor {
+
+    private static final Logger logger = LoggerFactory.getLogger(NamespaceVisitor.class);
+
+    private GatewayConfigNamespace ns;
```
jitsni: Seems like final field

server/src/main/java/org/kaazing/gateway/server/config/parse/translate/NamespaceVisitor.java
```diff
@@ -0,0 +1,72 @@
...
+
+import org.jdom.Document;
+import org.jdom.Element;
+import org.jdom.Namespace;
+import org.kaazing.gateway.server.config.parse.GatewayConfigNamespace;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Set the namespace on the given DOM to be the given {@ GatewayConfigNamespace} .
```
jitsni: {@link GatewayConfigNamespace}

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultAcceptOptionsContext.java
```diff
@@ -73,356 +81,59 @@
...
-    private final List<String> wsProtocols;
-    private final List<String> wsExtensions;
-    private final long tcpMaximumOutboundRate;
-    private final String udpInterface;
-    private final URI pipeTransportURI;
+    private Map<String, String> options;
 
     public DefaultAcceptOptionsContext() {
-        this(ServiceAcceptOptionsType.Factory.newInstance(), ServiceAcceptOptionsType.Factory.newInstance());
+        this.binds = new HashMap<String, String>();
```
jitsni: I would just use <> instead of <String, String>

server/src/main/java/org/kaazing/gateway/server/test/Gateway.java
```diff
@@ -267,20 +262,14 @@ private void setServiceDefaults(GatewayConfig gatewayConfig,
...
-            String acceptOptionValue = configuredAcceptOptions.get(acceptOptionName);
-            if (acceptOptionValue != null) {
-                setAcceptOption(newAcceptOptions, acceptOptionPropertyInfo, acceptOptionValue);
-            }
+        Node domNode = newAcceptOptions.getDomNode();
+        Document ownerDocument = domNode.getOwnerDocument();
+        for (Entry<String, String> acceptOption : configuredAcceptOptions.entrySet()) {
+            Element newElement = ownerDocument
+                    .createElementNS(domNode.getNamespaceURI(), acceptOption.getKey());
+            Text newTextNode = ownerDocument.createTextNode((String) acceptOption.getValue());
```
jitsni: String cast is not required (also other places in this file)

server/src/test/java/org/kaazing/gateway/server/config/parse/GatewayConfigParserTest.java
```diff
@@ -229,7 +232,7 @@ public void testValidWsInactivityTimeout() throws Exception {
         }
     }
 
-    @Test(expected = GatewayConfigParserException.class)
+    @Test(expected = GatewayConfigParserException.class) @Ignore
```
jitsni: Do you really want to ignore this one and the one below ?

server/src/test/java/org/kaazing/gateway/server/config/parse/ServiceDefaultsTest.java
```diff
@@ -26,7 +30,7 @@ public void testDefaultConnectOptions() throws Exception {
                 .connectOption("udp.interface", "en0")
                 .connectOption("tcp.transport", "socks://localhost:8000")
                 .connectOption("http.keepalive", "disabled")
-                .connectOption("http.keepaliveTimeout", "5sec")
+                .connectOption("http.keepalive.timeout", "5sec")
```
jitsni: I really like this (keeping the same config option)

server/src/test/java/org/kaazing/gateway/server/context/resolve/AcceptOptionsTest.java
```diff
@@ -74,13 +77,13 @@ public void setAllowedServices() {
         serviceList.add("directory");
     }
 
-    @Test
+    @Test @Ignore
```
jitsni: Do you want to ignore these tests in this file ? or just forgot to remove the annotation ?

service/spi/src/main/java/org/kaazing/gateway/service/ConnectOptionsContext.java
```diff
@@ -21,26 +21,22 @@
...
+     * Set the map of options on the ConnectOptionsContext.  The options are String key/value pairs.
+     * @param options the map of String key/value pairs to set on the ConnectOptionsContext
+     */
+    void setOptions(Map<String, String> options);
 
+    /**
+     * Set the map of default options on the ConnectOptionsContext.  The options are String key/value pairs.
+     * If the given default option does not exist in the option map it is added.  Any options already set
+     * are left untouched.
+     * @param options the map of String key/value pairs to set as the default options on the ConnectOptionsContext
```
jitsni: the parameter is defaultOptions.

service/spi/src/main/java/org/kaazing/gateway/service/AcceptOptionsContext.java
```diff
@@ -72,29 +52,18 @@
...
-    long getTcpMaximumOutboundRate();
+    void setOptions(Map<String, String> options);
 
     /**
-     * @return The configured idle timeout for the given protocol scheme, or null if there is none
-     *         configured and there is no default for the scheme
+     * Set the map of default options on the AcceptOptionsContext.  The options are String key/value pairs.
+     * If the given default option does not exist in the option map it is added.  Any options already set
+     * are left untouched.
+     * @param options the map of String key/value pairs to set as the default options on the AcceptOptionsContext
```
jitsni: the parameter is defaultOptions

service/http.proxy/src/test/java/org/kaazing/gateway/service/http/proxy/HttpProxyPersistenceIT.java
```diff
@@ -54,7 +54,7 @@
                             .connect(URI.create("http://localhost:8080"))
                             .type("http.proxy")
                             // NOTE: even though in the config file it's http.keepalive.timeout
```
jitsni: Just remove the comment

service/echo/src/main/java/org/kaazing/gateway/service/echo/EchoService.java
```diff
@@ -61,25 +61,7 @@ public void init(ServiceContext serviceContext) throws Exception {
...
-                } else {
-                    result = new ArrayList<>(supportedProtocols);
-                    result.addAll(echoProtocols);
-                }
-
-                return result;
-            }
-            
-        });
+        serviceContext.bind(serviceContext.getAccepts(), handler, new AcceptOptionsContext.Wrapper(acceptOptions) { });
```
jitsni: I think you can use two argument bind() method like other services.

server/src/main/java/org/kaazing/gateway/server/config/parse/translate/GatewayConfigTranslatorFactory.java
```diff
@@ -57,9 +57,29 @@ public GatewayConfigTranslator getTranslator(GatewayConfigNamespace ns) {
...
+                //
+                // This means that the switch statement MUST have the
+                // namespaces labels in chronological order, oldest to newest.
+
+            case MARCH_2015:
+                // The March 2015 schema enforces an order of elements in the accept-options
+                // and connect-options.  Make sure that this translator is always run so that
+                // config files built with this schema or any earlier schema do not require
+                // a specific order, but instead the elements are correctly ordered at runtime
+                // before the parsing/validation.
```
jitsni: That explains my reordering question. But why does mar 2015 enforce order ??

server/src/main/xsd/gateway-config-201503.xsd
```diff
@@ -0,0 +1,1288 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dpwspoon: Is a new namespace required?  It seems all old namespace configs would parse in this one.  

If we do update the namespace does this affect the next version we release the community gateway at.

server/src/test/java/org/kaazing/gateway/server/config/parse/ServiceDefaultsTest.java
dpwspoon: +1

.gitignore
```diff
@@ -17,6 +17,7 @@ out
 build-local.properties
 .gradle
 build
```
dpwspoon: What is the .checkstyle?  Eclipse adds this?

dpwspoon:
LGTM other then whats noted by jitsni and the question on the namespace

server/src/test/java/org/kaazing/gateway/server/config/parse/GatewayConfigParserTest.java
krismcqueen: I updated the Ignore annotation to give a reason why the test is no longer valid.  This is a candidate for removal altogether, but I didn't want to do that yet since maybe a better way of specifying the options that allow for XSD validation would be suggested.

.gitignore
krismcqueen: Possibly, not sure.  It was in every directory though so I added it to .gitignore.

krismcqueen:
@dpwspoon @jitsni   I incorporated some feedback.  Additional comments would be appreciated.  Thanks!

server/src/main/java/org/kaazing/gateway/server/config/parse/translate/GatewayConfigTranslatorFactory.java
```diff
@@ -57,7 +57,7 @@ public GatewayConfigTranslator getTranslator(GatewayConfigNamespace ns) {
         switch (ns) {
 
             case SEPTEMBER_2014:
-                // Currently no per-namespace translator to add in here, just validate
+                /// Currently no per-namespace translator to add in here, just validate
```
dpwspoon: minor but there is an extra "/"

dpwspoon:
LGTM

server/src/test/java/org/kaazing/gateway/server/config/parse/ServiceDefaultsTest.java
brennangaunce: See also kaazing/roadmap#180
(Story for standardizing names)

Is there Doc Impact here?

krismcqueen: The name wasn't changed, this is the configuration value that goes in the XML.  Based on how I changed things the ConfigBuilder uses those values that were in the XML rather than the camel case names.

I removed the strongly typed accept/connect options from the XSD, but didn't change any other behavior.  I was assuming an end user wouldn't notice and thus didn't think there was doc impact.  Correct me if I'm wrong.

PR #95, opened 2015-04-15
Changes for apns ws extension

cmebarrow:
Just a minor change to make WsExtensionValidation.STATUS public. This whole class will go away once we redo the WebSocket extension framework but for now it gets us out of trouble for the other work I am doing (jms service, which requires enabling the transport.ws.extension.apns extension in EE).

### DIFF
--- a/transport/ws/src/main/java/org/kaazing/gateway/transport/ws/extension/WsExtensionValidation.java
+++ b/transport/ws/src/main/java/org/kaazing/gateway/transport/ws/extension/WsExtensionValidation.java
@@ -23,7 +23,7 @@ package org.kaazing.gateway.transport.ws.extension;
 
 public class WsExtensionValidation {
 
-    static enum Status {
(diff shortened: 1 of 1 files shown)


transport/ws/src/main/java/org/kaazing/gateway/transport/ws/extension/WsExtensionValidation.java
```diff
@@ -23,7 +23,7 @@
 
 public class WsExtensionValidation {
```
dpwspoon: nite: isn't the static implicit/redundant.  See [here](http://stackoverflow.com/questions/23127926/static-enum-vs-non-static-enum)?

cmebarrow: Correct, fixed

PR #97, opened 2015-04-17
Fix how ws.inactivity.timeout was applied though accept-options

krismcqueen:
The configuration values are supposed to be in the dotted lowercase format.  Once applied as an optionsMap it should be turned into ws.inactivityTimeout.  However, pulled from the accept-options it should be ws.inactivity.timeout.  This is because the accept-options had the name in the XSD set to ws.inactivity.timeout, so this should always be the name in the accept-options.


krismcqueen:
ping @jitsni I had a bug in my earlier pull request and here is the fix

jitsni:
Looks fine. Will merge (or you can) after the travis build.

PR #105, opened 2015-04-22
Exposing accept-options and connect-options in the management layer...

krismcqueen:
...in a flexible manner to support extensibility.

### DIFF
--- a/management/src/main/java/org/kaazing/gateway/management/config/ServiceDefaultsConfigurationBeanImpl.java
+++ b/management/src/main/java/org/kaazing/gateway/management/config/ServiceDefaultsConfigurationBeanImpl.java
@@ -24,13 +24,17 @@ package org.kaazing.gateway.management.config;
 import static java.util.Arrays.asList;
 
 import java.net.URI;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.json.JSONObject;
 import org.kaazing.gateway.management.gateway.GatewayManagementBean;
 import org.kaazing.gateway.server.context.ServiceDefaultsContext;
 import org.kaazing.gateway.service.AcceptOptionsContext;
+import org.kaazing.gateway.service.ConnectOptionsContext;
 import org.kaazing.gateway.util.Utils;
 
 public class ServiceDefaultsConfigurationBeanImpl implements ServiceDefaultsConfigurationBean {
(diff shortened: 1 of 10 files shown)


krismcqueen:
@jitsni any feedback would be appreciated, especially in how to better manage the strings from the config as related to the strings in the options Map.

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultServiceContext.java
```diff
@@ -282,9 +287,18 @@ public boolean equals(Object otherObject) {
 
     @Override
     public int hashCode() {
-        int result = serviceType != null ? serviceType.hashCode() : 0;
-        result = 31 * result + (accepts != null ? accepts.hashCode() : 0);
-        return result;
+        if (hashCode == -1) {
+            int result = serviceType != null ? serviceType.hashCode() : 0;
+            result = 31 * result + (accepts != null ? accepts.hashCode() : 0);
```
jitsni: Objects.hashCode(accepts) should correspond to (accepts != null ? accepts.hashCode() : 0)
Perhaps even better, why cannot we use Objects.hash(serviceType, accepts, serviceName)

management/src/main/java/org/kaazing/gateway/management/config/ServiceDefaultsConfigurationBeanImpl.java
```diff
@@ -95,35 +99,120 @@ public String getAcceptOptions() {
                     jsonOptions.put("ssl.verify-client", "none");
                 }
 
-                jsonOptions.put("ws.maximum.message.size", acceptOptions.get("ws.maximum.message.size"));
+                jsonOptions.put("ws.maximum.message.size", acceptOptions.remove("ws.maxMessageSize"));
```
jitsni: ws.maximum.message.size seems to be expanded into ws.maxMessageSize, ws[ws/rfc6455].ws[ws/rfc6455].maxMessageSize, ws[ws/draft-7x].ws[ws/draft-7x].maxMessageSize in DefaultAcceptOptionsContext

Do we need to put all three values ? How do we handle such cases ?

jitsni:
@dpwspoon can you also review here (I don't have full understanding here)

management/src/main/java/org/kaazing/gateway/management/config/ServiceDefaultsConfigurationBeanImpl.java
krismcqueen: They are explicitly ignored before processing the rest of the options.  See the loop towards the bottom of the method.

krismcqueen:
@jitsni code has been updated as per your feedback.  Let me know if there is anything else.

PR #109, opened 2015-04-25
Let the origin server header flow through http.proxy service

jitsni:


dpwspoon:
LGTM.

service/http.proxy/src/test/java/org/kaazing/gateway/service/http/proxy/SecureOriginServer.java
```diff
@@ -58,11 +58,14 @@ public void run() {
...
-                } catch (Exception ioe) {
+                } catch (IOException ioe) {
                     ioe.printStackTrace();
+                    if (this.ioe != null) {
+                        this.ioe = ioe;
+                    }
                 }
             } catch (IOException ioe) {
-                this.ioe = ioe;
+                // no-op since socket.accept() may throw IOE
```
brennangaunce: Please log this exception at the trace level.  No exception should ever go uncaught.

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpAcceptProcessor.java
```diff
@@ -361,7 +361,7 @@ public static void setServerHeader(DefaultHttpSession httpSession, HttpResponseM
         if (httpSession != null) {
             ResourceAddress address = httpSession.getLocalAddress();
             boolean serverHeaderEnabled = address.getOption(HttpResourceAddress.SERVER_HEADER_ENABLED);
-            if (serverHeaderEnabled) {
+            if (serverHeaderEnabled && !httpResponse.hasHeader("Server")) {
```
brennangaunce: The code above appears to set the header if server header is enabled, but do not overwrite.  In addition, when the server header is disabled the Server header is passed through.

Should we have an enhancement to *remove* the passthrough Server header?  (Valid options might be: Enabled, Disabled, Remove, Overwrite)

service/http.proxy/src/test/java/org/kaazing/gateway/service/http/proxy/SecureOriginServer.java
jitsni: accept() will throw an exception when the socket is closed. So it is an expected exception.

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpAcceptProcessor.java
jitsni: Good point. I was thinking about it too. I thought there would be ways to configure at the origin server level. Moreover, I think more generic solution to configure any header.

service/http.proxy/src/test/java/org/kaazing/gateway/service/http/proxy/SecureOriginServer.java
brennangaunce: Please log it at TRACE level.  No exception should ever go uncaught.

PR #111, opened 2015-04-29
Add resource.address.httpxe test dependency to server\pom.xml since it i...

cmebarrow:
...s needed for the WebSocket transport. This allows other projects such as jms.server.itests to use the server pom.xml as a BOM without needing to add specific dependencies like this.

### DIFF
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -61,6 +61,12 @@
         </dependency>
         <dependency>
             <groupId>org.kaazing</groupId>
+            <artifactId>gateway.resource.address.httpxe</artifactId>
+            <version>${httpxe.address.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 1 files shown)


cmebarrow:
David please release Gateway after merging this (or I can do it if you tell me how). Thanks, chris

server/pom.xml
```diff
@@ -482,6 +488,7 @@
         <ssl.address.version>${project.version}</ssl.address.version>
```
jitsni: Not related this check-in, but I think we can get rid of all these properties as they are all reside in the same project and released together.

server/pom.xml
```diff
@@ -61,6 +61,12 @@
         </dependency>
         <dependency>
             <groupId>org.kaazing</groupId>
+            <artifactId>gateway.resource.address.httpxe</artifactId>
```
jitsni: looks good

PR #115, opened 2015-05-01
resolves #114 and better describes xsd

### LINKED ISSUE #114: Test is using ephemeral port to bind
BindingsTest. shouldBeAbleToBindAndUnBindDistinctAddressesAndHaveReferenceCountsCorrect

dpwspoon:


jitsni:
Travis ci failure will not be fixed by this one. It seems like the test created some duplicate URIs( since findFreePort() returns the same port as there is no physical bind in this test). The following supports it :

java.lang.RuntimeException: Tried to bind address [tcp://127.0.0.1:48645 (tcp://localhost:48645)]
[tcp://[0:0:0:0:0:0:0:1]:48645 (tcp://localhost:48645)] to a different handler (old,new)

I think we can change the test to maintain Set of ports and then create the URIs. That way there won't be any duplicate ports and hence there won't be any duplicate URIs

dpwspoon:
@jitsni What about just rolling through the ports, so they are deterministic?  See updates

jitsni:
That looks good. But I wonder whether the test needs to use findFreePort() at all since it doesn't physically bind. Just pick ports and use it.

dpwspoon:
K, I changed that

jitsni:
looks good. will wait for travis ci build

PR #116, opened 2015-05-01
Restore strongly typed accept-options / connect-options

krismcqueen:
This should give XSD validation.


krismcqueen:
@jitsni Any feedback you can provide would be appreciated.

PR #123, opened 2015-05-07
Allow a proxy.connector to be injected into the N…

krismcqueen:
…ioSocketConnector to handle cases were a proxy is configured in the JVM to handle certain connections.  See java.net.Proxy and java.net.ProxySelector.Also fix up some minor formatting.

### DIFF
--- a/transport/nio/src/main/java/org/kaazing/gateway/transport/nio/NioSocketConnector.java
+++ b/transport/nio/src/main/java/org/kaazing/gateway/transport/nio/NioSocketConnector.java
@@ -45,6 +45,7 @@ import org.jboss.netty.channel.socket.nio.NioWorker;
 import org.jboss.netty.channel.socket.nio.WorkerPool;
 import org.kaazing.gateway.resource.address.ResourceAddress;
 import org.kaazing.gateway.resource.address.ResourceAddressFactory;
+import org.kaazing.gateway.transport.BridgeConnector;
 import org.kaazing.gateway.transport.BridgeServiceFactory;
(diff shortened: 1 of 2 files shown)


krismcqueen:
@jitsni the CE part of the fix to enable proxy transport in EE...

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/NioSocketConnector.java
```diff
@@ -97,6 +104,12 @@ protected BridgeServiceFactory initBridgeServiceFactory() {
             ResourceAddress address, IoHandler handler,
             IoSessionInitializer<T> initializer) {
 
+        if (proxyConnector != null) {
+            ConnectFuture future = proxyConnector.connect(address, handler, initializer);
+            if (future != null) {
```
jitsni: I see that you moved the code to create proxy:// to proxy transport component. That seems to be ok but it means the Connector#connect is returning a null ConnectFuture. I don't think any Connector.connect returns null future. We don't have javadoc on the connect methods to explain the contract but my guess is that it returns non-null ConnectFuture object. This is something to think about it.

Other option is to create proxy address here.

jitsni:
Overall looks good.

brennangaunce:
Do we have a test that validates this feature?  Please include positive and negative tests with this sprint.

jitsni:
The ProxySelectorIT test in EE will verify this. In CE, this is just an extension point.

PR #131, opened 2015-05-14
Allow proxy service to be extended

krismcqueen:
Add a hook point for (a) proxy service extensions to be created and (b) the extension to be invoked when the full proxied connection is ready, meaning both accept side and connect side are established.


service/proxy/src/main/java/org/kaazing/gateway/service/proxy/ProxyServiceExtensionSpi.java
```diff
@@ -0,0 +1,38 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.kaazing.gateway.service.proxy;
+
+import org.kaazing.mina.core.session.IoSessionEx;
+
+public abstract class ProxyServiceExtensionSpi {
```
jitsni: If we want to have more flexibility it can be interface too (esp, java 8 has default methods)

jitsni:
Overall, looks good.

PR #135, opened 2015-05-16
Fixed an issue where gateway wasn't flushing PING frame followed by RECONNECT frame in long polling mode

pkhanal:


jitsni:
Looks good.

PR #137, opened 2015-05-16
Wseb transport session

cmebarrow:
This request contains the code changes for story [#309 wseb modifications needed for new WebSocket extensions SPI](/kaazing/roadmap/issues/309). The wseb transport is modified to provide a filter chain where all incoming and outgoing  WebSocket frames transit, so the are visible to any filters added to it (in particular, WebSocket extension filters, but also WsCheckAliveFilter for the ws inactivity timeout feature which is now placed onto that filter chain). This new filter chain is that of new IoSession "TransportSession" which is an inner class in WsebSession. There is one TransportSession for each WsebSession.
````
 codec   WsMessage                                                          IoBuffer
upstream ---------> (WsebUpstreamHandler)TransportSession(TransportHandler) --------> WsebSession
downstream <------ (WsebAcceptProcessor)TransportSession <--------- (WsebSessionProcessor)WsebSession
 codec                                                    WsMessage
````

* WsebAcceptProcessor and WsebConnectProcessor are now set on the TransportSession (via a wrapper: TransportProcessor, also an inner class in WsebSession).
* The processing previously done in flushInternal in WsebAcceptProcessor and WsebConnectProcessor is now split between those classes and WsebSessionProcessor (another inner class in WsebSession), which is set as the processor on the WsebSession. 
* WsebSessionProcessor converts outgoing IoBufferEx messages into WsMessages and writes them to the TransportSession. Then they are written to the downstream by WsebAcceptProcessor or WsebConnectProcessor.
* The wseb codec filter is still set on the upstream and downstream sessions (later we could put it on the TransportSession filter chain, but for now I want to limit the extent of the code changes).
* WsebUpstreamHandler is modified to pass incoming messages, as WsMessages, to fireMessageReceived on the filter chain of the TransportSession.TransportHandler (the handler set on TransportSession) then passes the WsMessage payload into the filter chain of WsebSession (for binary and text messages) and does appropriate handling for control frames.

There is also some testware cleanup: MethodExecutionTrace in the test.util module is now a TestRule so it can be used in rule chains. The default constructor automatically loads the log4j-trace.properties, which is provided on the class path by test.util. All of the duplicate copies of that file in various modules are therefore no longer needed and have been removed. Use of  log4j-trace.properties assures that trace level log messages will be printed for any tests which fail, within test clauses using MethodExecutionTrace.

### CODE REVIEW

**Comment by cmebarrow** [comment]
@jfallows Pull request for wseb transport session is now ready for review

File: transport/wseb/src/main/java/org/kaazing/gateway/transport/wseb/WsebAcceptProcessor.java [hunk 1]
```diff
@@ -179,35 +175,8 @@ protected void flushInternal(final WsebSession session) {
...
-                        }
-                        // flush the buffer out to the session
-                        lastWrite = flushNowInternal(writer, wsebMessage, wsBuffer, filterChain, request);
-                    }
-                    else {
-                        // flush the buffer out to the session
-                        lastWrite = flushNowInternal(writer, new WsBinaryMessage(buf), buf, filterChain, request);
-                    }
+                    // flush the message out to the session
+                    lastWrite = flushNowInternal(writer, frame, buf, filterChain, request);
```
jitsni: I like the new code (no special cases for ping/pong). I wonder if we can simplify further by calling checkBuffer() only once in that loop.

test.util/src/main/java/org/kaazing/test/util/MethodExecutionTrace.java
```diff
@@ -32,49 +36,55 @@
...
-            PropertyConfigurator.configure(log4jConfigPropertiesFileName);
+        if (log4jPropertiesResourceName != null) {
+            // Initialize log4j using a properties file available on the class path
+            Properties log4j = new Properties();
+            InputStream in = Thread.currentThread().getContextClassLoader().getResourceAsStream(log4jPropertiesResourceName);
+            if (in == null) {
+                throw new RuntimeException(String.format("Could not load resource %s", log4jPropertiesResourceName));
+            }
+            try {
+                log4j.load(in);
```
jitsni: in needs to be closed (seems like Properties.load() doesn't close it). Use ARM block

test.util/src/main/java/org/kaazing/test/util/MethodExecutionTrace.java
```diff
@@ -32,49 +36,55 @@
...
-        return currentFullTestName;
-    }
+public class MethodExecutionTrace extends TestWatcher {
 
     /**
      * Use this constructor to set up a particular log4j configuration using a properties file,
-     * e.g. to use MemoryAppender.
+     * or null if you do not want to load any log4j configuration properties.
      * See gateway.server src/test/resources/log4j-trace.properties for an example.
      * @param log4jConfigPropertiesFileName
```
jitsni: can you fix the param name

test.util/src/main/java/org/kaazing/test/util/MethodExecutionTrace.java
```diff
@@ -32,49 +36,55 @@
...
      */
-    public MethodExecutionTrace(String log4jConfigPropertiesFileName) {
+    public MethodExecutionTrace(String log4jPropertiesResourceName) {
         MemoryAppender.initialize();
```
dpwspoon: I never really liked the idea of having log4j properties files all over the place and I really like this change.  However, I was curious if you considered exposing the [programatic configuration](https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/BasicConfigurator.html) of the log4j.   I'm not familiar with the API or how easy it is to use but it MAY be easier to customize logging programmatically rather than adding a log4j resource file.

transport/wseb/src/test/java/org/kaazing/gateway/transport/wseb/WsebTransportTest.java
```diff
@@ -79,12 +86,7 @@
 	private NioSocketAcceptor tcpAcceptor;
 	private HttpAcceptor httpAcceptor;
 	private WsebAcceptor wsebAcceptor;
-
-    @BeforeClass
-    public static void debugging() throws Exception {
-        PropertyConfigurator.configure("src/test/resources/log4j-trace.properties");
-    }
-
+
```
jitsni: Is it possible to use the new test rule for WsebConnector here ? Just for my understanding, no need to change the code here.

transport/wseb/src/main/java/org/kaazing/gateway/transport/wseb/WsebConnectProcessor.java
```diff
@@ -47,23 +47,19 @@
...
 import org.kaazing.mina.core.future.DefaultWriteFutureEx;
 import org.kaazing.mina.core.future.WriteFutureEx;
+import org.kaazing.mina.core.session.IoSessionEx;
 import org.kaazing.mina.core.write.DefaultWriteRequestEx;
 import org.kaazing.mina.core.write.WriteRequestEx;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class WsebConnectProcessor extends BridgeConnectProcessor<WsebSession> {
+class WsebConnectProcessor extends BridgeConnectProcessor<WsebSession> {
```
jitsni: Shouldn't this be more of BridgeConnectProcessor<TransportSession> ? It is primarily flushing messages in TransportSession

transport/wseb/src/main/java/org/kaazing/gateway/transport/wseb/WsebAcceptProcessor.java
```diff
@@ -100,7 +100,7 @@ protected void flushInternal(final WsebSession session) {
         }
 
         // get write request queue and process it
-        final WriteRequestQueue writeRequestQueue = session.getWriteRequestQueue();
+        final WriteRequestQueue writeRequestQueue = session.getTransportSession().getWriteRequestQueue();
```
jitsni: Shouldn't this be processor for TransportSession as it is primarily flushing TransportSession messages ?

transport/wseb/src/main/java/org/kaazing/gateway/transport/wseb/WsebAcceptProcessor.java
cmebarrow: I think we have to do checkBuffer() for each write request in the queue, so one per execution of the loop, to make sure we issue a reconnect (downstream) as soon as we have written more that the .kb parameter

test.util/src/main/java/org/kaazing/test/util/MethodExecutionTrace.java
cmebarrow: ok, will do

test.util/src/main/java/org/kaazing/test/util/MethodExecutionTrace.java
cmebarrow: done

test.util/src/main/java/org/kaazing/test/util/MethodExecutionTrace.java
cmebarrow: Good idea but I don't have time to bite that off right now. Adding a TODO note in the class.

transport/wseb/src/test/java/org/kaazing/gateway/transport/wseb/WsebTransportTest.java
cmebarrow: Yes I think we could use it. I actually created it by copying code from this test! But I'd rather hold off until we do the work of building a WsebAcceptorRule too and including all the wseb spec tests.
Once we do that we can actually remove this test.

cmebarrow:
@jitsni Acted upon your comments, please re-review

transport/wseb/src/main/java/org/kaazing/gateway/transport/wseb/WsebAcceptProcessor.java
cmebarrow: Basically it is the processor for TransportSession now, but I was not able to directly set it as the processor on TransportSession because of type checking (would require extensive code changes which I don't want to bite off), that's why I am using TransportProcessor instead, which just wraps the WsebAcceptProcessor or WsebConnectProcessor instance.

transport/wseb/src/main/java/org/kaazing/gateway/transport/wseb/WsebConnectProcessor.java
cmebarrow: Not sure what you mean. I would have liked to make it BridgeConnectProcess<TransportSession> but that's not possible without introducing a whole new transport layer which I am trying to avoid (see also my previous comment about why WsebAcceptProcessor cannot directly be the processor on TransportSession)

transport/wseb/src/main/java/org/kaazing/gateway/transport/wseb/WsebAcceptProcessor.java
jitsni: I think the same as it involves extensive changes. Let us get the current changeset in and later worry about other things.

jitsni:
Filed an issue for travis failure https://github.com/kaazing/gateway/issues/153
The issue is unrelated for this changeset.

jitsni:
LGTM. Can this be merged ?

PR #147, opened 2015-05-19
HTTP response without Content-Length or chunked encoding causes parsing loop

jitsni:


cmebarrow:
LGTM

cmebarrow:
There was a test failure on openjdk7:
````
Failed tests: 
  FramingBinaryMessagesIT.sendBinaryMessageWithPayloadLength65536 Specified behavior did not match expected:<...xf2 0x9f 0xa9 0x79]
[read [0x88 0x02 0x03 0xe8] 

close
clos]ed
> but was:<...xf2 0x9f 0xa9 0x79]
[disconnect]ed
>
````
Could this be related to your change?

jitsni:
The test failure may not be related this change (as it seems gateway didn't send close frame)

PR #151, opened 2015-05-20
Resolves #150

### LINKED ISSUE #150: testSendLargeFile(org.apache.mina.transport.socket.nio.NioFileRegionTest) occasionally failing
This is reproducing on my Mac consistently for some stretch of time and then does not reproduce for a long periods of time.  I have not seen it reproduce elsewhere

dpwspoon:
Made the test use localhost for binding, as there is a bug in get next available port on the any address and the test is not designed to test the any address.

### DIFF
--- a/mina.core/core/src/test/java/org/apache/mina/transport/AbstractFileRegionTest.java
+++ b/mina.core/core/src/test/java/org/apache/mina/transport/AbstractFileRegionTest.java
@@ -22,7 +22,6 @@ package org.apache.mina.transport;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
-import java.net.InetAddress;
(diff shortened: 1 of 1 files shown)


mina.core/core/src/test/java/org/apache/mina/transport/AbstractFileRegionTest.java
```diff
@@ -22,6 +22,7 @@
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.net.InetAddress;
```
krismcqueen: clean up imports?

dpwspoon: fixed

PR #155, opened 2015-05-26
New ws extension spi

cmebarrow:
See specification [Server-side WebSocket Extension SPI ](https://docs.google.com/document/d/17UvMyWrO9aX1rqlNCv2J6hUBCBqiGGu7oMFYByqM_3Q/edit#heading=h.p6hxkd14m3xl) for details. Here's a summary:
- New module transport/ws/extensions contains the new SPI.
- WebSocketExtensionFactory loads available extensions using service loader and also contains the method to negotiate extensions
- Most classes from the previous framework have been removed or repurposed, notably:
  - ActiveWsExtensions is removed, replaced by the List<WebSocketExtension> returned by WebSocketExtensionFactory.negotiate). 
  - WsExtension, WsExtensionBuilder and WsExtensionParameter have been renamed to ExtensionHeader, ExtensionHeaderBuilder and ExtensionParameter and are only used for extension header parsing and construction.
- GatewayContextResolver is enhanced to create and inject the WebSocketExtensionFactory instance as a resource, and inject resources into it
- WebSocketExtensionFactory injects resources into the individual WebSocketExtensionFactorySpi instances
- Robot tests ExtensionFilterIT (for wsn) and WsebExtensionFilterIT (for wseb) show a test extension being loaded dynamically and modifying traffic in both directions.


transport/ws.extension/src/main/java/org/kaazing/gateway/transport/ws/extension/ExtensionHeaderBuilder.java
```diff
@@ -0,0 +1,158 @@
...
+
+    // Default equality is by extension token, ignoring parameters.
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || !(o instanceof ExtensionHeader)) return false;
+
+        ExtensionHeader that = (ExtensionHeader) o;
+
+        return !(extensionToken != null ? !extensionToken.equals(that.getExtensionToken()) : that.getExtensionToken() != null);
```
jitsni: Can we not use Objects.equals(extensionToken, that.getExtensionToken()) ?

transport/ws.extension/pom.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jitsni: If it is a ws extension, shouldn't this be part of ws transport (rather than a separate module) ?

cmebarrow: That's the way I did it to start with, but I run into a circular dependency when I did the resource injection to GatewayContextResolver. I could not make server module depend on transport/ws because that created a circular dependency (since it depends on transport.http which in turn server depends on). Thinking about it again I am not quite sure why that was an issue so I can try gain if you want.

transport/ws.extension/src/main/java/org/kaazing/gateway/transport/ws/extension/ExtensionHeaderBuilder.java
cmebarrow: Good idea, will do, didn't know about that method.

server/src/main/java/org/kaazing/gateway/server/context/resolve/GatewayContextResolver.java
```diff
@@ -1246,6 +1253,11 @@ private void injectResources(Collection<ServiceContext> services,
         // inject bridge service factory
         injectResources(bridgeServiceFactory, injectables);
 
+        // inject websocket extensions
+        for (WebSocketExtensionFactorySpi factory : webSocketExtensionFactory.availableExtensions()) {
```
jitsni: Do we want gateway.server need to know about WebSocketExtensionFactorySpi ?  Can it not be added on WsTransport/WsnTransport  etch that return extension instances. Say, List<WebSocketExtension> WsTransport#getExtensions(...).

Just thinking aloud (I need to go through the rest of changeset).

transport/wsn/pom.xml
```diff
@@ -28,6 +28,11 @@
         </dependency>
         <dependency>
             <groupId>org.kaazing</groupId>
+            <artifactId>gateway.transport.ws.extension</artifactId>
```
jitsni: not required anymore, right ?

cmebarrow: correct, fixed

PR #160, opened 2015-05-28
Removed empty folder/subproject

dpwspoon:


dpwspoon:
@jitsni Please merge, I've checked with Viv and this is not needed

PR #163, opened 2015-05-28
Move truststore to its own repository, gateway.truststore

ahousing:


pom.xml
```diff
@@ -198,6 +197,12 @@
                 <artifactId>json</artifactId>
                 <version>20090211</version>
             </dependency>
```
dpwspoon: I think that because this is no longer in the BOM than the enterprise.gateway will fail when it picks up the next release of kaazing/gateway.  Did you build enterprise.gateway using kaazing/gateway develop-SNAPSHOT. 

I see no reason why this couldn't be in the BOM with a version range (if boms work like that).  Alternatively, you could update enterprise.gateway with a similar change, but that is duplicating code.

ahousing: I think you mis-read the diff.  Lines 198-200 are unchanged.

dpwspoon: This is a comment on the closest change.  Line 200 through 205, which I think needs to be in bom/pom.xml.  Am I missing something on github? I don't believe there is a way to select a group of lines....

ahousing: Ah, the way it was emailed (and looks in github), I thought it was to the previous lines.  Maybe comments are "supposed" to be after the code?

I just did a "maven clean install" on ahousing/gateway, followed by "maven clean verify" on enterprise.gateway.  It's fine.  The reason, I believe, is that [enterprise.gateway/pom.xml](https://github.com/kaazing/enterprise.gateway/blob/develop/pom.xml) first lists a dependency on gateway.bom, but the second dependency is on gateway.

dpwspoon: Note: you would need to change the dependency version of gateway in enterprise.gateway to "develop-SNAPSHOT".  

But I understand how it would work now, so I'll merge it.  Thanks

Issue #167, opened 2015-05-29
Long start time on systems with low entropy (containers) due to calls to getSecureBytes.


dpwspoon:
This is due to calls on getSecureBytes which by default Java uses /dev/random.  dev/random is populated by the entropy of the system and is a [blocking call](http://en.wikipedia.org/?title=/dev/random).

```
"main" #1 prio=5 os_prio=0 tid=0x00007fa8f000b000 nid=0x7 runnable [0x00007fa8f78cf000]
   java.lang.Thread.State: RUNNABLE
	at java.io.FileInputStream.readBytes(Native Method)
	at java.io.FileInputStream.read(FileInputStream.java:255)
	at sun.security.provider.SeedGenerator$URLSeedGenerator.getSeedBytes(SeedGenerator.java:539)
	at sun.security.provider.SeedGenerator.generateSeed(SeedGenerator.java:144)
	at sun.security.provider.SecureRandom$SeederHolder.<clinit>(SecureRandom.java:203)
	at sun.security.provider.SecureRandom.engineNextBytes(SecureRandom.java:221)
	- locked <0x00000000f5f95c80> (a sun.security.provider.SecureRandom)
	at java.security.SecureRandom.nextBytes(SecureRandom.java:468)
	- locked <0x00000000f5f95fa0> (a java.security.SecureRandom)
	at org.snmp4j.security.Salt.<init>(Salt.java:55)
	at org.snmp4j.security.Salt.getInstance(Salt.java:80)
	at org.snmp4j.security.PrivDES.<init>(PrivDES.java:57)
	at org.snmp4j.security.SecurityProtocols.addDefaultProtocols(SecurityProtocols.java:152)
	- locked <0x00000000f5f775f8> (a org.snmp4j.security.SecurityProtocols)
	at org.kaazing.gateway.management.snmp.SnmpManagementServiceHandler$KaazingConfigManager.getSupportedSecurityProtocols(SnmpManagementServiceHandler.java:910)
	at org.kaazing.gateway.management.snmp.SnmpManagementServiceHandler$KaazingConfigManager.createUSM(SnmpManagementServiceHandler.java:920)
	at org.kaazing.gateway.management.snmp.SnmpManagementServiceHandler$KaazingConfigManager.initSecurityModels(SnmpManagementServiceHandler.java:726)
	at org.kaazing.gateway.management.snmp.SnmpManagementServiceHandler$KaazingConfigManager.initialize(SnmpManagementServiceHandler.java:1014)
[...]

PR #169, opened 2015-05-29
Resolves kaazing/gateway#167: slow start ups on machines with low entropy

dpwspoon:
solved by having java set -Djava.security.egd=file:/dev/urandom on startup

### DIFF
--- a/distribution/src/main/gateway/bin/gateway.start
+++ b/distribution/src/main/gateway/bin/gateway.start
@@ -41,9 +41,17 @@ GW_HOME=`cd "$PRGDIR/.." ; pwd`
 
 # You can define various Java system properties by setting the value
 # of the GATEWAY_OPTS environment variable before calling this script.
-# The script itself should not be changed. For example, the setting
-# below sets the Java maximum memory to 512MB.
-[ -z "$GATEWAY_OPTS" ] && GATEWAY_OPTS="-Xmx512m"
+# The script itself should not be changed.
+if [ -z "$GATEWAY_OPTS" ]; then
+  # Sets the Java maximum memory to 512MB
+  GATEWAY_OPTS="-Xmx512m"
+  # Sets the java.security.edg to use /dev/urandom. By default /dev/random
+  # is used.  /dev/random is a blocking call which can hang on machines that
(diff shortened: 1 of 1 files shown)


distribution/src/main/gateway/bin/gateway.start
```diff
@@ -41,9 +41,17 @@ GW_HOME=`cd "$PRGDIR/.." ; pwd`
...
 # of the GATEWAY_OPTS environment variable before calling this script.
-# The script itself should not be changed. For example, the setting
-# below sets the Java maximum memory to 512MB.
-[ -z "$GATEWAY_OPTS" ] && GATEWAY_OPTS="-Xmx512m"
+# The script itself should not be changed.
+if [ -z "$GATEWAY_OPTS" ]; then
+  # Sets the Java maximum memory to 512MB
+  GATEWAY_OPTS="-Xmx512m"
+  # Sets the java.security.edg to use /dev/urandom. By default /dev/random
+  # is used.  /dev/random is a blocking call which can hang on machines that
```
brennangaunce: Suggest:

By default, Java uses /dev/random to gather entropy data for cryptographic needs.  However, using /dev/random can cause delays during Gateway startup, especially in virtualized environments.  /dev/urandom does not require collection of entropy data in subsequent runs.  See:  https://github.com/kaazing/gateway/issues/167

jitsni:
Should we change to /dev/urandom file only docker environments ?

dpwspoon:
We did consider this and thought it would be better to have in the generic gateway.start because this could hypothetically happen on VMs, cloud environments, and linux containers in general.  

Additionally, from what I can tell urandom is secure (reference [post1](http://security.stackexchange.com/questions/3936/is-a-rand-from-dev-urandom-secure-for-a-login-key) and [post2](http://www.2uo.de/myths-about-urandom/)).  

And then we should probably add documentation to deploying in production environment that if your server generates enough entropy remove this setting. PING @vjwang @robinzimmermann @sbadugu  (not sure how to properly tag that this is required, should I open a new issue?, I believe Robin and Satish were talking about this today)

brennangaunce:
Yes - file a new issue to doc this appropriately.

PR #172, opened 2015-05-30
Ignored test that is buggy on mac os, see https://github.com/kaazing/…

dpwspoon:
…gateway/issues/171


dpwspoon:
Does anyone know if this passes on Windows?

PR #175, opened 2015-06-01
Reducing the no of iterations in the test

jitsni:
the itermittent failures

### DIFF
--- a/transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/WsnConnectorTest.java
+++ b/transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/WsnConnectorTest.java
@@ -315,7 +315,7 @@ public class WsnConnectorTest {
     @Test (timeout = 30000)
     public void shouldNotHangOnToHttpConnectSessionsWhenEstablishingAndTearingDownWsnConnectorSessions() throws Exception {
 
-        long iterations = 50;
+        long iterations = 100;
 
         final URI location = URI.create("wsn://localhost:8000/echo");
         Map<String, Object> addressOptions = Collections.emptyMap(); //Collections.<String, Object>singletonMap("http.transport", URI.create("pipe://internal"));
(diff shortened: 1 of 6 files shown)


dpwspoon:
The same test failed, so it didn't fix the issue :-(

dpwspoon:
NVM, maybe I was looking wrong travis build... My mistake, looks like it is passing now

jitsni:
Yes, changing the iterations didn't fix. I added one more commit, let us see if that fixes  its.

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/WsnConnectorTest.java
```diff
@@ -335,12 +335,8 @@ public void doMessageReceived(final IoSessionEx session, Object message)
...
                 System.out.println("Acceptor: received message: " + Utils.asString(buf.buf()));
                 IoBufferAllocatorEx<?> allocator = sessionEx.getBufferAllocator();
-                session.write(allocator.wrap(asByteBuffer("Reply from acceptor"))).addListener(new IoFutureListener<IoFuture>() {
-                    @Override
-                    public void operationComplete(IoFuture future) {
-                        session.close(true);
-                    }
-                });
+                WriteFuture writeFuture = session.write(allocator.wrap(asByteBuffer("Reply from acceptor")));
+                writeFuture.addListener(future -> session.close(false));
```
brennangaunce: Note session.close(true) before and session.close(false) after the change.  Otherwise the semantics of the code before and after the change are the same.

jitsni:
Don't merge the pull request. Trying to understand why WsnConnectFuture is not finished.

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/WsnConnectorTest.java
jitsni: Originally session.close(true) which abruptly tears the connection. session.close(false) initiates orderly shutdown and test passes. Even though test passes, I am not sure why tearing down the connection doesn't fire wsn's ConnectFuture. Hence I am still investigating (as I cannot produce on my machine I am adding more diagnostic messages)

jitsni:
Closing this one and will open a new one once fix is available.

PR #178, opened 2015-06-04
 http.keepalive.max.connections config to cache http idle connections.

jitsni:


jitsni:
@dpwspoon can you also review the changes

server/src/main/xsd/gateway-config-201409.xsd
```diff
@@ -1253,6 +1253,16 @@
                     </documentation>
                 </annotation>
             </element>
+            <element name="http.keepalive.max.connections" maxOccurs="1" minOccurs="0" type="nonNegativeInteger" default="5">
```
jitsni: @robinzimmermann can you review the name. It is something equivalent to nginx 's keepalive connections config. see here http://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultConnectOptionsContext.java
```diff
@@ -208,6 +210,19 @@ private Integer getHttpKeepaliveTimeout() {
         return httpKeepaliveTimeout;
     }
 
+    private Integer getHttpKeepaliveMaxConnections() {
+        int maxConnections = DEFAULT_HTTP_KEEPALIVE_MAX_CONNECTIONS;
+        String connectionsValue = options.get("http.keepalive.max.connections");
+        if (connectionsValue != null) {
+            int val = Integer.parseInt(connectionsValue);
+            if (val > 0) {
+                maxConnections = val;
```
dpwspoon: Do we wan't to throw a warning in an else block that it is misconfigured.  Would a 0 max connections mean don't allow keep alive?

resource.address/http/src/main/java/org/kaazing/gateway/resource/address/http/HttpResourceAddress.java
```diff
@@ -248,6 +256,12 @@ private HttpKeepAliveTimeoutOption() {
         }
     }
 
+    private static final class HttpKeepAliveMaxConnectionsOption extends HttpResourceOption<Integer> {
+        private HttpKeepAliveMaxConnectionsOption() {
+            super(Kind.KEEP_ALIVE_MAX_CONNECTIONS, "keepalive.max.connections", 5);
```
krismcqueen: Is it possible to use a constant for the default so that there is only one constant for here, the ConnectOptions, etc.?

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultConnectOptionsContext.java
```diff
@@ -56,11 +56,12 @@
 
     private static final long DEFAULT_WS_INACTIVITY_TIMEOUT_MILLIS = 0L;
     private static final int DEFAULT_HTTP_KEEPALIVE_TIMEOUT = 30; //seconds
+    private static final int DEFAULT_HTTP_KEEPALIVE_MAX_CONNECTIONS = 5;
```
krismcqueen: See feedback for HttpResourceAddress about having only one constant that is shared to avoid cases where default values diverge.

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultConnectOptionsContext.java
krismcqueen: I second the point for what zero means here.  Is it possible to disable http keepalive?  Is that desired in any scenarios?

service/http.proxy/src/test/java/org/kaazing/gateway/service/http/proxy/HttpProxyPersistenceTest.java
```diff
@@ -0,0 +1,184 @@
...
+            Thread t1 = new Thread(new HttpClient());
+            t1.start(); t1.join();
+            // server should have received only one connection
+            // pool should have max configured connections = 1
+            assertEquals(1, handler.getConnections());
+
+            // t2 client sends 2 requests
+            Thread t2 = new Thread(new HttpClient());
+            t2.start(); t2.join();
+
```
dpwspoon: Could we make this deterministic on which happens by making it a k3po test?

service/http.proxy/src/test/java/org/kaazing/gateway/service/http/proxy/HttpProxyPersistenceTest.java
```diff
@@ -0,0 +1,184 @@
...
+        ServerHandler handler = new ServerHandler();
+        OriginServer originServer = new OriginServer(8080, handler);
+
+        try {
+            originServer.start();
+            gateway.start(configuration);
+
+            // t1 client sends 2 requests
+            Thread t1 = new Thread(new HttpClient());
+            t1.start(); t1.join();
```
krismcqueen: If you need to do this synchronously then is there a need for a new thread?

transport/http/src/main/java/org/kaazing/gateway/transport/http/PersistentConnectionPool.java
```diff
@@ -21,62 +21,90 @@
...
+        IoSession transportSession = httpSession.getParent();
+
+        SERVER_ADDRESS.set(transportSession, serverAddress);
+
+        // Take care of transport session close
+        CloseFuture closeFuture = transportSession.getCloseFuture();
         closeFuture.addListener(closeListener);
-        addIdleFilter(session, keepAliveTimeout);
+
+        // Track transport session idle
```
krismcqueen: Perhaps you can elaborate in this comment to describe the requirements for tracking session idle?  Restating what the code is doing in some comment is typically a cause to remove said comment, but giving motivations / requirements is useful for future reading of the code to understand why some code is there.

transport/http/src/main/java/org/kaazing/gateway/transport/http/PersistentConnectionPool.java
```diff
@@ -21,62 +21,90 @@
...
-        }
-        sessions.add(session);
-        CloseFuture closeFuture = session.getCloseFuture();
+
+        HttpResourceAddress serverAddress = (HttpResourceAddress)httpSession.getRemoteAddress();
+        IoSession transportSession = httpSession.getParent();
+
+        SERVER_ADDRESS.set(transportSession, serverAddress);
+
+        // Take care of transport session close
```
krismcqueen: What does "take care of" mean in this case?  Is this about ensuring the session is removed from the connection pool on transport close?

transport/http/src/main/java/org/kaazing/gateway/transport/http/PersistentConnectionPool.java
```diff
@@ -85,46 +113,92 @@ public void recycle(IoSession session, Integer keepAliveTimeout) {
...
+        if (transportSession != null) {
+            // Got a cached persistent connection
 
-            if (logger.isDebugEnabled()) {
-                logger.debug(String.format("Reusing (removing from pool) http connect persistent connection %s", session));
+            // Remove session idle tracking for this session
+            transportSession.getConfig().setBothIdleTime(0);
+            IoFilterChain filterChain = transportSession.getFilterChain();
+            if (filterChain.contains(IDLE_FILTER)) {
+                filterChain.remove(IDLE_FILTER);
```
krismcqueen: Can an in-use session go idle and require being closed?  Is this handled by a different filter?

krismcqueen:
@jitsni Looks good, just some feedback about possibly tying to I/O Worker threads to have less randomness in terms of how idle connections get re-used, and that feedback was discussed in person.

dpwspoon:
LGTM

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultConnectOptionsContext.java
jitsni: It is possible to disable keep-alive using <http.keepalive>disabled</http.keepalive> connect option. It may be done for security reasons. It is also useful if there is a bug in reusing cached connections or if we know that upstream server doesn't support keepalive etc.

But it may be useful to just warn if the number is negative and use the default value.

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultConnectOptionsContext.java
jitsni: gateway.server doesn't have any dependency on http resource address. So, I guess gateway.server would just pass the config (if it is specified) to resource address. Then resource address will deal with default value. That way there is only one place we deal with the default values. So removing the constant here.

transport/http/src/main/java/org/kaazing/gateway/transport/http/PersistentConnectionPool.java
jitsni: I don't think gateway tracks in-use session idle for both http accept and connect sessions.

jitsni:
@krismcqueen,
 ready for second round reivew

krismcqueen:
@jitsni Looks good.  The build stalled with no apparent errors, so I restarted it.  If it passes then I will merge this.

jitsni:
Let us not merge until robin reviews the config change.

server/src/main/xsd/gateway-config-201409.xsd
```diff
@@ -1253,6 +1253,18 @@
                     </documentation>
                 </annotation>
             </element>
+            <element name="http.keepalive.max.connections" maxOccurs="1" minOccurs="0" type="nonNegativeInteger" default="2">
```
jitsni: @robinzimmermann 
Can you review this config.

vjwang: @veschup Please also review the config.

dpwspoon:
LGTM, holding off on merge until notified to do so

server/src/main/xsd/gateway-config-201409.xsd
veschup: The annotation is not clear. It would be better if you took the perspective of predicting and answering questions such as these: When would I use this? Why? Add recommendations/best practices for setting size (small enough?).  

Also terminology is a problem. What is an upstream server -- is that one of the Gateways (DMZ or internal/Enterprise)? What is the "worker"?

PR #186, opened 2015-06-09
Expose extension point for amqp.proxy service

krismcqueen:
Expose an extension point for AMQP proxy service such that additional behaviors can be layered in once the accept side of the proxied connection is ready for initialization.


krismcqueen:
@jitsni note:  There maybe be an opportunity to consolidate this code with the ProxyServiceExtensionSpi.  However, the use case for this extension was a little different so for now I have made it a separate extension.

jitsni:
I have the similar concerns. May be it can be consolidated with ProxyServiceExtensionSpi. If we add initAcceptSession(..) method on ProxyServiceExtensionSpi, would that be any better ?

krismcqueen:
@jitsni But then any extension would have to implement both methods presumably.  That seems less than ideal.  Unless, of course, the extension SPI is changed to have an empty implementation rather than an abstract method, giving an extension the opportunity to pick and choose.  Would that be better?

krismcqueen:
@jitsni updated with feedback for consolidating the extension SPIs into a single extension SPI.

service/proxy/src/main/java/org/kaazing/gateway/service/proxy/ProxyServiceExtensionSpi.java
```diff
@@ -34,5 +36,32 @@
...
+
+    /**
+     * Whenever a new client connection is received (on the configured "accept")
+     * the extension is giving an opportunity to initialize the associated session.
+     *
+     * @param acceptSession - the accept side of the end-to-end proxy connection
+     * @param properties - the service properties that contain the values used to
+     *                     configure the proxy service
+     */
+    default void initAcceptSession(IoSession acceptSession, ServiceProperties properties) {
```
jitsni: At some point, we will remove ServiceProperties. For now, we will continue with this changeset.

jitsni:
Looks good to me.

PR #192, opened 2015-06-14
Fix up the timing of when proxy service extensions are invoked (initAcceptSession)

krismcqueen:
Move the initAcceptSession call on any extensions until after the proxy service has done its own initialization in the sessionOpened() method.  This ensures the filter chain is set up as per the proxy service's needs before any extension adds to the filter chain (most importantly the codec will be in the right place so extensions see the correct decoded messages).

### DIFF
--- a/service/proxy/src/main/java/org/kaazing/gateway/service/proxy/ProxyServiceHandler.java
+++ b/service/proxy/src/main/java/org/kaazing/gateway/service/proxy/ProxyServiceHandler.java
@@ -66,12 +66,6 @@ public class ProxyServiceHandler extends AbstractProxyAcceptHandler {
             final Object serviceRegistration = acceptSession.getAttribute(HttpAcceptor.SERVICE_REGISTRATION_KEY);
             final String nextProtocol = BridgeSession.NEXT_PROTOCOL_KEY.get(acceptSession);
 
-            // hook in any ProxyServiceExtensions here and give them a chance
-            // to initialize the accept side of the proxy connection
-            for (ProxyServiceExtensionSpi extension : extensions) {
-                extension.initAcceptSession(acceptSession, getServiceContext().getProperties());
-            }
-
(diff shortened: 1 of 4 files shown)


service/proxy/src/main/java/org/kaazing/gateway/service/proxy/ProxyServiceHandler.java
```diff
@@ -106,6 +100,12 @@ public void initializeSession(IoSession connectSession, ConnectFuture future) {
             }
 
             super.sessionOpened(acceptSession);
```
jitsni: Should the super.sessionOpened() be called after the extensions are invoked ?

service/proxy/src/test/java/org/kaazing/gateway/service/proxy/ProxyServiceExtensionTest.java
```diff
@@ -0,0 +1,117 @@
...
+                }
+            }
+        }, "backendServer-thread");
+        t.start();
+
+        boolean backendBound = backendReady.await(1, TimeUnit.SECONDS);
+        assertTrue("Backend Server failed to bind, pre-conditions not established", backendBound);
+
+        // connect to the service to ensure the extension is executed
+        Socket clientSocket = new Socket("localhost", 8888);
```
jitsni: See if you can use ARM block (try with resources syntax). Same goes with ServerSocket in the above

jitsni:
Overall looks good. Once you are ready, let me know.

service/proxy/src/main/java/org/kaazing/gateway/service/proxy/ProxyServiceHandler.java
krismcqueen: No.  If the extension needs to put something on the filter chain then the code needs to be added first so that extensions can just add themselves to the end.  Basically this is saying that the Gateway will do its work first, and then extensions will get a shot.

krismcqueen:
@jitsni feeback applied, ready for next round

PR #196, opened 2015-06-16
Adding a helper class to log exceptions

jitsni:
- With debug level, it prints stack traces along with the message
- With info level, just the message


jitsni:
Forward porting from 4.0

PR #198, opened 2015-06-16
Fixing two issues with WsnConnector impl w.r.t handshake

jitsni:
- Now, sends random Sec-WebSocket-Key header
- Validates Sec-WebSocket-Accept header value


jitsni:
Forward porting from 4.x

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/OpeningHandshakeIT.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private final WsnConnectorRule connectorRule = new WsnConnectorRule();
+
+    @Rule
+    public final TestRule chain = outerRule(connectorRule).around(k3po).around(timeout);
+
+
+    @BeforeClass
+    public static void initLogging() throws Exception {
+        PropertyConfigurator.configure("src/test/resources/log4j-trace.properties");
```
krismcqueen: I'm not up to date on the latest MethodExecuteTrace, but could you also use the MethodExecuteTrace as a rule with the given log4j-trace.properties?  I'm not sure what our best practices are for getting consistent logging in the tests.

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/WsnConnectorRule.java
```diff
@@ -0,0 +1,140 @@
...
+        Map<String, Object> connectOptions = new HashMap<>();
+        if (wsInactivityTimeout != null) {
+            connectOptions.put("inactivityTimeout", wsInactivityTimeout);
+        }
+        final ResourceAddress connectAddress =
+                addressFactory.newResourceAddress(URI.create(connect), connectOptions);
+
+        ConnectFuture future = wsnConnector.connect(connectAddress, connectHandler, null);
+
+        future.await(TimeUnit.MILLISECONDS.toMillis(3000));
```
krismcqueen: Just out of curiosity, should the test decide whether to wait on the future and for how long to different tests might be able to use different behavior?  For example, if this just did a

return wsnConnector.connect(connectAddress, connectHandler, null);

it might give individual tests a chance to wait different amounts of time in the event failure is expected.  However, this seems fine to me now given the current requirements and maybe this can be refactored in the future.

krismcqueen:
@jitsni Review finished, one minor bit of feedback on MethodExecuteTrace v. PropertyConfigurator

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/OpeningHandshakeIT.java
jitsni: Update to use MethodExecutionTrace

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/WsnConnectorRule.java
jitsni: I guess I just followed WsebConnectorRule ! But this makes sense. I modified WsnConnectorRule to return future and the test can decide how to wait and get the values. We evolve further in future as the connect method may need some more arguments

PR #201, opened 2015-06-18
Removed docker build to unstable-gateway docker hub

dpwspoon:


dpwspoon:
If we need unstable gateway releases we can build it in tools.docker.  We will keep the module around as a convenience for developers who want a docker image from develop-SNAPSHOT

PR #205, opened 2015-06-20
New TCP transport extension SPI: interfaces TcpExtensionFactorySpi an…

cmebarrow:
…d TcpExtension.
This pull request is for **review purposes only**. Please **do not merge** until the implementation is complete

### DIFF
--- a/transport/nio/src/main/java/org/kaazing/gateway/transport/nio/spi/TcpExtension.java
+++ b/transport/nio/src/main/java/org/kaazing/gateway/transport/nio/spi/TcpExtension.java
@@ -1,0 +1,41 @@
+/**
+ * Copyright (c) 2007-2014 Kaazing Corporation. All rights reserved.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.kaazing.gateway.transport.nio.spi;
+
+import org.apache.mina.core.session.IoSession;
+
+/**
+ * {@link TcpExtension} is part of <i>Service Provider Interface</i> <em>(SPI)</em> for TCP transport extension developers.
+ * When TCP URI is bound, an instance of this class may be created using the corresponding
+ * {@link TcpExtensionFactorySpi} that is registered through META-INF/services. This class can perform actions during
+ * IoSession initialization when a client connects to the bound TCP URI, for example, if could add a filter to the
+ * filter chain.
+ */
+public interface TcpExtension {
+
+    /**
+     * Called to give the extension an opportunity to act on each new client connection
+     * @param session  The IoSession representing a new client connection
+     */
+    void initializeSession(IoSession session);
+
+}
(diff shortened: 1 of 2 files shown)


cmebarrow:
@jitsni @jfallows Please review the proposed API but do not merge yet

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/spi/TcpExtensionFactorySpi.java
```diff
@@ -0,0 +1,47 @@
...
+ * <p>
+ */
+public interface TcpExtensionFactorySpi {
+
+    /**
+     * Called whenever a TCP accept URI is bound
+     * @param address  TCP address being bound (URI plus accept options)
+     * @return  A TcpExtension object to be used for all accepted sessions for this bind, or null if no extension is needed
+     */
+    TcpExtension bind(ResourceAddress address);
```
jitsni: address is used pass config (via address's resource options) to TcpExtension ?

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/spi/TcpExtension.java
```diff
@@ -0,0 +1,41 @@
...
+ * IoSession initialization when a client connects to the bound TCP URI, for example, if could add a filter to the
+ * filter chain.
+ */
+public interface TcpExtension {
+
+    /**
+     * Called to give the extension an opportunity to act on each new client connection
+     * @param session  The IoSession representing a new client connection
+     */
+    void initializeSession(IoSession session);
```
jitsni: Is this called during IoSession#sessionCreated() or when tcp session is constructed ?

jitsni:
Adding @krismcqueen too

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/spi/TcpExtensionFactorySpi.java
cmebarrow: Yes, that's correct. The accept option that concerns us for bandwidth capping is  "tcp.maximum.outbound.rate"

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/spi/TcpExtension.java
jitsni: I see the impl details in the doc on how this method is invoked

cmebarrow: The plan is to call it from BridgeSessionInitializer.initialize which is called from sessionOpened (AbstractNioSocketAcceptor.tcpBridgeHandler.doSessionOpened)

cmebarrow:
Thanks Jitu. Please do not merge until I have completed the implementation. I will use the same branch.

jfallows:
Hey @cmebarrow, looks good.

Given that we are distinguishing between public api and internal implementation, please make sure the code is separated as follows:
```
package org.kaazing.gateway.transport.nio.spi;
```
for these public api interfaces, and
```
package org.kaazing.gateway.transport.nio.internal.*;
```
for the internal implementation.

We need to take opportunities like this to iterate towards _all_ internal implementation in a _internal_ package.

Feel free to perform the re-package in a separate PR (still part of this task though) and then catch-up here to keep the code review diffs pristine for this PR.

cmebarrow:
I'll do the internal package refactoring along with the implementation in a separate branch (tcp_extension_impl) and create a separate pull request for it.

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/spi/TcpExtensionFactorySpi.java
krismcqueen: The extension is not binding, correct?  Did you consider a different name for this method to indicate the bind has happened already and now the extension can act?

krismcqueen:
Looks good to me.

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/spi/TcpExtensionFactorySpi.java
cmebarrow: Actually this method is called during bind, and allows the extension factory to decide if an extension needs to be added based on the resource address and its options. Would you prefer me to rename it to getExtension or something else? I chose the name bind because that makes it clear it happens during bind.

cmebarrow:
Closed because we do not actually want to merge this, it was just a spike

PR #207, opened 2015-06-23
Have the cobertura build happen during normal building rather than during after_success

ahousing:
Must  be done due to lack of reporting of failures during after_success by travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: Java
 jdk:
   - oraclejdk8
-install: mvn -q clean -DskipTests -DskipITs install && mvn -q clean cobertura:cobertura-integration-test coveralls:report
+install: mvn -q clean -DskipTests -DskipITs install
 script: mvn -q clean verify && mvn -q clean cobertura:cobertura-integration-test coveralls:report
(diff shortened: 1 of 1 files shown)


brennangaunce:
Still getting this error.  @ahousing please file a separate ticket for this and investigate the cause.

```
[ERROR] net.sourceforge.cobertura.javancss.parser.ParseException: Encountered " ":" ": "" at line 313, column 71.
Was expecting one of:
    "instanceof" ...
    ")" ...
```

.travis.yml
```diff
@@ -1,7 +1,5 @@
 language: Java
 jdk:
   - oraclejdk8
-install: mvn -q clean -DskipTests -DskipITs install
-script: mvn -q clean verify
-after_success:
-  - mvn -q clean cobertura:cobertura-integration-test coveralls:report
+install: mvn -q clean -DskipTests -DskipITs install && mvn -q clean cobertura:cobertura-integration-test coveralls:report
```
dpwspoon: This is running the cobertura tests and report twice. Once in the install and once in the script.  The install should only download dependencies.  So I would remove the cobertura from the install step.

ahousing: Ah, I was thinking "after_success" was more like "final" in bamboo, not "the next stage", like in maven.

ahousing:
Filed #208 about the cobertura error itself.

Issue #203, opened 2015-06-19
Sporadic test failures in SseSameOriginIT on Windows 7


cmebarrow:
Build of gateway.transport.sse frequently fails on Windows with failures like the following:
````
sseConnectFromBridgeAndGetDataThroughBroadcast(org.kaazing.gateway.transport.sse.SseCrossOriginIT) starting
SseCrossOriginIT.sseConnectFromBridgeAndGetDataThroughBroadcast FAILED with exception org.junit.ComparisonFailure: Specified behavior did not match expected:<...ait TEST_OVER
close
[clos]ed


#Front end 
con...> but was:<...ait TEST_OVER
close
[disconnect]ed


#Front end 
con...>
org.junit.ComparisonFailure: Specified behavior did not match expected:<...ait TEST_OVER
close
[clos]ed


#Front end 
con...> but was:<...ait TEST_OVER
close
[disconnect]ed


#Front end 
con...>
	at org.junit.Assert.assertEquals(Assert.java:115)
````

cmebarrow:
The trace level log messages (now available because I added the MethodExecutionTrace rule to the test) indicate the underlying cause may be that the backend TCPConnection is being closed immediately after writing the data that should be sent to the client. This may cause the client connection to be prematurely closed before the data can be written out to it.

cmebarrow:
The previous comment is incorrect. That is not the cause of the problem because the broadcast service does do not close client connections when the backend is closed. I managed to get a WireShark of SseSameOriginIT.sseIe8HttpxeConnectAndGetData succeeding and then failing on the next run. The Wireshark shows the traffic on the wire is identical in both cases (for the client and backend connections). So this looks like an issue with k3po. This is confirmed by the suspicious nature of the reported difference. The script has 
````
read /:.*/ "[\n"
````
so the reported error that the received text was 
````
read /:.*/ "[ ]"
````
makes no sense. k3po should be reading any bytes up to the first '\n' byte.

cmebarrow:
Here is the console output for the failed (second) run
[sse_issue_203_ok_then_failed.out | uploaded via ZenHub](https://files.zenhub.io/55877f02c4315ca333764178)

Here is the wireshark trace for the succeeded and failed runs. Look for tcp.port==8005 for the client tcp streams and tcp.port=7556 for the backend streams (2 each, for the successful run than the failed run)
[...]

PR #200, opened 2015-06-18
Update Doc links

chadpowers:
This request contains changes to doc links within the Markdown files. Links are updated to direct to correct path, either by updating the link's filename or by replacing a link with the absolute URL (for cross-repo doc links). These errors originally occurred during Doc migration.

Issue #162, opened 2015-05-28
sporadic failure of wsn build: WsnConnectorTest.shouldNotHangOnToHttpConnectSessionsWhenEstablishingAndTearingDownWsnConnectorSessions


ahousing:
In one of 8 builds I did on my machine, I got the following failure.

```
WsnConnectorTest.shouldNotHangOnToHttpConnectSessionsWhenEstablishingAndTearingDownWsnConnectorSessions FAILED with exception org.junit.runners.model.TestTimedOutException: test timed out after 30000 milliseconds
org.junit.runners.model.TestTimedOutException: test timed out after 30000 milliseconds
	at java.lang.Object.wait(Native Method)
	at org.apache.mina.core.future.DefaultIoFuture.await(DefaultIoFuture.java:98)
	at org.apache.mina.core.future.DefaultConnectFuture.await(DefaultConnectFuture.java:106)
	at org.kaazing.gateway.transport.wsn.WsnConnectorTest.shouldNotHangOnToHttpConnectSessionsWhenEstablishingAndTearingDownWsnConnectorSessions(WsnConnectorTest.java:373)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:483)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.lang.Thread.run(Thread.java:745)
=================== BEGIN STORED LOG MESSAGES ===========================
...
=================== END STORED LOG MESSAGES ===========================
[...]

PR #211, opened 2015-06-24
#203 Avoid syntax like read /:.*/ "\n" in transport/sse k3po test scr…

cmebarrow:
…ipts by including \n in the regular expression, to avoid failures when the data is fragmented. See [k3po issue #200](https://github.com/k3po/k3po/issues/200) for more background info.


krismcqueen:
The build failed due to a known issue.  I'm re-running it.  (Issue https://github.com/kaazing/gateway/issues/162).

PR #222, opened 2015-07-03
Added changes for supporting multiple gateways on the same host and merged refactoring from Marina's fork (removed StringManager and latestException)

irina-mitrea-luxoft:
Added changes for supporting multiple gateways on the same host and merged latestException and Agrona file String support from Marina's fork


pom.xml
```diff
@@ -33,6 +33,7 @@
         <jmock.version>2.6.0</jmock.version>
         <slf4j.log4j.version>1.7.5</slf4j.log4j.version>
         <agrona.version>0.3.1</agrona.version>
+        <metrics.version>develop-SNAPSHOT</metrics.version>
```
krismcqueen: This change caused the build to break.  I suspect metrics has to be officially released and then this updated to use the released version.

management/src/main/java/org/kaazing/gateway/management/agrona/ex/AtomicStringEntity.java
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.kaazing.gateway.management.agrona.ex;
```
krismcqueen: What is the ex package?  I think I missed something here to help me understand this.  Is this an internal implementation that should go into a *.internal package?

management/src/main/java/org/kaazing/gateway/management/monitoring/entity/impl/DefaultStringMonitoringEntityStub.java
```diff
@@ -0,0 +1,46 @@
...
+import org.kaazing.gateway.management.monitoring.entity.StringMonitoringEntity;
+
+/**
+ * Agrona specific String monitoring entity which uses AtomicStringEntity as the underlying implementation.
+ */
+public class DefaultStringMonitoringEntityStub implements StringMonitoringEntity {
+
+    @Override
+    public StringMonitoringEntity setValue(String value) {
+        return this;
```
krismcqueen: was this supposed to store the value in anyway, or is it a dummy implementation?

krismcqueen:
@irina-mitrea-luxoft I've made a first pass, ping me when the build issue is addressed and other questions answered and I'll take another look.  Thanks.

management/src/main/java/org/kaazing/gateway/management/monitoring/entity/impl/DefaultStringMonitoringEntityStub.java
mgherghe: this is a dummy implementation, used when Agrona is disabled in the gateway.

management/src/main/java/org/kaazing/gateway/management/agrona/ex/AtomicStringEntity.java
mgherghe: The agrona.ex package is an extension of the Agrona library. We needed this for implementing a CountersManager like functionality for persisting Strings in Agrona.

pom.xml
irina-mitrea-luxoft: A mail was sent by Marius on this topic, we need to clarify the component release process.

management/src/main/java/org/kaazing/gateway/management/agrona/ex/AtomicStringEntity.java
```diff
@@ -0,0 +1,85 @@
...
+     * @return the latest value for the String monitoring entity.
+     */
+    public String get() {
+        return buffer.getStringUtf8(offset, NATIVE_BYTE_ORDER);
+    }
+
+    /**
+     * Free the counter slot for reuse.
+     */
+    public void close() {
```
irina-mitrea-luxoft: We can add the override annotation to this method.

management/src/main/java/org/kaazing/gateway/management/agrona/ex/StringsManager.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.Deque;
+import java.util.LinkedList;
+import java.util.function.BiConsumer;
+
+import uk.co.real_logic.agrona.concurrent.AtomicBuffer;
+
+/**
+ * Manages the allocation and freeing of String monitoring entities.
+ */
+public class StringsManager {
```
irina-mitrea-luxoft: We could have an interface for this class exposing its contract.

management/src/main/java/org/kaazing/gateway/management/agrona/ex/StringsManager.java
```diff
@@ -0,0 +1,157 @@
...
+
+    public static final int LABEL_SIZE = 1024;
+    public static final int STRING_ENTITY_SIZE = 1024;
+    public static final int UNREGISTERED_STRING_SIZE = -1;
+
+    private static final ByteOrder NATIVE_BYTE_ORDER = ByteOrder.nativeOrder();
+
+    private final AtomicBuffer labelsBuffer;
+    private final AtomicBuffer valuesBuffer;
+    private final Deque<Integer> freeList = new LinkedList<>();
```
irina-mitrea-luxoft: We could specify: = new LinkedList<Integer>() above.

management/src/main/java/org/kaazing/gateway/management/monitoring/entity/impl/AgronaMonitoringEntityFactory.java
```diff
@@ -39,17 +42,18 @@
 public class AgronaMonitoringEntityFactory implements MonitoringEntityFactory {
 
     private CountersManager countersManager;
+    private StringsManager stringsManager;
 
     // These are needed for the cleanup work that needs to be done in the close method.
     private File monitoringDirectory;
     private MappedByteBuffer mappedMonitorDirectory;
     private List<AtomicCounter> counters = new CopyOnWriteArrayList<AtomicCounter>();
+    private List<AtomicStringEntity> entities = new CopyOnWriteArrayList<AtomicStringEntity>();
```
irina-mitrea-luxoft: We could define this list name to something more specific (e.g. stringEntities), since counters could also be considered to be entities.

distribution/src/main/gateway/bin/gateway.start.bat
```diff
@@ -19,33 +19,46 @@
...
+rem Create the classpath.
+
+rem Add a directory for management support
+set JAVA_LIBRARY_PATH=%GW_HOME%\lib\sigar
+
+rem Verify if the gateway identifier was provided (required by Agorna multiple gateway instances)
+
+set GW_ID=
+
+if NOT "%GATEWAY_OPTS%"=="%GATEWAY_OPTS:org.kaazing.gateway.management.AGRONA_ENABLED=true=%" (
```
cmebarrow: What exactly is the effect of this? How would users turn on the feature?

distribution/src/main/gateway/bin/gateway.start
```diff
@@ -56,5 +56,15 @@ fi
 # Add a directory for management support
 JAVA_LIBRARY_PATH="$GW_HOME/lib/sigar"
 
-exec java $GATEWAY_OPTS -Djava.library.path="$JAVA_LIBRARY_PATH" -XX:+HeapDumpOnOutOfMemoryError -cp "$GW_HOME/lib/*" org.kaazing.gateway.server.Main $*
+# Verify if the gateway identifier was provided (required by Agorna multiple gateway instances)
+if [ "$GATEWAY_IDENTIFIER" == "" ] && [[ $GATEWAY_OPTS == *"org.kaazing.gateway.management.AGRONA_ENABLED=true"* ]]; then
```
cmebarrow: What if $GATEWAY_IDENTIFIER is set and "org.kaazing.gateway.management.AGRONA_ENABLED=true" is not present in GATEWAY_OPTIONS? If I understand the code correctly we would in this case go down the else branch and start the metrics server, which seems wrong.
What is the exact intention of this script change?

distribution/src/main/gateway/bin/gateway.start.bat
irina-mitrea-luxoft: This performed a replace on GATEWAY_OPTS for the "org.kaazing.gateway.management.AGRONA_ENABLED" String. It then verified that the initial GATEWAY_OPTS String was different than the one with the replace performed, which indicated the fact that the substring was found in the parent String. 
Discovered that this implementation however had a bug when AGRONA_ENABLED was set to false due to the presence of the "=" character in the search String (which is the character which indicated the separation between the substring to be replace and the value to perform the replace with).
Due to this issue and due to the fact that the previous implementation is difficult to understand switched to an implementation which may be more explicit.

distribution/src/main/gateway/bin/gateway.start.bat
```diff
@@ -19,33 +19,56 @@
...
-)
-
-rem Create the classpath.
-
-rem Add a directory for management support
-set JAVA_LIBRARY_PATH=%GW_HOME%\lib\sigar
-
-java %GATEWAY_OPTS% -Djava.library.path="%JAVA_LIBRARY_PATH%" -XX:+HeapDumpOnOutOfMemoryError -cp "%GW_HOME%\lib\*" org.kaazing.gateway.server.WindowsMain %*
-
+@echo off
```
cmebarrow: Why are there so many diffs here? I would would expect to see diffs only for the part you added. Do you have git cponfig setting core.autocrlf=true? You can see this by running the following command: 
```
git config --list
```

distribution/src/main/assembly/windows-bin.xml
```diff
@@ -30,6 +30,12 @@
   <includeBaseDirectory>true</includeBaseDirectory>
 
   <fileSets>
+
+    <fileSet>
+      <outputDirectory>lib/metrics.statsd</outputDirectory>
```
cmebarrow: This results in a subdirectory metrics.statd being created inside {gateway_home}/lib with further subdirectories bin, conf and lib. I do not think this is the right place for this since the gateway/lib directory is supposed to contain just jar files. I think a better location would be just put it directly under the Gateway home, which I think could be done by changing to
```
<outputDirectory>metrics.statsd</outputDirectory>
```
We'll need to see with PM (Robin) ultimately on this, but for now I think this is a reasonable proposal to go with.

distribution/src/main/assembly/unix-bin.xml
```diff
@@ -30,6 +30,12 @@
   <includeBaseDirectory>true</includeBaseDirectory>
 
   <fileSets>
+
+    <fileSet>
+      <outputDirectory>lib/metrics.statsd</outputDirectory>
```
cmebarrow: See my comment on windows-bin.xml: I recommend changing to just metrics.statsd

distribution/src/main/gateway/bin/scripts/metrics.statsD.start
```diff
@@ -0,0 +1,24 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+cd ../lib/metrics.statsd/bin
+./metrics.statsd.start "$@"
```
cmebarrow: End of file market is missing, please add a newline

distribution/src/main/gateway/bin/scripts/metrics.statsD.start.bat
```diff
@@ -0,0 +1,32 @@
...
+
+if "%OS%" == "Windows_NT" SETLOCAL EnableDelayedExpansion
+rem ---------------------------------------------------------------------------
+rem Windows start script for Kaazing Gateway
+rem ---------------------------------------------------------------------------
+
+cd %~dp0
+
+cd ../../lib/metrics.statsd/bin
+metrics.statsd.start.bat %*
```
cmebarrow: Same issue with end of line missing on last line.

pom.xml
cmebarrow: All we need do is make sure I emerged the metrics upon request first, then we release metrics, then I can merge this request.

cmebarrow:
@irina-mitrea-luxoft I have finished reviewing this. Please react to my review comments. Now I will look at the metrics pull request, but I may not be able to finish reviewing today.

distribution/src/main/gateway/bin/gateway.start.bat
irina-mitrea-luxoft: I have now set this value to "true" (the value was previously set to "input").

distribution/src/main/gateway/bin/scripts/metrics.statsD.start
```diff
@@ -0,0 +1,24 @@
...
+# 
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+cd ../metrics.statsd/bin
```
cmebarrow: Does this need to be `cd ../../ metrics.statd.bin` like the .bat file?

distribution/src/main/gateway/bin/gateway.start.bat
```diff
@@ -19,33 +19,56 @@
...
+    set AGRONA_ENABLED=true
+)
+
+rem Agrona is enabled
+if "%AGRONA_ENABLED%"=="true" (
+    if "%GATEWAY_IDENTIFIER%" NEQ "" (
+        set GW_ID="-Dorg.kaazing.gateway.server.GATEWAY_IDENTIFIER=%GATEWAY_IDENTIFIER%"
+        
+        rem Startup the StatsD publisher
+        START CMD /C CALL "scripts/metrics.statsD.start.bat" %GATEWAY_IDENTIFIER%
```
cmebarrow: Why do we need the intermediate script? Can't we just call the metrics startup file directly, like this:
```
START CMD /C CALL "..\metrics.statsd\bin\metrics.statsD.start.bat" %GATEWAY_IDENTIFIER%
```
I tested this and it works provided you add the following line to the start of metrics.statsd\bin\metrics.statsD.start.bat:
`cd %~dp0`

distribution/src/main/gateway/bin/scripts/metrics.statsD.start
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
cmebarrow: I suggest removing the script and the .bat, as suggested in by comment on gateway.start.bat

distribution/src/main/gateway/bin/scripts/metrics.statsD.start.bat
```diff
@@ -0,0 +1,32 @@
+@REM
```
cmebarrow: Please rename this file to avoid the uppercase D - it's just likely to catch people out ( especially on Linux)

distribution/src/main/gateway/bin/gateway.start
```diff
@@ -56,5 +56,17 @@ fi
...
-exec java $GATEWAY_OPTS -Djava.library.path="$JAVA_LIBRARY_PATH" -XX:+HeapDumpOnOutOfMemoryError -cp "$GW_HOME/lib/*" org.kaazing.gateway.server.Main $*
+GW_ID=""
 
+# Verify if the gateway identifier was provided (required by Agorna multiple gateway instances)
+if [[ $GATEWAY_OPTS == *"org.kaazing.gateway.management.AGRONA_ENABLED=true"* ]]; then
+  if [ "$GATEWAY_IDENTIFIER" != "" ]; then
+    GW_ID="-Dorg.kaazing.gateway.server.GATEWAY_IDENTIFIER=$GATEWAY_IDENTIFIER"
+
+    # Startup the StatsD publisher
+    ./scripts/metrics.statsD.start $GATEWAY_IDENTIFIER &
```
cmebarrow: We can call ../metrics.statsd/metrics.statsd.start directly if we alter that script so it can be called from anywhere. I have proposed a change to allow that in my metrics pull request https://github.com/kaazing/metrics/pull/8

cmebarrow:
@irina-mitrea-luxoft I have done another pass. Please react to my comments (mainly change metrics dependency in pom.xml and remove the intermediate scripts - see also my pull request https://github.com/kaazing/metrics/pull/8)

irina-mitrea-luxoft:
@cmebarrow I have performed the changes mentioned apart from the pom.xml metrics dependency update. It seems as though the metrics release was performed in the staging repo, however I was unable to locate any other resulting artifacts.

irina-mitrea-luxoft:
@cmebarrow Updated metrics dependency version in order to highlight metrics release issue.

cmebarrow:
retest this please

cmebarrow:
@irina-mitrea-luxoft You are right, I missed the final step in releasing metrics (which is to manually release from the Sonatype staging repository), done now. I have kicked off another build. If it works I will merge this request.

cmebarrow:
@irina-mitrea-luxoft As per an email I just sent to Marius cc to you, following a discussion with John we do not want metrics to be included in the Gateway distribution. So please remove all changes to files under the distribution directory.

management/src/main/java/org/kaazing/gateway/management/monitoring/entity/manager/impl/ServiceSessionCounterManagerImpl.java
```diff
@@ -29,6 +29,7 @@
 public class ServiceSessionCounterManagerImpl implements
```
cmebarrow: Please rename this class to ServiceCounters so it can be used for all service-level counters as we go forward (it's worth looking at Aeron class [SystemCounters](https://github.com/real-logic/Aeron/blob/master/aeron-driver/src/main/java/uk/co/real_logic/aeron/driver/SystemCounters.java) which is similar - maybe you already saw it). A few easy changes will be enough to make it more general (rename incrementCounters to incrementSessionCounters and rename decrementCounters to decrementSessionCounters). I would also recommend simplifying the getXxxCounter methods by renaming them simply to xxxCounter(), following the pattern in SystemCounters, for example numberOfSessionsCounter instead of getNumberOfSessionsCounter, to make calling code more concise and simpler to read.

management/src/main/java/org/kaazing/gateway/management/monitoring/entity/manager/impl/ServiceSessionCounterManagerImpl.java
```diff
@@ -44,26 +45,28 @@
...
+            String serviceName, String gatewayId) {
         this.monitoringEntityFactory = monitoringEntityFactory;
         this.serviceName = serviceName;
+        this.gatewayId = gatewayId;
    }
 
     @Override
     public void initializeCounters() {
-        numberOfSessionsCounter = monitoringEntityFactory.makeLongMonitoringCounter(serviceName
+        numberOfSessionsCounter = monitoringEntityFactory.makeLongMonitoringCounter(gatewayId + SEPARATOR + serviceName
```
cmebarrow: Is it possible to get rid of the LongMonitoringCounter abstraction and just use Agrona's AtomicCounter (as in Aeron's SystemCounters class)? Not a big deal, but I'm curious to know why we need that extra abstraction.

pom.xml
cmebarrow: Please back out all changes to this file since the plan now is that we should not include metrics in the Gateway distribution, at least for now.

irina-mitrea-luxoft:
@cmebarrow The review changes have been implemented and the metrics integration removal has been performed.

irina-mitrea-luxoft:
Updated previous comment.

cmebarrow:
retest this please

irina-mitrea-luxoft:
@cmebarrow This has been tested under Linux and Windows.

sbadugu:
@irina-mitrea-luxoft "retest this please" is a shortcut phrase to rerun the build on Travis. I think Chris was trying to run the build. So in future, if you see a phrase like "retest this please" it's a command for travis to run the build.

cmebarrow:
Merging since the Gateway build failures are definitely unrelated.

PR #225, opened 2015-07-07
Fixed nits in http.proxy doc

veschup:
Per today's review with PM, made some nit fixes to the http.proxy documentation. Issue #357


dpwspoon:
There is a bug already filed for this failure.  As the bug is sporadic and these files are just markdown it should be fine to merge this

vjwang:
Thanks, @dpwspoon.

PR #230, opened 2015-07-09
Features/http revalidate

krismcqueen:
Pull request for getting code review feedback and for cleaning up outstanding issues.

Of note:  there is a TEMPORARY work around for setting authorization timeout using reflection.  There was not time to finish a generic setProperty() method to avoid the need for reflection and so that remains a TODO.

Another outstanding issue is use of TransportSession in the WsebSession.  It was exposed publicly and a few methods that formerly were available on AbstractWsBridgeSession were added to push http revalidate closer to completion.  This usage pattern deserves attention and proper implementation.


cmebarrow:
Closing because this will be replaced with a new pull request from me

PR #231, opened 2015-07-10
Configuration incorrect for ticket granting gateway #26

michaelcretzman:
changed udp:// to tcp://


vjwang:
@dpwspoon Since the coveralls check is minimal, is this OK to merge?

dpwspoon:
Yup! don't let the coveralls check block you unless its a a large percentage

PR #253, opened 2015-07-28
Added rfc7235 spec tests in gateway.

a-zuckut:
Relating to "Adding rfc7235 specification" pull request in K3PO.

*Should fail as waiting on publication of K3PO with new spec tests.


a-zuckut:
Checked with @jitsni and gateway does not support 407(Proxy Authentication) status codes. As this is in the spec tests, going to leave at least one of the tests in gateway IT just to let the reader know that gateway doesn't support 407 responses.

jitsni:
There is a new PR https://github.com/kaazing/gateway/pull/253 which includes this work. So closing this one.

PR #255, opened 2015-07-29
Changes for Geolocation Security

sanjay-saxena:
Add hooks in Gateway so that we can register enterprise-specific Callbacks used in enterprise-specific LoginModules. Propagate remote IP address from the accept side to the connect side when proxies are used. Updated test scripts with to include "X-Forwarded-For" HTTP header.

### DIFF
--- a/server/src/main/java/org/kaazing/gateway/server/context/resolve/GatewayContextResolver.java
+++ b/server/src/main/java/org/kaazing/gateway/server/context/resolve/GatewayContextResolver.java
@@ -127,6 +127,7 @@ public class GatewayContextResolver {
 
     private static final String SERVICE_TYPE_CLASS_PREFIX = "class:";
     private static final String LOGIN_MODULE_TYPE_CLASS_PREFIX = "class:";
+    private static final String EMPTY_STRING = "";
 
     // a map of file-extension to mime-type.  For backward compatibility, we'll
     // hardcode this initial set based on the values in Dragonfire HttpUtils.getContentType().
@@ -961,7 +962,12 @@ public class GatewayContextResolver {
                             Node node = childNodes.item(i);
                             if (Node.ELEMENT_NODE == node.getNodeType()) {
                                 NodeList content = node.getChildNodes();
-                                options.put(node.getLocalName(), content.item(0).getNodeValue());
+                                if ((content != null) && (content.item(0) != null)) {
+                                    options.put(node.getLocalName(), content.item(0).getNodeValue());
+                                }
+                                else {
+                                    options.put(node.getLocalName(), EMPTY_STRING);
+                                }
                             }
                         }
                     }
--- a/service/proxy/src/test/java/org/kaazing/gateway/service/proxy/ProxyServiceExtensionTest.java
+++ b/service/proxy/src/test/java/org/kaazing/gateway/service/proxy/ProxyServiceExtensionTest.java
@@ -90,7 +90,7 @@ public class ProxyServiceExtensionTest {
 
         // connect to the service to ensure the extension is executed
         try (Socket clientSocket = new Socket("localhost", 8888)) {
-            boolean success = latch.await(50, TimeUnit.SECONDS);
+            boolean success = latch.await(5, TimeUnit.SECONDS);
             assertTrue("Failed to execute all phases of proxy service extension", success);
         } catch (Exception ex) {
             fail("Unexpected exception in client connecting to server: " + ex);
--- a/service/proxy/src/test/java/org/kaazing/gateway/service/proxy/ProxyServiceExtensionTest.java
+++ b/service/proxy/src/test/java/org/kaazing/gateway/service/proxy/ProxyServiceExtensionTest.java
@@ -90,7 +90,7 @@ public class ProxyServiceExtensionTest {
 
         // connect to the service to ensure the extension is executed
         try (Socket clientSocket = new Socket("localhost", 8888)) {
-            boolean success = latch.await(5, TimeUnit.SECONDS);
+            boolean success = latch.await(50, TimeUnit.SECONDS);
             assertTrue("Failed to execute all phases of proxy service extension", success);
         } catch (Exception ex) {
             fail("Unexpected exception in client connecting to server: " + ex);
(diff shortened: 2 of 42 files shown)


jitsni:
Are we adding all the hops IP addresses to X-Forwarded-For ? for e.g. X-Forwarded-For: client ip, dmz gw ip, enterprise gw ip

It doesn't seem like the code handles multiple values since you are using setWriteHeader()  and getHeader() methods. The other thing is that any intermediary( like F5) could add to this list.

sanjay-saxena:
HttpAcceptor parses the value of `X-Forwarded-For` header and uses the first IP address in the list of comma separated IPs that can be augmented by the Intermediaries such as F5. Here is a [link](http://johannburkard.de/blog/programming/java/x-forwarded-for-http-header.html) that talks about using the first IP address in the comma separated list to determine the remote IP address.

I agree that it's a good idea to use `addWriteHeader()` instead of `setWriteHeader()`.

service/proxy/src/test/java/org/kaazing/gateway/service/proxy/ProxyServiceExtensionTest.java
```diff
@@ -90,7 +90,7 @@ public void run() {
 
         // connect to the service to ensure the extension is executed
         try (Socket clientSocket = new Socket("localhost", 8888)) {
-            boolean success = latch.await(5, TimeUnit.SECONDS);
+            boolean success = latch.await(50, TimeUnit.SECONDS);
```
jitsni: Do you really need 50 secs for this test or just just forgot to revert ?

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpAcceptor.java
```diff
@@ -593,6 +610,80 @@ public void removeBridgeFilters(IoFilterChain filterChain) {
...
     }
 
+    // If HTTP's X-FORWARDED-FOR header is set, then use it's value to determine the remote
+    // IP address. If X-FORWARDED-FOR header is NOT set, then use IoSession to determine
+    // the remote IP address.
+    private String getRemoteIpAddress(IoSession session, HttpRequestMessage httpRequest) {
+        String remoteIpAddr = httpRequest.getHeader(HEADER_X_FORWARDED_FOR);
+
+        if (remoteIpAddr == null) {
+            SocketAddress socketAddr = session.getRemoteAddress();
```
jitsni: The logic to get IP address walking down the session tree seems to be common (for e.g it is in proxy handler). Can you create a util method for that ?

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpAcceptor.java
```diff
@@ -593,6 +610,80 @@ public void removeBridgeFilters(IoFilterChain filterChain) {
...
+            }
+        }
+        else {
+            // X-FORWARDED-FOR HTTP header is inserted by proxies to identify the remote IP address.
+            // If the proxies are chained, then X-FORWARDED-FOR header can contain multiple IP
+            // addresses separated by commas. In such a scenario, the first IP address represents
+            // the remote IP address.
+            int index = remoteIpAddr.indexOf(',');
+            if (index > -1) {
+                remoteIpAddr = remoteIpAddr.substring(0, index).trim();
```
jitsni: HttpRequestMessage.getHeader() returns the first value if there are more values in the header. So no need to parse the header value

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpAcceptor.java
```diff
@@ -593,6 +610,80 @@ public void removeBridgeFilters(IoFilterChain filterChain) {
...
+        }
+
+        return null;
+    }
+
+    private static HttpSubjectSecurityFilter getSubjectSecurityFilter(Logger logger) {
+        Class<HttpSubjectSecurityFilter> clazz = HttpSubjectSecurityFilter.class;
+        ServiceLoader<HttpSubjectSecurityFilter> loader = ServiceLoader.load(clazz);
+
+        if (loader != null) {
```
jitsni: loader will not be null

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpAcceptor.java
```diff
@@ -593,6 +610,80 @@ public void removeBridgeFilters(IoFilterChain filterChain) {
...
+            if ((loader.iterator() != null) && loader.iterator().hasNext()) {
+                return loader.iterator().next();
```
jitsni: If the filter is found by service loader, how is the filter get hold of given logger ?

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpConnector.java
```diff
@@ -357,6 +358,25 @@ public void initializeSession(IoSession session, T future) {
         };
     }
 
+    private IoSession getTcpSession(IoSession session) {
```
jitsni: Since this logic is common in many places, just use a util function.

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpConnector.java
```diff
@@ -386,6 +406,19 @@ public DefaultHttpSession call() throws Exception {
                                 parentEx,
                                 new HttpBufferAllocator(parentAllocator));
                         parent.setAttribute(HTTP_SESSION_KEY, httpSession);
+
+                        // Set X-FORWARDED-FOR header on the HTTP session.
```
jitsni: I think the following could be wrong. This would add the IP address of the origin server. Run any of the http proxy tests and check the port to verify this. IIRC, the httpsession is created even before it sent the http request to the origin server (and in this case tcp address would origin server's address).

server/src/main/java/org/kaazing/gateway/server/context/resolve/GatewayContextResolver.java
```diff
@@ -961,7 +962,12 @@ private RealmsContext resolveRealms(SecurityType securityConfig, SecurityContext
                             Node node = childNodes.item(i);
                             if (Node.ELEMENT_NODE == node.getNodeType()) {
                                 NodeList content = node.getChildNodes();
-                                options.put(node.getLocalName(), content.item(0).getNodeValue());
+                                if ((content != null) && (content.item(0) != null)) {
+                                    options.put(node.getLocalName(), content.item(0).getNodeValue());
+                                }
+                                else {
+                                    options.put(node.getLocalName(), EMPTY_STRING);
```
jitsni: Which are the options that getting initialized with empty string ?

sanjay-saxena:
Closing this PR as I would like to start with a clean slate based on our latest discussion.

PR #269, opened 2015-08-12
Versioning jar files and detecting duplicate jar files

alingherasim:
This pull request contains the versioning of the jar files and an implementation that does not allow the gateway to start if duplicate jar files are loaded by the class loader.


bridge/gateway.bridge/pom.xml
```diff
@@ -36,6 +36,20 @@
...
         <plugins>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Implementation-Title>${project.name}</Implementation-Title>
+                            <Implementation-Version>${project.version}</Implementation-Version>
+                            <Jar-Name>${artifactId}</Jar-Name>
```
dpwspoon: Why not use the {artifactId} for the Kaazing-Product category as well? You may then be able to use inheritance to minimize the number of plugins that you need, i.e. just put this in the top level pom (maybe, not sure how inheritance works it the submodule is not of packaging jar

server/src/main/java/org/kaazing/gateway/server/util/version/DuplicateJarFinder.java
```diff
@@ -0,0 +1,113 @@
...
+import org.slf4j.Logger;
+
+/**
+ * Class used for detecting if duplicate jar files are loaded by the gateway.
+ */
+public class DuplicateJarFinder {
+
+    private static final String JAR_FILE_WITH_VERSION_LOGGING_MESSAGE = "The jar file {} with the version {} was loaded.";
+    private static final String CONFLICTING_JARS_LOGGING_MESSAGE =
+            "Failed to start because of conflicting versions of artifact on the classpath: {}.jar.";
```
dpwspoon: @robinzimmermann, Since this string is driven by your requirements, are you alright with changing "conflicting" to "duplicate"

@alingherasim, Robin is out right now so don't let this block you

server/src/main/java/org/kaazing/gateway/server/util/version/DuplicateJarFinder.java
```diff
@@ -0,0 +1,113 @@
...
+        for (String classPathEntry : classPathEntries) {
+            parseManifestFileFromClassPathEntry(classPathEntry);
+        }
+        logErrorForDuplicateJars();
+        throwExceptionIfDuplicateJarsAreFound();
+    }
+
+    private void parseManifestFileFromClassPathEntry(String classPathEntry) throws IOException {
+        Attributes manifestAttributes = classPathParser.getManifestAttributesFromClassPathEntry(classPathEntry);
+        String product = manifestAttributes.getValue(MANIFEST_PRODUCT);
```
dpwspoon: I don't immediately see the need to for having to check whether it is a Kaazing Product or not, this would also make your like simpler with the jar plugin.  Is there some jars that you hit that we have multiple of in testing?  

Another option would be to put the group Id and artifact Id in the jar name, and then check if the group Id or artifact Id has "kaazing" in it.  In fact after thinking about this, I think this would be better then additing a Kaazing-Product entry to the manifest.  The reason being that jars in the lib directory have com.kaazing or org.kaazing in front of them so it will be easier for customers to identify, and there may be a org.kaazing and com.kaazing jar that has the same artifactId

server/src/test/java/org/kaazing/gateway/server/util/version/DuplicateJarFinderTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @Before
+    public void setUp() {
+        context = new Mockery();
+
+        context.setImposteriser(ClassImposteriser.INSTANCE);
+    }
+
+    @Test
+    public void testFindDuplicateJarsShouldNotThrowExceptionIfOneKaazingProduct() throws IOException {
```
dpwspoon: Awesome

server/src/main/java/org/kaazing/gateway/server/util/version/DuplicateJarFinder.java
alingherasim: There was a situation where my local build added another version of the log4j library from the sigar distribution so I wanted to be guarded in this case.

dpwspoon:
See pull request to your branch.  I've basically added the feedback there.  I managed to get the jar plugin to work with inheritance by adding it in the plugin management section.  When you run `mvn help:effective-pom` in a submodule you can see the properties are inherited.

dpwspoon:
closing in favor of kaazing/gateway#276 which consumes these checkins

PR #271, opened 2015-08-12
Revert "Add wsn spec tests"

dpwspoon:
Reverts kaazing/gateway#270


cmebarrow:
@dpwspoon Why do you need to revert kaazing/gateway#270 ?

cmebarrow:
Closing without merging because Jitu is going to add a missing file to fix the previous pull (#270)

PR #275, opened 2015-08-13
Added support for the new MMF format

irina-mitrea-luxoft:
This pull request contains a set of changes which were implemented as WIP for supporting the new MMF format.
Listing some of the items still pending:
- Code cleaning and javadocs
- Unit testing
- Integration with the reader -- in progress


management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/impl/DefaultMMFMonitoringDataManagerStub.java
```diff
@@ -0,0 +1,58 @@
...
+import org.kaazing.gateway.management.monitoring.configuration.MonitoringDataManager;
+import org.kaazing.gateway.management.monitoring.entity.manager.impl.ServiceCounterManagerImpl;
+import org.kaazing.gateway.management.monitoring.service.MonitoredService;
+import org.kaazing.gateway.service.DefaultLongMonitoringCounterStub;
+import org.kaazing.gateway.service.LongMonitoringCounter;
+import org.kaazing.gateway.service.MonitoringEntityFactory;
+
+public class DefaultMMFMonitoringDataManagerStub implements MonitoringDataManager {
+
+    private MonitoringEntityFactory monitorigEntityFactory = new ServiceCounterManagerImpl(new MonitoringEntityFactory() {
```
cmebarrow: Typo: rename to monitoringEntityFactory

cmebarrow:
I would recommend considering renaming MonitoringEntityFactory to MonitoringFactory for brevity.

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/MonitoringDataManagerInjector.java
```diff
@@ -21,18 +21,16 @@
...
+public interface MonitoringDataManagerInjector {
 
     /**
-     * Creates the monitoring entity factory
-     * @return the monitoring entity factory
+     * Creates the monitoring data manager
+     * @return
      */
-    MonitoringEntityFactory makeMonitoringEntityFactory();
+    MonitoringDataManager makeMonitoringDataManager();
```
cmebarrow: I can't find anywhere where this method is called. If it is dead code please remove this interface and the implementations.

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/impl/DefaultMMFMonitoringDataManagerStub.java
```diff
@@ -0,0 +1,58 @@
+/**
```
cmebarrow: This class is just an empty implementation of MonitoringDataManager and has no MMF awareness so it can just be called MonitoringDataManagerStub.

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/impl/MMFMonitoringDataManager.java
```diff
@@ -23,84 +23,78 @@
 
 import java.io.File;
```
cmebarrow: This looks like dead code too.

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/impl/MonitorFileWriterImpl.java
```diff
@@ -0,0 +1,316 @@
...
+ * +-----------------------------------------------------------------------+
+ * Metadata length: NUMBER_OF_INTS_IN_HEADER * BitUtil.SIZE_OF_INT + SIZEOF_STRING +
+ * servicesCount * (SIZEOF_STRING + NUMBER_OF_INTS_PER_SERVICE * BitUtil.SIZE_OF_INT)
+ */
+public final class MonitorFileWriterImpl implements MonitorFileWriter {
+    private static final int OFFSETS_PER_SERVICE = 4;
+    private static final int NUMBER_OF_INTS_PER_SERVICE = 5;
+    private static final int NUMBER_OF_INTS_IN_HEADER = 8;
+    private static final int SIZEOF_STRING = 128;
+    private static final int MAX_SERVICE_COUNT = 10;
```
cmebarrow: 10 services is a bit limiting. I would recommend 100, or set it based on the actual number of services plus some margin (like 10).

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/impl/MonitorFileWriterImpl.java
```diff
@@ -0,0 +1,316 @@
...
+import org.kaazing.gateway.management.monitoring.writer.impl.MMFGatewayWriter;
+import org.kaazing.gateway.management.monitoring.writer.impl.MMFSeviceWriter;
+import org.kaazing.gateway.service.MonitoringEntityFactory;
+
+import uk.co.real_logic.agrona.BitUtil;
+import uk.co.real_logic.agrona.DirectBuffer;
+import uk.co.real_logic.agrona.concurrent.UnsafeBuffer;
+
+/**
+ * Class responsible with storing/writing information regarding the MMF format.
```
cmebarrow: responsible for

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/MonitoringDataManagerInjector.java
irina-mitrea-luxoft: This is called from ManagementGatewayListener->startingGateway() method.

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/impl/DefaultMMFMonitoringDataManagerStub.java
irina-mitrea-luxoft: As described above, MonitoringDataManagerInjectorImpl is used from ManagementGatewayListener->startingGateway() method.

cmebarrow: Oh I see, sorry missed that

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/impl/DefaultMMFMonitoringDataManagerStub.java
```diff
@@ -0,0 +1,58 @@
...
+import org.kaazing.gateway.service.LongMonitoringCounter;
+import org.kaazing.gateway.service.MonitoringEntityFactory;
+
+public class DefaultMMFMonitoringDataManagerStub implements MonitoringDataManager {
+
+    private MonitoringEntityFactory monitorigEntityFactory = new ServiceCounterManagerImpl(new MonitoringEntityFactory() {
+
+        @Override
+        public LongMonitoringCounter makeLongMonitoringCounter(String name) {
+            return new DefaultLongMonitoringCounterStub();
```
cmebarrow: I would suggest using a static final member variable to avoid creating an object on each call to this method.

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/impl/DefaultMMFMonitoringDataManagerStub.java
```diff
@@ -0,0 +1,61 @@
...
+package org.kaazing.gateway.management.monitoring.configuration.impl;
+
+import org.kaazing.gateway.management.monitoring.configuration.MonitoringDataManager;
+import org.kaazing.gateway.management.monitoring.entity.manager.impl.ServiceCounterManagerImpl;
+import org.kaazing.gateway.management.monitoring.service.MonitoredService;
+import org.kaazing.gateway.service.DefaultLongMonitoringCounterStub;
+import org.kaazing.gateway.service.LongMonitoringCounter;
+import org.kaazing.gateway.service.MonitoringEntityFactory;
+
+public class DefaultMMFMonitoringDataManagerStub implements MonitoringDataManager {
```
cmebarrow: This class is just a no-op implementation of MonitoringDataManager. I would therefore drop MMF from the name since it has nothing to do with MMF. Also, going forward it won't be the default (we will want gathering statistics in MMF to be the default, see metrics configuration story), so I would recommend also removing the word Default from the name. That leaves us with the name MonitoringDataManagerStub. You could also perhaps consider the name NoopMonitoringDataManager or DoNothingMonitoringDataManager.

management/src/main/java/org/kaazing/gateway/management/filter/ManagementFilter.java
```diff
@@ -57,18 +51,11 @@
...
         this.serviceBean = serviceBean;
         this.managementContext = serviceBean.getGatewayManagementBean().getManagementContext();
         this.serviceContext = serviceBean.getServiceContext();
-        String gatewayId = InternalSystemProperty.GATEWAY_IDENTIFIER.getProperty(configuration);
-
-        CounterManagerFactory counterFactory = new CounterManagerFactoryImpl();
-        serviceCounterManager = counterFactory.makeServiceCounterManager(monitoringEntityFactory,
-                serviceName, gatewayId);
+        serviceCounterManager = (ServiceCounterManager) serviceContext.getMonitoringFactory();
         serviceCounterManager.initializeSessionCounters();
```
cmebarrow: There is a single instance of ServiceCounterManager per service but there is a new instance of ManagementFilter per client connection (IoSession). The ServiceCounterManager instance is therefore shared across all Client connections, and must be able to operate correctly when used simultaneously by multiple threads. Looking at the implementation of initializeSessionCounters it should be called only once per instance of ServiceCounterManager, so I don't think you can call it for each new instance of the filter. It needs to be called once only during service startup, One possible solution would be to get rid of the ServiceCounterManager.initializeSessionCounters method altogether and just do the counter initialization logic in the ServiceCounterManagerImpl constructor.

management/src/main/java/org/kaazing/gateway/management/context/ManagementGatewayListener.java
```diff
@@ -69,8 +92,8 @@ public void startingService(ServiceContext serviceContext) {
     }
 
     @Override
-    public void stoppedGateway() {
```
cmebarrow: Looks as if you removed the stoppedGateway method. Are you sure we don't need it?

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/MonitorFileWriter.java
```diff
@@ -0,0 +1,103 @@
...
+     */
+    UnsafeBuffer createServiceCounterValuesBuffer(ByteBuffer buffer, int index);
+
+    /**
+     * Method returning gateway monitoring entity factory
+     * @param mappedMonitorFile
+     * @param monitoringDir
+     * @return
+     */
+    MonitoringEntityFactory getGwMonitoringEntityFactory(MappedByteBuffer mappedMonitorFile, File monitoringDir);
```
cmebarrow: Please rename to getGatewayMonitoringEntityFactory for consistency with other methods here and in the rest of the gateway code base (we don't normally use abbreviations like Gw).

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/MonitorFileWriter.java
```diff
@@ -0,0 +1,103 @@
...
+     * @param buffer
+     * @param index
+     * @return
+     */
+    UnsafeBuffer createServiceCounterValuesBuffer(ByteBuffer buffer, int index);
+
+    /**
+     * Method returning gateway monitoring entity factory
+     * @param mappedMonitorFile
+     * @param monitoringDir
```
cmebarrow: The monitoringDir parameter to this method and the next method is only needed so it can be passed through the MMFSeviceWriter instance into the AgronaMonitoringEntityFactory instance that is created for the gateway counters and for each service. And the only way monitoringDir  is used is in AgronaMonitoringEntityFactory.close() which actually deletes the MMF and monitoring directory. This will not work because we must not delete the MMF and monitoring directory until all services have been shut down during Gateway shutdown. I would strongly recommend removing the monitoringDir parameter to this method and the following method and make it the responsibility of the code external to MonitorFileWriter to manage the complete life cycle of the MMF and monitoring directory, in other words, that code should be in charge of deleting those objects as well as creating them. A suitable place to put the logic might be In the stoppedGateway method in ManagementGatewayListener.

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/impl/MonitorFileWriterImpl.java
```diff
@@ -0,0 +1,321 @@
...
+
+    private static final int MONITOR_VERSION = 1;
+    private static final int MONITOR_VERSION_OFFSET = 0;
+    private static final int GW_DATA_REFERENCE_OFFSET = MONITOR_VERSION_OFFSET + BitUtil.SIZE_OF_INT;
+    private static final int SERVICE_DATA_REFERENCE_OFFSET = GW_DATA_REFERENCE_OFFSET + BitUtil.SIZE_OF_INT;
+
+    private static final int GW_ID_OFFSET = SERVICE_DATA_REFERENCE_OFFSET + BitUtil.SIZE_OF_INT;
+    private static final int GW_DATA_OFFSET = GW_ID_OFFSET;
+
+    private static final int GATEWAY_COUNTER_VALUES_BUFFER_LENGTH = 1024 * 128;
```
cmebarrow: Where is the number 1024 coming from?

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/impl/MonitorFileWriterImpl.java
```diff
@@ -0,0 +1,321 @@
...
+    /**
+     * Creates the counter values buffer for the service identifier by index
+     * @param buffer - the underlying byte buffer
+     * @param metaDataBuffer - the meta data buffer from which we compute the offset
+     * @param index - service identifier
+     * @return the counter values buffer
+     */
+    @Override
+    public UnsafeBuffer createServiceCounterValuesBuffer(final ByteBuffer buffer,
+                                                         int index) {
```
cmebarrow: Just to check my understanding, index is the zero-based service index, 0 for the first service, 1 for the next, etc, right?

management/src/main/java/org/kaazing/gateway/management/monitoring/writer/impl/MMFSeviceWriter.java
```diff
@@ -0,0 +1,71 @@
+/**
```
cmebarrow: Typo: the name of this class should be MMFServiceWriter (by the way, I expect you already know this, but just in case, it's really easy to rename methods or classes in Eclipse by right clicking on the name and choosing menu option Refactor).

management/src/test/java/org/kaazing/gateway/management/monitoring/configuration/impl/MonitoringEntityFactoryInjectorTest.java
```diff
@@ -37,29 +31,29 @@
 
     @Test
     public void testMonitoringEntityFactoryInjectorAgronaEnabledUnset() {
-        MonitoringEntityFactoryInjector injector = new MonitoringEntityFactoryInjectorImpl(new Properties());
-        MonitoringEntityFactory monitoringEntityFactory = injector.makeMonitoringEntityFactory();
-        monitoringEntityFactory.close();
-        assertTrue(monitoringEntityFactory instanceof DefaultMonitoringEntityFactoryStub);
+//        MonitoringEntityFactoryInjector injector = new MonitoringEntityFactoryInjectorImpl(new Properties());
```
cmebarrow: I presume this is still a work in progress

cmebarrow:
@irina-mitrea-luxoft I have finished reviewing the main code. I have not looked much at the tests since it is still a work in progress. There are a couple of comments that are important and definitely need addressing, plus some more minor cosmetic ones. Let me know if you have any questions. Thanks, Chris.

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/impl/MonitorFileWriterImpl.java
irina-mitrea-luxoft: 1024 represents the number of entries in the counter manager buffers.

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/impl/MonitorFileWriterImpl.java
irina-mitrea-luxoft: Yes, that is correct.

management/src/test/java/org/kaazing/gateway/management/monitoring/configuration/impl/MonitoringEntityFactoryInjectorTest.java
irina-mitrea-luxoft: Unit tests have not been addressed yet.

management/src/main/java/org/kaazing/gateway/management/context/ManagementGatewayListener.java
irina-mitrea-luxoft: The ManagementGatewayListener initially contained the following code:
@Override
public void stoppedGateway() {
	managementContext.close();
}
In DefaultManagementContext (implementation of ManagementContext), the close method performed a close on the monitoringEntityFactory.

However, due to the fact that in the ManagementGatewayListener the following method was added (performing a cleanup on the monitoringEntityFactory):
@Override
public void stoppedService(ServiceContext serviceContext) {
	serviceContext.getMonitoringFactory().close();
}
I have removed the stoppedGateway method in ManagementGatewayListener and the close method from ManagementContext (and its implementation, DefaultManagementContext).

Additional information on this can be obtained from ManagementGatewayListener, ManagementContext and DefaultManagementContext Git logs. Please give me your feedback on this.

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/MonitorFileWriter.java
irina-mitrea-luxoft: Just checked in a fix for this issue.

irina-mitrea-luxoft: Thank you, just checked in a fix for this issue.

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/impl/MonitorFileWriterImpl.java
cmebarrow: Wow, that's pretty big isn't it? I can't imagine us having that many gateway or service counters. I would think we could reduce it to say 50. Where did 1024 come from? Whatever then we have please make sure this unit test to prove that we will throw suitable exception if the value is exceeded (i.e. some service tries to add more counters that we can support).

cmebarrow:
@irina-mitrea-luxoft Looking good. Let me know when you are done with the test work so I can review the tests and merge.

service/spi/src/main/java/org/kaazing/gateway/service/DefaultLongMonitoringCounterStub.java
```diff
@@ -19,9 +19,9 @@
  * under the License.
```
cmebarrow: This file should not have been moved out of gateway.management. It does not belong in gateway.service because that represents the SPI for creating services. Services should not need to use implementation classes like this one.

management/src/test/java/org/kaazing/gateway/management/monitoring/service/impl/MonitoredServiceImplTest.java
```diff
@@ -0,0 +1,93 @@
...
+        assertTrue(service1.equals(service3));
+        assertTrue(!service1.equals(service2));
+    }
+
+    /**
+     * Method instantiating a new service context
+     * @param serviceName 
+     * @return
+     */
+    private DefaultServiceContext createDefaultServiceContext(String serviceName) {
```
cmebarrow: Why don't you use mock objects for ServiceContext? Seems it would be easier and usually we want unit tests to test one class not use other classes

management/src/test/java/org/kaazing/gateway/management/monitoring/configuration/impl/MonitorFileWriterImplTest.java
```diff
@@ -0,0 +1,56 @@
...
+                monitoringDirName = LINUX_DEV_SHM_DIRECTORY + monitoringDirName;
+            }
+        }
+
+        MonitorFileWriter monitorFileWriter = new MonitorFileWriterImpl(MONITOR_FILE_NAME);
+        File monitoringFile = new File(monitoringDirName, MONITOR_FILE_NAME);
+        monitorFileWriter.initialize(monitoringFile);
+        monitorFileWriter.close(monitoringFile);
+    }
+}
```
cmebarrow: I presume you will be adding more test methods here?

management/src/test/java/org/kaazing/gateway/management/monitoring/configuration/impl/MMFMonitoringDataManagerTest.java
```diff
@@ -0,0 +1,99 @@
...
+        monitoringDataManager.close();
+    }
+
+    @Test
+    public void addService() {
+        MMFMonitoringDataManager monitoringDataManager = new MMFMonitoringDataManager(MONITORING_FILE);
+        assertNotNull(monitoringDataManager);
+        MonitoringEntityFactory monitoringEntityFactory = monitoringDataManager.initialize();
+        assertNotNull(monitoringEntityFactory);
+        MonitoredService monitoredService = new MonitoredServiceImpl(createDefaultServiceContext("serviceName"));
```
cmebarrow: Please use mock objects for ServiceContext. Usually we want unit tests to test one class not use other classes, plus using mock objecst lets you verify the class under test is making the expected method calls to it.

management/src/main/java/org/kaazing/gateway/management/monitoring/configuration/impl/MMFMonitoringDataManager.java
```diff
@@ -0,0 +1,122 @@
...
+public class MMFMonitoringDataManager implements MonitoringDataManager {
+
+    private static final String LINUX = "Linux";
+    private static final String OS_NAME_SYSTEM_PROPERTY = "os.name";
+    private static final String LINUX_DEV_SHM_DIRECTORY = "/dev/shm";
+    private static final String MONITOR_DIR_NAME = "/kaazing";
+
+    private MonitorFileWriter monitorFileWriter;
+    private File monitoringDir;
+    private Collection<MonitoredService> services = new HashSet<>();
```
cmebarrow: This member variable is only being used to track how many services have been added. So it would be more efficient just to replace it with int serviceCount, starting at 0 and incremented each time addService is called.

management/src/main/java/org/kaazing/gateway/management/context/ManagementGatewayListener.java
```diff
@@ -69,8 +92,13 @@ public void startingService(ServiceContext serviceContext) {
     }
 
     @Override
+    public void stoppedService(ServiceContext serviceContext) {
+        serviceContext.getMonitoringFactory().close();
```
cmebarrow: We shouldn't be closing and freeing the counters when a service is stopped because it could be started again. This should be done only when the service is destroyed. All you need to do to fix this is rename this method to "destroyedService".

cmebarrow:
@irina-mitrea-luxoft Just reviewed all the tests and made some comments spotted a couple of things in the code too. It's getting really close. Please try to wrap everything up tomorrow if possible. I really want to get this merged (and I am sure you do too!). thanks,Chris

irina-mitrea-luxoft:
@cmebarrow Hello, Chis. Please give me your feedback on the latest changes. Thank you, Irina.

cmebarrow:
Pull request https://github.com/kaazing/metrics-demo/pull/10 with the required changes has now been merged.

PR #276, opened 2015-08-17
Adds jar version to manifest entry such that it can be reported when duplicate jars are

dpwspoon:
on the classpath, or for debug information


server/src/main/java/org/kaazing/gateway/server/impl/GatewayImpl.java
```diff
@@ -81,6 +82,7 @@
     private Launcher gateway;
     private Gateway baseGateway;
     private KaazingFileWatchdog watchDog;
+    private DuplicateJarFinder duplicateJarFinder;
```
jitsni: can be final field

server/src/main/java/org/kaazing/gateway/server/util/version/ClassPathParser.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.jar.Attributes;
+import java.util.jar.JarFile;
+import java.util.jar.Manifest;
+
+/**
+ * Class used for parsing the class path system property.
+ */
+public class ClassPathParser {
+
+    public ClassPathParser() {
```
jitsni: Do we need the constructor ?

server/src/main/java/org/kaazing/gateway/server/util/version/ClassPathParser.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /**
+     * Retrieves the class path entries.
+     * @return the class path entries
+     */
+    public String[] getClassPathEntries() {
+        return System.getProperty("java.class.path").split(System.getProperty("path.separator"));
```
jitsni: Rather than using classpath system property, I think loading of MANIFEST.MF resources using the classloader would be better.

server/src/main/java/org/kaazing/gateway/server/util/version/DuplicateJarFinder.java
```diff
@@ -0,0 +1,113 @@
...
+ * Class used for detecting if duplicate jar files are loaded by the gateway.
+ */
+public class DuplicateJarFinder {
+
+    private static final String JAR_FILE_WITH_VERSION_LOGGING_MESSAGE = "The jar file {} with the version {} was loaded.";
+    private static final String CONFLICTING_JARS_LOGGING_MESSAGE =
+            "Failed to start because of conflicting versions of artifact on the classpath: {}.jar.";
+    private static final String MANIFEST_VERSION = "Implementation-Version";
+    private static final String MANIFEST_JAR_NAME = "Artifact-Name";
+    private static final String MANIFEST_PRODUCT = "Kaazing-Product";
```
jitsni: Are we using this MANIFEST_PRODUCT ? If so, what's the purpose of this attribute ?

server/src/test/java/org/kaazing/gateway/server/util/version/DuplicateJarFinderTest.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+    
+    private Attributes getAttributesForKaazingProduct() {
+        Attributes attributes = new Attributes();
+        attributes.putValue("Implementation-Version", "1.0");
+        attributes.putValue("Artifact-Name", MOCK_JAR_FILE_NAME);
+        return attributes;
+    }
+
+    private Attributes getAttributesForNoneKaazingProduct() {
```
jitsni: getAttributesForKaazingProduct() and getAttributesForNoneKaazingProduct() seem to do the same thing

server/src/test/java/org/kaazing/gateway/server/util/version/DuplicateJarFinderTest.java
```diff
@@ -0,0 +1,114 @@
...
+            will(returnValue(MOCK_CLASS_PATH_ENTRIES));
+            oneOf(classPathParser).getManifestAttributesFromClassPathEntry(MOCK_CLASS_PATH_ENTRIES[0]);
+            will(returnValue(getAttributesForKaazingProduct()));
+            allowing(gatewayLogger).debug(with(any(String.class)), with(any(Object.class)), with(any(Object.class)));
+            never(gatewayLogger).error(with(any(String.class)), with(any(Object.class)));
+        }});
+
+        try {
+            duplicateJarFinder.findDuplicateJars();
+        } catch (Exception e) {
```
jitsni: I would just let the exception propagate and let the junit test fail with that exception

server/src/main/java/org/kaazing/gateway/server/util/version/DuplicateJarFinder.java
```diff
@@ -0,0 +1,113 @@
...
+
+    private ClassPathParser classPathParser;
+    private Logger gatewayLogger;
+    private Set<String> duplicateJars = new HashSet<String>();
+    private Set<String> loadedJars = new HashSet<String>();
```
jitsni: I would just keep one Map<jar, List<version string>> field and just loop at the end print *all* the duplicate jars with their versions.

server/src/test/java/org/kaazing/gateway/server/util/version/DuplicateJarFinderTest.java
dpwspoon: resolved

server/src/main/java/org/kaazing/gateway/server/util/version/DuplicateJarFinder.java
dpwspoon: Not sure if I follow this one, there is a set for duplicate jars and loaded jars.  While the implementation of parseManifestFileFromClassPathEntry could use a map I see nothing wrong with it as it is now.  Am I missing something...?  Are you asking for the version to be kept in a map

dpwspoon:
@jitsni Thanks for the feedback.  I've applied changes with one question for clarification

server/src/main/java/org/kaazing/gateway/server/util/version/ClassPathParser.java
```diff
@@ -0,0 +1,55 @@
...
+import java.io.IOException;
+import java.net.URL;
+import java.util.Enumeration;
+import java.util.jar.Attributes;
+import java.util.jar.Manifest;
+
+/**
+ * Class used for parsing the class path system property.
+ */
+public class ClassPathParser {
```
jitsni: You may want to change the file name and the comment as it no longer parses classpath property

server/src/main/java/org/kaazing/gateway/server/util/version/DuplicateJarFinder.java
```diff
@@ -0,0 +1,107 @@
...
+import org.slf4j.Logger;
+
+/**
+ * Class used for detecting if duplicate jar files are loaded by the gateway.
+ */
+public class DuplicateJarFinder {
+
+    private static final String JAR_FILE_WITH_VERSION_LOGGING_MESSAGE = "The jar file {} with the version {} was loaded.";
+    private static final String CONFLICTING_JARS_LOGGING_MESSAGE =
+            "Failed to start because of conflicting versions of artifact on the classpath: {}.jar.";
```
jitsni: Do you want to give the versions also in the log message ?

dpwspoon: Not particularly since the versions get printed out prior to this

dpwspoon: I guess those are in debug, so yeah we can do it

dpwspoon:
@jitsni ready for merge?

PR #279, opened 2015-08-18
Minor Cleanup and 3 New Tests

justinma246:


service/http.directory/src/main/java/org/kaazing/gateway/service/http/directory/HttpDirectoryServiceHandler.java
```diff
@@ -148,9 +149,19 @@ protected void doSessionOpened(HttpAcceptSession session) throws Exception {
         }
 
         if (!underBaseDir) {
-            reportError(session, HttpStatus.CLIENT_BAD_REQUEST);
-            session.close(false);
-            return;
+        	boolean symbolicChecker = Files.isSymbolicLink(requestFile.toPath());
+        	if(symbolicChecker)
+        	{
```
dpwspoon: It should actually always be a 403 or a 404.  I would recommend always returning a 404

```
10.4.4 403 Forbidden

The server understood the request, but is refusing to fulfill it. Authorization will not help and the request SHOULD NOT be repeated. If the request method was not HEAD and the server wishes to make public why the request has not been fulfilled, it SHOULD describe the reason for the refusal in the entity. If the server does not wish to make this information available to the client, the status code 404 (Not Found) can be used instead.

10.4.5 404 Not Found

The server has not found anything matching the Request-URI. No indication is given of whether the condition is temporary or permanent. The 410 (Gone) status code SHOULD be used if the server knows, through some internally configurable mechanism, that an old resource is permanently unavailable and has no forwarding address. This status code is commonly used when the server does not wish to reveal exactly why the request has been refused, or when no other response is applicable.

```

service/http.directory/src/test/webapp/test.rtf
```diff
@@ -0,0 +1,7 @@
+{\rtf1\ansi\ansicpg1252\cocoartf1348\cocoasubrtf170
```
dpwspoon: Can we make these files dead simple, like just hello world

service/http.directory/src/test/webapp/symlink/index.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.01//EN" "http://www.w3.org/TR/html4/strict.dtd">
+<html>
```
dpwspoon: Can we make these files dead simple, like just hello world

service/http.directory/src/main/java/org/kaazing/gateway/service/http/directory/HttpDirectoryServiceHandler.java
```diff
@@ -148,9 +149,19 @@ protected void doSessionOpened(HttpAcceptSession session) throws Exception {
...
+        		reportError(session, HttpStatus.CLIENT_NOT_FOUND);
```
dpwspoon: The formatting here looks funky, can we use the kaazing formatter @ https://github.com/kaazing/code.quality/blob/develop/src/main/resources/org/kaazing/code/quality/eclipse/CodeFormatterProfile.xml

dpwspoon:
Awesome! Thanks Justin

service/http.directory/src/main/java/org/kaazing/gateway/service/http/directory/HttpDirectoryServiceHandler.java
```diff
@@ -148,9 +149,9 @@ protected void doSessionOpened(HttpAcceptSession session) throws Exception {
         }
 
         if (!underBaseDir) {
-            reportError(session, HttpStatus.CLIENT_BAD_REQUEST);
-            session.close(false);
-            return;
+        	reportError(session, HttpStatus.CLIENT_NOT_FOUND);
```
dpwspoon: please no tabs

PR #280, opened 2015-08-18
Binding explicitly for .../;api endpoints for wsn and wseb cases

jitsni:
- HttpOperationFilter is removed in http layer when the request is httpxe
- /;api requests with X-Origin are elevated to httpxe
- HttpRequestMessage.isComplete() returns correctly if there is data
in the request


jitsni:
Forward porting changeset

PR #283, opened 2015-08-19
Adding info message when wsn is falling back to wse

jitsni:
Forward porting


jitsni:
forward porting from 4.0.x.y

Issue #282, opened 2015-08-18
TravisCI build tests hanging


dpwspoon:
For [example](https://api.travis-ci.org/jobs/76187608/log.txt?deansi=true).  

This one hung on Running org.kaazing.gateway.transport.wsn.WsnBindingsTest
But I also saw it hang on WsebAcceptorTest.java, and one in mina.netty

PR #286, opened 2015-08-19
Travis ci failures

dpwspoon:
resolves https://github.com/kaazing/gateway/issues/282


mina.core/core/src/test/java/org/apache/mina/transport/AbstractFileRegionTest.java
```diff
@@ -49,7 +49,10 @@
     protected abstract IoAcceptor createAcceptor();
     protected abstract IoConnector createConnector();
 
+    // Ignored for https://github.com/kaazing/gateway/issues/221
+    // Not sure how to ignore in junit.framework with inheritance
```
cmebarrow: I think you can just put @Ignore on this method. Did you try that?

transport/wsr/src/test/java/org/kaazing/gateway/transport/wsr/WsrAcceptorTest.java
```diff
@@ -29,12 +29,18 @@
 import java.util.Collections;
```
cmebarrow: Please just remove this file altogether since we are no longer supporting wsr. You can close the bug (issue #287) too.

cmebarrow:
@dpwspoon Done reviewing, couple of minor suggested changes

mina.core/core/src/test/java/org/apache/mina/transport/AbstractFileRegionTest.java
dpwspoon: Yeah, it looked like the test still ran.  Ignore was org.junit, but I couldn't find a junit.framework package equivilant

transport/wsr/src/test/java/org/kaazing/gateway/transport/wsr/WsrAcceptorTest.java
dpwspoon: Removed file but #287 also tracks wse bind failures so that has to stay open :-(

dpwspoon:
@cmebarrow Applied feedback, thanks

Issue #267, opened 2015-08-11
OcspIT.testGoodCertificate fails locally


justinma246:
Output is 

```
Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.397 sec <<< FAILURE! - in org.kaazing.gateway.transport.ssl.OcspIT
testGoodCertificate(org.kaazing.gateway.transport.ssl.OcspIT)  Time elapsed: 0.125 sec  <<< ERROR!
java.net.SocketException: Broken pipe
	at java.net.SocketOutputStream.socketWrite0(Native Method)
	at java.net.SocketOutputStream.socketWrite(SocketOutputStream.java:109)
	at java.net.SocketOutputStream.write(SocketOutputStream.java:153)
	at sun.security.ssl.OutputRecord.writeBuffer(OutputRecord.java:431)
	at sun.security.ssl.OutputRecord.write(OutputRecord.java:417)
	at sun.security.ssl.SSLSocketImpl.writeRecordInternal(SSLSocketImpl.java:876)
	at sun.security.ssl.SSLSocketImpl.writeRecord(SSLSocketImpl.java:847)
	at sun.security.ssl.SSLSocketImpl.writeRecord(SSLSocketImpl.java:717)
	at sun.security.ssl.Handshaker.sendChangeCipherSpec(Handshaker.java:1077)
	at sun.security.ssl.ClientHandshaker.sendChangeCipherAndFinish(ClientHandshaker.java:1210)
	at sun.security.ssl.ClientHandshaker.serverHelloDone(ClientHandshaker.java:1122)
	at sun.security.ssl.ClientHandshaker.processMessage(ClientHandshaker.java:344)
	at sun.security.ssl.Handshaker.processLoop(Handshaker.java:979)
	at sun.security.ssl.Handshaker.process_record(Handshaker.java:914)
	at sun.security.ssl.SSLSocketImpl.readRecord(SSLSocketImpl.java:1062)
	at sun.security.ssl.SSLSocketImpl.performInitialHandshake(SSLSocketImpl.java:1375)
	at sun.security.ssl.SSLSocketImpl.writeRecord(SSLSocketImpl.java:747)
	at sun.security.ssl.AppOutputStream.write(AppOutputStream.java:123)
	at sun.nio.cs.StreamEncoder.writeBytes(StreamEncoder.java:221)
	at sun.nio.cs.StreamEncoder.implFlushBuffer(StreamEncoder.java:291)
[...]

PR #290, opened 2015-08-21
(#267) Ignore OcspIT.testGoodCertificate since we have not yet found …

cmebarrow:
…a way to get k3po to generate a good certificate response accepted by Oracle JDK 1.8.0_51.

Also added ITUtil, containing a utility methods for creating rule chains for robot tests, and changed a bunch of robot tests to use it, so all test methods have a timeout and trace log if they fail.

@jitsni This is **not yet ready for merging**, I just want to get your feedback before I go modify all the other robot tests,


cmebarrow:
@jitsni This is now ready for merging

PR #294, opened 2015-08-26
Feature/config translator ii

mjolie:


server/src/main/java/org/kaazing/gateway/server/config/parse/translate/CEGatewayConfigTranslatorFactorySpi.java
```diff
@@ -0,0 +1,64 @@
...
+
+import org.kaazing.gateway.server.config.parse.GatewayConfigNamespace;
+import org.kaazing.gateway.server.config.parse.GatewayConfigParser;
+import org.kaazing.gateway.server.config.parse.translate.spi.GatewayConfigTranslatorFactorySpi;
+
+/**
+ * Classes which translate/transform a DOM representing the config file implement this interface.  These classes are used by the
+ * {@link GatewayConfigParser}
+ */
+public class CEGatewayConfigTranslatorFactorySpi implements GatewayConfigTranslatorFactorySpi {
```
dpwspoon: Recommend using something other than CE, maybe BaseGatewayConfigTranslatorFactorySpi

server/src/main/java/org/kaazing/gateway/server/config/parse/GatewayConfigParser.java
```diff
@@ -158,13 +128,13 @@ private void translate(final GatewayConfigNamespace ns,
     private void translate(final GatewayConfigNamespace ns,
```
dpwspoon: unused method above this?

server/src/main/java/org/kaazing/gateway/server/config/parse/translate/CEGatewayConfigTranslatorFactorySpi.java
mjolie: This is the spi implementation for CE and is not used by EE ,     EE has its own implementation.

server/src/main/java/org/kaazing/gateway/server/config/parse/GatewayConfigParser.java
```diff
@@ -412,7 +379,13 @@ private void validateGatewayConfig(GatewayConfigDocument configDoc, List<String>
                         LOGGER.error("  Line: " + line + " Column: " + column);
                     }
                 }
-                LOGGER.error("  " + error.getMessage().replaceAll("@" + GATEWAY_CONFIG_NS, ""));
+                LOGGER.error("  " + error.getMessage().replaceAll("@" + GatewayConfigNamespace.CURRENT_NS, ""));
+                if (error.getMessage().contains("notify-options")) {
```
dpwspoon: Presumably it could also be [notify element](http://developer.kaazing.com/documentation/jms/4.0/admin-reference/r_conf_service.html#notifyele)

server/src/main/java/org/kaazing/gateway/server/config/parse/GatewayConfigParser.java
```diff
@@ -433,37 +406,6 @@ private void validateGatewayConfig(GatewayConfigDocument configDoc, List<String>
...
 
     /**
-     * Write out an input stream to a new file.
-     *
-     * @param in       the input stream
-     * @param fileName the file name
-     * @return whether file was successfully written
-     * @throws IOException
-     */
-    private void writeToNewFile(InputStream in, String fileName) throws IOException {
```
dpwspoon: In the past we output the new config .xml as gateway-config.xml.new or something like this.  (4.x line),  I'm assuming by the removal of this method we no longer do that?  I suggest that we should do that... thoughts?

server/src/main/java/org/kaazing/gateway/server/config/parse/translate/AddServiceNameVisitor.java
```diff
@@ -0,0 +1,138 @@
+/**
```
dpwspoon: This and August2012ToSeptember2012Translator will never be loaded in CE as far as I can tell.  Meaning its essentially dead code for the public.  Probably would be better to move internally.

server/src/main/java/org/kaazing/gateway/server/config/parse/GatewayConfigParser.java
mjolie: I was told APNS is the only notify we had supported  ?

server/src/test/java/org/kaazing/gateway/server/config/parse/translate/GatewayConfigTranslatorFactoryTest.java
```diff
@@ -0,0 +1,43 @@
...
+public class GatewayConfigTranslatorFactoryTest {
+
+    @Test
+    public void getTranslatorCurrentNSTest() {
+        try {
+            GatewayConfigTranslatorPipeline  translator = (GatewayConfigTranslatorPipeline) GatewayConfigTranslatorFactory.newInstance().getTranslator(
+                    GatewayConfigNamespace.CURRENT_NS);
+             assertTrue(translator.getTranslators().get(0) instanceof September2014Validator);
+        } catch (Exception e) {
+            // TODO Auto-generated catch block
```
dpwspoon: remove TODO

server/src/main/java/org/kaazing/gateway/server/config/parse/GatewayConfigParser.java
dpwspoon: Yeah, suggesting if (error.getMessage().contains("notify-options") || error.getMessage().contains("notify")).  Because both those elements are specific to APNS, and it's not clear to me which exception would get thrown first

dpwspoon:
LGTM for the most part.  However, there are several classes that are essentially dead code for the public.  I would recommend pulling them into our enterprise edition as we will keep config translation for older versions alive there.  Thoughts?

server/src/main/java/org/kaazing/gateway/server/config/parse/translate/CEGatewayConfigTranslatorFactorySpi.java
dpwspoon: Understand, AFAIK this is the first time we have stated community edition or equivalent to the public so wary of using it as abbreviation.  Maybe just CommunityGatewayConfigTranslatorFactory.  This is just a nit, whatever you think is best

mjolie:
I will move some of classes  to EE gateway server,   so hold off on that pull request till tomorrow.

server/src/main/java/org/kaazing/gateway/server/config/parse/GatewayConfigParser.java
mjolie: ok will add the or

server/src/main/java/org/kaazing/gateway/server/config/parse/GatewayConfigParser.java
mjolie: In the 4.0 code it looked like this method was only called if an exception happened during translation, and not just in general.  I will double check but pretty sure on on error path.

server/src/main/java/org/kaazing/gateway/server/config/parse/translate/CEGatewayConfigTranslatorFactorySpi.java
mjolie: I like  CommunityGatewayConfigTranslatorFactory I can change it to that

server/src/main/java/org/kaazing/gateway/server/config/parse/GatewayConfigParser.java
mjolie: removed

server/src/main/java/org/kaazing/gateway/server/config/parse/GatewayConfigParser.java
mjolie: the code that writes out the new config file is on line 149ish inside the translate method

mjolie:
@dpwspoon applied all code review feedback

dpwspoon:
Looks like the test failures are related to your changes :-(

server/src/test/java/org/kaazing/gateway/server/config/parse/GatewayConfigParserTest.java
```diff
@@ -305,12 +305,9 @@ public void testCanConvertAwsHostToAwsHostnameIfAvailable() {
...
         DefaultUtilityHttpClient httpClient = new DefaultUtilityHttpClient();
         try {
-            httpClient.performGetRequest("http://169.254.169.254/2014-02-25/meta-data/");
+            httpClient
+                    .performGetRequest("http://169.254.169.254/2014-02-25/meta-data/");
             onAWS = true;
-            // Need to check if running on Travis CI, Travis CI uses AWS container virtualization. If this test runs on
-            // container on AWS then it will fail because the hostname it gets in the query will not map via dns to the
-            // IP address the container has.
-            onTravisCI = System.getenv().containsKey("TRAVIS");
```
dpwspoon: These should not be removed and this is why the build is failing

server/src/main/java/org/kaazing/gateway/server/config/parse/GatewayConfigParser.java
dpwspoon: cool!

server/src/main/java/org/kaazing/gateway/server/config/parse/translate/AddServiceNameVisitor.java
dpwspoon: Still suggest moving this one to EE along with other files. But if you have looked at this and think it will be useful for CE in the future I'm fine leaving it here, just comment that you have seen this

dpwspoon:
@mjolie reviewed. Looks good almost there

server/src/test/java/org/kaazing/gateway/server/config/parse/GatewayConfigParserTest.java
mjolie: I did not remove this I am not sure how this happened

server/src/main/java/org/kaazing/gateway/server/config/parse/translate/AddServiceNameVisitor.java
mjolie: It has been moved ?

dpwspoon: I'm specifically talking about the AddServiceNameVisitor class being moved out of CE.

mjolie: right I got it , I was able to move that a few more classes

mjolie:
@dpwspoon ok ready for another review

mjolie:
retest this please

dpwspoon:
Thanks @mjolie

PR #297, opened 2015-08-27
Remove wsr from gateway

mjolie:
I removed all references  except for 3 java script files here:

 /gateway.distribution/target/bower-dependency/command-center/commandcenter/js

       I have not work with js for years, so not sure how to edit and test and I am not sure what code uses these js files  ?


mjolie:
@chao-sun-kaazing @jitsni   I assigned this PR to Chris but I think one of you should review this also.   thanks

resource.address/tcp/src/main/java/org/kaazing/gateway/resource/address/tcp/TcpResourceAddressFactorySpi.java
```diff
@@ -226,12 +226,6 @@ protected void setOptions(TcpResourceAddress address, ResourceOptions options, O
...
         super.setOptions(address, options, qualifier);
 
-        // note:  need to propagate bind address and network context,
-        //        you might think they are already handled during address creation
-        //        but in fact due to dynamic creation of WSR RTMP addresses they should be propagated
-        //        so that the new TCP layer can know about rtmp/1.0, the bind address and the network context
-        //        because we need to map addresses, bind properly for the WSR RTMP tcp transport layer.
-
         address.setOption0(BIND_ADDRESS, options.getOption(BIND_ADDRESS));
         address.setOption0(MAXIMUM_OUTBOUND_RATE, options.getOption(MAXIMUM_OUTBOUND_RATE));
```
cmebarrow: The comment you removed implies maybe we don't need the two address.setOption0 calls and so you could remove them too. @jitsni What do you think?

cmebarrow:
@mjolie /gateway.distribution/target/bower-dependency/command-center/commandcenter/js is in target so is a product of the build. Are you saying that contains references to wsr?

dpwspoon:
gateway.distribution/target/bower-dependency/command-center/commandcenter/ is to deal with command center.  There is probably special cases for WSR in command center, like icons and what not.  I don't think it is necessary to remove WSR from command center as it should still function fine.

resource.address/tcp/src/main/java/org/kaazing/gateway/resource/address/tcp/TcpResourceAddressFactorySpi.java
mjolie: @jitsni @cmebarrow are we waiting just waiting on @jitsni  to answer this question before we can merge ?

jitsni: The method copies all the options from one address to another (all the ResourceAddressFactorySpi impl do) So, let us keep them.

PR #303, opened 2015-09-04
Gateway enhancements for improved extensibility

cmebarrow:
- Allow extended properties on security realm (authentication context)
- Add ExtensionHelper to provide extra features for WebSocket extensions
- Move AMQP service XSD element type definition into the standard gateway XSD file instead of having it in its own file under service/amqp


server/src/main/java/org/kaazing/gateway/server/test/config/RealmConfiguration.java
```diff
@@ -37,6 +40,7 @@
     private Suppressible<String> _httpChallengeScheme;
     private Suppressible<String> _authorizationMode;
     private Suppressible<String> _sessionTimeout;
+    private Map<String, Suppressible<String>> _extendedProperties = new HashMap<String, Suppressible<String>>();
```
jitsni: we should use <> instead of HashMap<...>

service/amqp/src/test/java/org/kaazing/gateway/service/amqp/CreateAmqpProxyServiceTest.java
```diff
@@ -36,9 +35,7 @@
 
     @Test
     public void createTestService() {
-        PropertyConfigurator.configure("src/test/resources/log4j-trace.properties");
-
-        ServiceFactory factory = ServiceFactory.newServiceFactory();
+                ServiceFactory factory = ServiceFactory.newServiceFactory();
```
jitsni: formatting is required (may be because of tab char)

transport/ws/src/main/java/org/kaazing/gateway/transport/ws/extension/WebSocketExtension.java
```diff
@@ -39,10 +39,10 @@
      * in the order that the extensions were specified in the HTTP request, so the first extension is closest to the network
      * (i.e. gets to see and possibly modify frames read from the client before other extensions, and  gets the final say for
      * Frames being written to the client).
-     * @param extension    Details of the negotiated extension and parameters
+     * @param helper    Provides methods which may be used within extension filters
      * @return A filter which is to be added to the filter chain, or null if none is to be added
      */
-    public IoFilter getFilter() {
+    public IoFilter getFilter(ExtensionHelper helper) {
```
jitsni: It seems to me that ExtensionHelper should be passed to extension during initialization. getFilter() getting ExtensionHelper object doesn't seem to be right.

dpwspoon:
Closing in favor of https://github.com/kaazing/gateway/pull/318

PR #318, opened 2015-09-12
Gateway enhancements for improved extensibility

dpwspoon:
Updates for https://github.com/kaazing/gateway/pull/303


jitsni:
Other thing I mentioned to Chris was, ResultAwareLoginContext becomes public API. May be that dependency can be removed. Perhaps, spi's LoginResult was good enough. He thought that might be possible as HttpRevalidateHandler doesn't have to send 403.

dpwspoon:
@jitsni, I took a look (crack at doing it).  Taking a break now but it seems doable but I hit two problem spots, both in AbstractWsBridgeSession.  

One is without the login ctx the getSubject is not available.  I'm unclear on the reasoning here on why the call to getSubject was added in the PR.  I will need to investigate further, if it is required could we add the subject to the login Result?

Two, the cleanup in `logout()`.  It looks like there was a way solved for this before the pull request but it seems a little hacky.

dpwspoon:
@jitsni, I attempted to pull out ResultAwareLoginContext and instead rely on LoginResult.  However, this has downstream impact.  Suggest merging this for now and revisiting.  Here is branch to [attempt](https://github.com/dpwspoon/gateway/tree/cmebarrow-revalidate2).   Essentially, ResultAwareLoginContext is still needed for downstream testing

PR #328, opened 2015-09-18
upgrade handler is not found during ws connect

jitsni:
https://github.com/kaazing/gateway/issues/162

Before upgrade handler is set (in HTTP ConnectFuture listener),
HTTP transport gets 101 and closes HttpSession. HttpConnectProcessor
tries to upgrade, but doesn't find upgrade handler

The logic in HTTP ConnectFuture listener is moved to HTTP CloseFuture
listener. The CloseFuture listener is added to HTTP session in
initializer.


cmebarrow:
@jitsni The changes look reasonable to me. Is this ready to merge or did you want to do further work on it?

jitsni:
I think it is ready to merge

PR #329, opened 2015-09-22
Closing wse pending writer when the wse session is closed

jitsni:


dpwspoon:
@jitsni  Can you do a catch up merge.  And should this go into the RC?

jitsni:
Not needed for RC

jitsni:
@cmebarrow do you think we need this one ? If it is required then I will do catchup merge on this branch.

cmebarrow:
@jitsni Looks like a reasonable change. What motivated making the change? Is there a test case that was failing?

jitsni:
@cmebarrow It just feels like the pending writer is not closed. I don't have a test case. Also, I haven't observed any failure in existing test cases.

sbadugu:
@dpwspoon @jitsni Can we merge this PR ?

dpwspoon:
@jitsni Please do a catch up merge and or close.  We can get this in next rc and run full regression tests to make sure there is no race

jitsni:
Closing this for now.

PR #346, opened 2015-10-23
Updated return code/status with '504 Gateway Timeout' for http.proxy

msalavastru:


msalavastru:
I am now working on the spec tests. I wanted to ask, how can I make the test I added in the PR pass? As I said it times out at "unbound".

msalavastru:
@jitsni I have created a test case for the scenario when the server is down starting on the example above. I have pushed the changes. I have also added the k3po scripts in my fork for the k3po: https://github.com/msalavastru/k3po/tree/http.proxy

The problem is that when running the HttpProxyStatusIT.proxyShouldReturn504ResponseWhenServerIsDown() test, the server does not accept the connection from the proxy. Can you please have a look, and let me know how I can fix it?

Also, is this test case enough?

jitsni:
* The test I referred to was using transport option in scripts, for e.g proxy.rpt

```
connect http://localhost:8080/index.html
        option transport tcp://localhost:8081
```

So the Host header would be still localhost:8080, but the acutal socket connection would be to localhost:8081

Unless I try it out, I wouldn't know. May be I will try it and create a new PR (including your commits)

* The RFC 2616 is obsolete. It is replaced by RFC 7230-7234. So your test should faill into one of them.

* I want a test case where there is no server listening at port 8081.  May be the same client script can be used in gateway as a separate test case.

msalavastru:
@jitsni I have moved my test in the correct RFC folder 7231 and slightly changed the client, proxy and server k3po scripts to use "transport option". I am now stuck at this 2 points:

- I am trying to build a test case where the end server suddenly closes the connection:
   -  the test in ResponseStatusCodesIT from the specification.http/org.kaazing.specification.http.rfc7231/ ( in the K3PO ) fails: it does not connect to the end server. 
  - for the test in HttpProxyResponseStatusCodesIT from the gateway.service.http.proxy I don't know how to add "option transport tcp://localhost:8081" to the .connect(URI.create("http://localhost:8080/server")) in the GatewayConfiguration
- Maybe you can elaborate on the test case where there is no server listening at port 8081, I don't know how to do that.

The changes for the gateway are updated here in the PR, and you can see the updated k3po script from my fork in the [commit](https://github.com/msalavastru/k3po/commit/93aac2f52a0e202407708103a40bacbd16acb774)

jitsni:
There two PRs including Mihai's work:
https://github.com/kaazing/gateway/pull/349
https://github.com/k3po/k3po/pull/245

msalavastru:
@jitsni What should we do with this PR ?

jitsni:
Yes close it (I think when the other pull request is merged this would be closed automatically). Once k3po is released, then I update my pull request.

Issue #347, opened 2015-10-23
Sporadic NPE from WsnAcceptor while closing the session


sanjay-saxena:
Here is the stack trace:

```
1730 [New I/O worker #265] WARN  DefaultIoFilterChain  - Unexpected exception from exceptionCaught handler.
java.lang.NullPointerException
	at org.kaazing.gateway.transport.wsn.WsnAcceptor$4.doExceptionCaught(WsnAcceptor.java:697)
	at org.kaazing.gateway.transport.wsn.WsnAcceptor$4.doExceptionCaught(WsnAcceptor.java:506)
	at org.kaazing.gateway.transport.IoHandlerAdapter.exceptionCaught(IoHandlerAdapter.java:31)
	at org.kaazing.gateway.transport.AbstractBridgeHandler.exceptionCaught(AbstractBridgeHandler.java:47)
	at org.kaazing.mina.core.filterchain.DefaultIoFilterChain$TailFilter.exceptionCaught(DefaultIoFilterChain.java:704)
	at org.kaazing.mina.core.filterchain.DefaultIoFilterChain.callNextExceptionCaught(DefaultIoFilterChain.java:493)
	at org.kaazing.mina.core.filterchain.DefaultIoFilterChainEx.callNextExceptionCaught(DefaultIoFilterChainEx.java:156)
	at org.kaazing.mina.core.filterchain.DefaultIoFilterChain$EntryImpl$1.exceptionCaught(DefaultIoFilterChain.java:798)
	at org.kaazing.gateway.transport.IoFilterAdapter.doExceptionCaught(IoFilterAdapter.java:86)
	at org.kaazing.gateway.transport.IoFilterAdapter.exceptionCaught(IoFilterAdapter.java:82)
	at org.kaazing.mina.core.filterchain.DefaultIoFilterChain.callNextExceptionCaught(DefaultIoFilterChain.java:493)
	at org.kaazing.mina.core.filterchain.DefaultIoFilterChainEx.callNextExceptionCaught(DefaultIoFilterChainEx.java:156)
	at org.kaazing.mina.core.filterchain.DefaultIoFilterChain$EntryImpl$1.exceptionCaught(DefaultIoFilterChain.java:798)
	at org.apache.mina.core.filterchain.IoFilterAdapter.exceptionCaught(IoFilterAdapter.java:107)
	at org.kaazing.mina.core.filterchain.DefaultIoFilterChain.callNextExceptionCaught(DefaultIoFilterChain.java:493)
[...]

PR #353, opened 2015-11-06
wsn session may be null when there is an exception

jitsni:
wsn session may be null when there is an exception (for e.g. wsn session is not even created, and i/o error happened on http layer). Guarding against it.


jitsni:
Forward porting the PR that fixes
https://github.com/kaazing/gateway/issues/347

PR #358, opened 2015-11-10
Added IT tests for RFC 7232 for Directory Service

sanjay-saxena:


service/http.directory/src/test/java/org/kaazing/gateway/service/http/directory/rfc7232/IfMatchIT.java
```diff
@@ -0,0 +1,125 @@
...
+import org.kaazing.gateway.server.test.GatewayRule;
+import org.kaazing.gateway.server.test.config.GatewayConfiguration;
+import org.kaazing.gateway.server.test.config.builder.GatewayConfigurationBuilder;
+
+/**
+ * RFC-7232, section 3.1 "If-Match"
+ */
+public class IfMatchIT {
+    private final K3poRule k3po = new K3poRule().setScriptRoot("org/kaazing/specification/http/rfc7232/preconditions/if.match");
+    private final TestRule timeout = new DisableOnDebug(new Timeout(5, SECONDS));
```
dpwspoon: See ITUtilRule, I have no preference on what we use here but I think Chris is trying to get people to standardize

pom.xml
```diff
@@ -356,6 +350,38 @@
                 </configuration>
             </plugin>
             <plugin>
```
dpwspoon: Thanks for fixing this

service/http.directory/src/test/java/org/kaazing/gateway/service/http/directory/rfc7232/IfMatchIT.java
sanjay-saxena: I think John wants the tests to be in terms of primitives. So, once we have a resolution, we can go and change all the tests to conform to one directive or the other.

dpwspoon:
Looks like you have a build error that needs to be resolved

```
'dependencies.dependency.version' for org.kaazing:specification.http:jar is missing. 
```

PR #360, opened 2015-11-17
Override host header in http.proxy with hostname from <connect>

msalavastru:
Overridden host header in HttpProxyServiceHandler and updated the k3po scripts for the ITests in http.proxy.


msalavastru:
Retest this please

msalavastru:
retest this please

irina-mitrea-luxoft:
retest this please

msalavastru:
This is the config file that I used to do the testing.

[gateway-config-minimal.xml](https://files.zenhub.io/564df1972d47521d4101db78)
Uploaded using [**ZenHub.io**](http://www.zenhub.io)

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpResponseDecodingState.java
```diff
@@ -257,7 +257,8 @@ protected DecodingState finishDecode(List<Object> childProducts,
 						// create the cookie name and value
 						DefaultHttpCookie cookie = new DefaultHttpCookie(
 								nvPair[0].trim());
-						cookie.setValue(nvPair[1].trim());
+						if (nvPair.length > 1)
+						    cookie.setValue(nvPair[1].trim());
```
cmebarrow: @msalavastru This fix seems fine. Where's the k3po test you mentioned that verifies this fix? Also, as mentioned in our call just now, please add a test method to unit test HttpResponseDecoderTest to confirm the fix (make sure it fails without the fix).

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -179,23 +178,25 @@ private static boolean processHopByHopHeaders(HttpSession src, HttpSession dest)
...
-        for(Map.Entry<String, List<String>> e : src.getReadHeaders().entrySet()) {
+        for (Map.Entry<String, List<String>> e : src.getReadHeaders().entrySet()) {
             String name = e.getKey();
-            for(String value : e.getValue()) {
+            for (String value : e.getValue()) {
                 if (!hopByHopHeaders.contains(name)) {
-                    dest.addWriteHeader(name, value);
+                    if (name.equalsIgnoreCase(HttpHeaders.HEADER_HOST)) {
+                        dest.addWriteHeader(name, dest.getRemoteAddress().getResource().getAuthority());
+                    } else {
```
jitsni: Can you also look at HttpConnectProcessor where it adds Host header ? Also see the history w.r.t Host header and may be a fix in that file would be better. Also, I need to think a bit here. 

We may provide a config option to override the Host header. In that case, we need to see what the fix would be.

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpResponseDecodingState.java
msalavastru: @cmebarrow I have added the unit test that covers this fix in HttpResponseDecoderTest. I think we no longer need the k3po test, as the unit test seems sufficient, and the k3po would just be redundant.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
msalavastru: @jitsni I have moved the fix to HttpConnectProcessor, it seemed to be a better place for it. And also modifies only the HttpRequestMessage with the updated header and not the headers from the session. Let me know what you think.

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpConnectProcessor.java
```diff
@@ -87,12 +87,13 @@ protected void finishConnect(DefaultHttpSession session) {
                 // TODO: strip port if default
                 httpRequest.setHeader(HttpHeaders.HEADER_HOST, resource.getAuthority());
             }
```
jitsni: I think this if statement is NOT required (as HEADER_HOST is set below unconditionally)

jitsni:
Also, resolve the conflicts.

msalavastru:
@jitsni I have applied the review change and resolved the conflicts.

PR #335, opened 2015-09-29
Remove/block use of accept in properties section of broadcast service 

mjolie:

PR #361, opened 2015-11-20
Fix BroadcastServiceTest.FastTestBackendProducer member variable runnable to be volatile 

cmebarrow:
to avoid hang in testSlowClient.


cmebarrow:
Closing because Marie's PR https://github.com/kaazing/gateway/pull/335 is the one that should be merged. This was just experimental.

PR #362, opened 2015-12-02
Streaming http response body

jitsni:
Adding the support to stream large http responses. We already have the support for this for http requests.
This should fix
https://github.com/kaazing/tickets/issues/206


transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpRequestDecodingState.java
```diff
@@ -477,47 +477,4 @@ private void flush(List<Object> childProducts, ProtocolDecoderOutput out) {
         return cookies;
     }
 
-    private static final class MaximumLengthDecodingState implements DecodingState {
```
jitsni: Moving to separate file as it is used by both request and response decoders

PR #367, opened 2015-12-03
Formalized support for extension services without requiring specific type…

sanjay-saxena:
…s in the XSD.


jitsni:
I think we don't want this xsl transformation. We should be able to handle in the code. For now, I am merging this one.

PR #370, opened 2015-12-09
Forward Port to 5.x: Log when users disconnect

msalavastru:
Started port with [Transport logging improvements #85](https://github.com/kaazing-private/gateway.server/pull/85). I will add comments as I advance.


msalavastru:
@cmebarrow  Based on what I found so far, I have a few comments about the  [Transport logging improvements #85](https://github.com/kaazing-private/gateway.server/pull/85) port:

- **WsebAcceptor**: apart from the changes in the section of the code that is similar between 4.x and 5.x, in gw 4.x there are more changes than in gw 5.x. This would imply adding new classes in gw5.x, suchs as HttpRevalidateHandler. Should I add this changes to 5.x?
- **WsebSession**: the constructor is different in 4.x vs 5.x ( the last parameter is a List vs Properties) this has implications in WsebAcceptor and WsebConnector. What constructor should I choose?
- **WsebAcceptProcessor** and **WsebConnecProcessor**: have extra code in 4.x vs 5.x : in flushInternal method starting from "else if (WsebSession.isPingRequest(request) || WsebSession.isPongRequest(request)..." Should I add this section of code to 5.x?
- **WsebConnectProcessor**: in 4.x it uses HttpHeaders from org.apache.http, in gw5.x I have used HttpHeaders from kaazing.gateway. Is this ok?
- **WsebConnectorlogging** - the parameters passed to the connect method ( from WsebConnectorRule ) are IoHandler's but the connect method in WsebConnectorRule is expecting a IoHandlerAdapter ( as it is in gw4.x). Should I alter the WsebConnectorlogging or the WsebConnectorRule?

Can you please take a look and let me know how I should handle the situations described above?
I did most of the other changes for issue #85, just waiting for your answers to finalize this part.

cmebarrow:
@msalavastru See my answers in-line in your comment above

msalavastru:
@cmebarrow Thanks for your answers. I have updated what was needed. I wil proceed to the next PR : https://github.com/kaazing-private/gateway.server/pull/87

cmebarrow:
@msalavastru Please look into the Travis build failure in case it has anything to do with your changes

msalavastru:
@cmebarrow I have found the tests that were failing. The fails are caused by the new changes. I have fixed some of them by adding a small fix to the WsebConnectorRule, but there are some more tests that are failing. For example the tests in **BinaryIT** ( from package **org.kaazing.gateway.transport.wseb.specification.wse.acceptor**) are failing with IllegalArgumentException : "Missing transport 'wsn'". This is thrown by the **resolveSchemes** method from GatewayContextResolver because transportContext ( of type DefaultTransportContext is null ) for the transportName "wsn". I am now blocked at trying to figure out why this is happening. I have pushed my changes so far. Maybe you can have a look and let me know what to do? Thanks.

cmebarrow:
@msalavastru Just add a test dependency on wsn in the pom for wseb transport:
```
       <dependency>
            <artifactId>gateway.transport.wsn</artifactId>
            <version>${project.version}</version>
            <scope>test</scope>
        </dependency>
```
With this change everything passes for me except:
- WsebAcceptorLoggingIT.shouldLogOpenWriteReceivedAndAbruptClose: this fails because the log message check comes too early. Irina fixed this in 4.x in her identity logging changes so when you forward port that work it should be ok.
- WsebConnectorIT.shouldReplyPongToPing: should be resolved once we complete the wseb spec tests

irina-mitrea-luxoft:
@cmebarrow @msalavastru Hello, additional fixes on "\*LoggingIT.java" ITests (where "\*" is a wildcard) can be found in the PRs attached to [this issue](https://github.com/kaazing/tickets/issues/203) (please note that both PRs in the above-listed issue are relevant). Thank you.

msalavastru:
@cmebarrow I did the fix you suggested and the tests that were failing before are now passing. The problem persists with the LoggingIT tests that fail, because we need the changes that Irina did in 4.x. Given that the next PR's that I need to port have changes in the files that were modified in the current PR how do you suggest I should proceed? Create PR's starting from the current changes in this PR in new branches ( any changes that we might do here, would also need to be made to the other PR's)?

Should I start with Irina's PR starting from the current changes?

cmebarrow:
@msalavastru Mihai, yes, you can always create PR's starting from the current changes in this PR in new branches. I would advise doing the forward ports in the order that the original PR's were merged. So for the PR's Irina mentioned (which I also added to https://github.com/kaazing/roadmap/issues/562 for tracking purposes), I would forward port them separately after this one is merged. I will go ahead and merge even if there are test failures that Irina's PRs would fix, because I know you will be forward porting those later.

msalavastru:
@cmebarrow I have added the other PR's for the [Log when user disconnects #583](https://github.com/kaazing/roadmap/issues/583) user story as comments in the [Engineering story](https://github.com/kaazing/roadmap/issues/626). They are for GW 5.x

PR #372, opened 2015-12-14
[Not to be merged] Add support for network interface syntax

irina-mitrea-luxoft:
Pull request not to be merged (resolution performed at GatewayContextResolver level) which will serve as reference point for having resolution performed at the lower layers (e.g. TcpResourceAddressFactorySpi).

Pending items:
- cosmetic fixes to the code: package refactoring, etc.

[This comment is to be updated as more changes are checked into this branch]


server/src/main/xsd/gateway-config-201409.xsd
```diff
@@ -476,7 +476,7 @@
                      type="gateway:CollapsedString"/>
 
             <element maxOccurs="unbounded" minOccurs="0" name="accept"
-                     type="anyURI"/>
```
jitsni: I think most schema processors just don't do much processing (perhaps only invalid chars)
Are you running into any issues here ?

irina-mitrea-luxoft: The implementation should work with xsd:anyURI also; had added the new type in order to be able to define specific patterns, the implementation should work with this reverted also.

irina-mitrea-luxoft:
@jitsni Hello, Jitu. Could you please perform a review on the pull request at its current stage (only a first set of testing was started)? It would be useful to know your opinion on the current approach. Thank you, Irina.

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/networkinterface/resolution/NetworkInterfaceResolutionIT.java
```diff
@@ -0,0 +1,99 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TestRule;
+import org.kaazing.gateway.server.test.GatewayRule;
+import org.kaazing.gateway.server.test.config.GatewayConfiguration;
+import org.kaazing.gateway.server.test.config.builder.GatewayConfigurationBuilder;
+import org.kaazing.k3po.junit.annotation.Specification;
+import org.kaazing.k3po.junit.rules.K3poRule;
+
+public class NetworkInterfaceResolutionIT {
```
irina-mitrea-luxoft: For the moment the interface passed to the test resolves to the loopback address on Windows. Still to do: get results regardless of OS.

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/networkinterface/resolution/NetworkInterfaceResolutionIT.java
```diff
@@ -0,0 +1,99 @@
...
+
```
irina-mitrea-luxoft: For the moment the interface passed to the test resolves to the loopback address on Windows. Item pending: get results regardless of OS.

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/networkinterface/resolution/NetworkInterfaceResolutionIT.java
```diff
@@ -0,0 +1,99 @@
...
+                            .type("proxy")
+                        .done()
+//                        .service()
+//                            .accept("http://localhost:8110")
+//                                .acceptOption("http.transport", "tcp//[@Software Loopback Interface 1]:7081")
+//                            .connect(URI.create("http://localhost:8080"))
+//                            .type("proxy")
+//                        .done()
+                        .service()
+                            .accept("tcp://[@Software Loopback Interface 1]:8001")
```
irina-mitrea-luxoft: The current implementation for the gateway configuration builder relies on the fact that the parameters passed to the accept methods in both services and clustering contexts are URIs, e.g.:

>
GatewayConfiguration configuration =
                    new GatewayConfigurationBuilder()
                         .service()
                            .accept(URI.create("tcp://localhost:8003")) //URI passed as parameter
                            .connect(URI.create("tcp://localhost:8004"))
                            .type("proxy")
                        .done()

In order to ensure support for String \<accept\>s which do not comply to standard URIs (e.g. network interfaces syntax) the testing config builders needed extending with accept methods which can receive Strings as parameters, both in services and clustering contexts.

This seems a bit redundant to me, since we now have "accept"s stored as URIs and "stringAccept"s stored as Strings on both clusters and services. 

Normally, switching to Strings as parameters instead of URIs for accept methods would have seemed to be the better option to me, but this would have meant refactoring the entire testing framework, which I'm not sure would be the desired approach for now (Jitu, can you please confirm this?). 

I would have considered this to be the better approach also due to the fact that existing implementation involved building the URIs based on the strings passed as accept method parameters, and then in the org.kaazing.gateway.server.test.Gateway class, the default implementation performs the conversion back to Strings via the java.net.URI.toASCIIString() method (basically the existing implementation wraps a String in an URI and then performs the unwrap), e.g. (from org.kaazing.gateway.server.test.Gateway.java):

>
private void setCluster(GatewayConfig gatewayConfig, ClusterConfiguration cluster) {
...
Collection<URI> accepts = cluster.getAccepts();
...
for (String accept : stringAccepts) {
      newCluster.addAccept(accept);
}

Can you please provide your input on this?

jitsni: I think we should bite the bullet and go for the end goal.
* Change ResourceAddressFactory, ResourceAddressFactorySpi methods that take URI parameter to String
* Similarly in GatewayConfigBuilder accept, connect methods
* TcpResourceAddressFactorySpi need to be changed to take care of String location with network interface addresses (add alternate addresses if an interface maps to multiple IP addresses)

We can start there and keep making the changes. This will change a lot of files, but John wanted to do that anyway. So let us do that. If you want me do some of the above changes, let me know.

irina-mitrea-luxoft: @jitsni Hello, Jitu, I think the above comment has a major impact on the existing pull request. As I undertand, this would impact:
>
Change ResourceAddressFactory, ResourceAddressFactorySpi methods that take URI parameter to String

- impact on the implementation

>
Similarly in GatewayConfigBuilder accept, connect methods

- impact on the existing testing framework together with all the ITests which make use of the gateway configuration

>
TcpResourceAddressFactorySpi need to be changed to take care of String location with network interface addresses (add alternate addresses if an interface maps to multiple IP addresses)

- impact on the implementation

Can we schedule a meeting to discuss this? As I understand, this would cancel the current implementation approach which performs resolution in the GatewayContextResolver. Thank you.

irina-mitrea-luxoft:
Integration testing implemented for clustering, accepts and accept options scenarios (transports and tcp.bind). 
[Observation on NetworkInterfaceResolution utils](https://kaazing.slack.com/archives/pull-requests/p1452017169000073) is still valid (giving reference here to [Jitu's reply to this topic](https://kaazing.slack.com/archives/pull-requests/p1452041930000101)).

irina-mitrea-luxoft:
Integration testing implemented for clustering, accepts and accept options scenarios (transports and tcp.bind) regardless of OS. 
[Observation on NetworkInterfaceResolution utils](https://kaazing.slack.com/archives/pull-requests/p1452017169000073) is still valid (giving reference here to [Jitu's reply to the topic](https://kaazing.slack.com/archives/pull-requests/p1452041930000101)). - listing these items due to the fact that the comments listed here are now by default hidden due to the fact that the diff on which they were placed is now outdated

irina-mitrea-luxoft:
Based on the discussion held with @jitsni and @sbadugu this pull request is not be merged and will serve as reference point for having resolution performed at the lower layers (e.g. TcpResourceAddressFactorySpi).

jitsni:
Can we close this one as the other PR https://github.com/kaazing/gateway/pull/437 is merged

jitsni:
Closing this PR as the network interfacer support is added using PR https://github.com/kaazing/gateway/pull/437

Issue #183, opened 2015-06-07
Need setup instructions for developing Gateway from Eclipse


brennangaunce:
Need setup instructions for developing Gateway from Eclipse

Start Eclipse (Luna)
Import Gateway (top-level project) as Maven Project
Currently seeing "This compilation is not on the build path of a Java project" when editing source (e.g. broadcast service: BroadcastListenHandler).

Need to add steps for setting up Eclipse to README or wiki to get started.

PR #382, opened 2015-12-24
Http.proxy does not append path in connect #183

msalavastru:


sbadugu:
@jitsni Please review this.

sbadugu:
@dpwspoon Please review this.

dpwspoon:
I suspect that this is not sufficient.   Say we have the following config

```xml
<type>http.proxy</type>
<accept>http://example.com/path1</accept>
<connect>http://kaazing.wan/path2<connect>
```

I believe the expected behavior should be the following, for the different scenarios

```
http://example.com/path1/   ----->   http://kaazing.wan/path2/
http://example.com/path1/abc    ----->   http://kaazing.wan/path2/abc
```

I don't see that logic in https://github.com/kaazing/gateway/blob/develop/service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java for the ConnectSessionInitializer.

We should probably have some tests too.

jitsni:
We should do little bit research on the requirements. nginx etc support path templating. We should loop @robinzimmermann for this.

dpwspoon:
I can confirm that nginx does the above by default.  With the following config

```
server {
    listen       8080;
    location /maven {
        proxy_pass http://nexus.kaazing.wan:8081/nexus/content/repositories/kaazing.enterprise.public;
    }

}
```

so /maven/example/index.html
ends up at http://nexus.kaazing.wan:8081//nexus/content/repositories/kaazing.enterprise.public/example/index.html

Which is the behavior as above.  It then returns a 404 if you try http://nexus.kaazing.wan:8081/maven/../

We should do this same behavior.  Happy to have @robinzimmermann review this but I had talked to him when we first discovered this and this is the behavior we agreed upon then.

dpwspoon:
Robin already commented on https://github.com/kaazing/roadmap/issues/515, it should be as above

msalavastru:
@dpwspoon Hi David, I have implemented the change to append the path from accept to the path in the connect, but it seems to be affecting some tests, such as the ones in WsnConnectorTest, so I will have to further investigate that.

I do have a question regarding the parameters in the query. Suppose we have the following config:

```xml
<type>http.proxy</type>
<accept>http://example.com/</accept>
<connect>http://kaazing.wan/connectpath?param2=value2<connect>
```

For accessing the link in accept with a query such as
```
http://example.com/acceptpath?param1=value1
```

should we expect the requestURI of the httpRequest to be like below?

```
/connectpath/acceptpath?param2=value2&param1=value1
```

Right now the requestURI is **/connectpath/acceptpath?param2=value2**. The parameters in accept seem to be lost.

jitsni:
Have you tried with nginx ? I think nginx can be configured to use templates, but can you research the default behaviour ?

dpwspoon:
@msalavastru, thanks for bring up query parameters.  At this point we don't have a use case for our users to customize query parameters through the gateway proxy. Other proxies (nginx) allow customizing it in many different ways, which if we were to fully support would probably require several accept and connect options.  So for now, lets not add behavior for query parameters, and instead throw an error when the gateway is started with query parameters in the accept or connect. 

Note: Query parameters from the client request should pass through.

So in your example above, it would error out saying something like "Query parameters are not allowed in http.proxy and accept, as they have no meaning"

Thus the configuration would have to be changed to

```xml
<type>http.proxy</type>
<accept>http://example.com/</accept>
<connect>http://kaazing.wan/connectpath?param2=value2<connect>
```

And a request of `http://example.com/acceptpath?param1=value1` would come out as 
`/connectpath/acceptpath?param1=value1`

msalavastru:
@dpwspoon Thanks for your great explanations and examples [here](https://github.com/kaazing/roadmap/issues/515#issuecomment-169765172). I have been trying to look for a good place where to implement the change, where to check for "../" and return 404 based on the situation, and was unable to find one. Can you please help with a suggestion where I should look?

Regarding the changes that I uploaded to the PR so far: the path in the client request is appended to the path in connect, and manually testing this seems ok, except the fact that the page retrieved does not have the correct formatting. It seems that the resources are not loaded ( css, js..etc ). From what I see the pathInfo in the acceptSession first contains any path appended in the client request to the address in the <accept> ( I was looking in the HtttpProxyServiceHandler), and then it contains the path to the resources. By using the implementation in my PR, the pathInfo is always appended to the path in connect, and this is not correct for the resources. I need to distinguish between these to know when to append the pathInfo, and when not to in the HttpConnectProccesor, so that the resources get loaded correctly. Do you have any suggestions? Is my approach ok?

I will also post a question regarding the requirements in the [user story](https://github.com/kaazing/roadmap/issues/515).

dpwspoon:
@msalavastru, It seems to me that a good place to implement the logic would be in the HttpProxyServiceHandler as it seems like this is http proxy specific code.  So yes, I was a little curious with the choice of putting it in the HttpConnectProcessor, sorry for not mentioning it sooner. 

@jitsni is more familiar/fluent with that http proxy code than I am and might have a better suggestion on where to actually put it, or how to implement it.  But at first glance the ConnectSessionInitializer seems to be using attributes from the acceptSession to populate the connect request.  Perhaps you could populate the path there.

Note, the directory service has very similar ../ requirements, and has an implementation in HttpDirectoryServiceHandler under doSessionOpened that may or may not be useful for your implementation.

Also, I'm assigning this back to @jitsni for now as he is sort of the expert and original implementor of the http proxy service.

jitsni:
ConnectSessionInitializer seems to be right place. I don't see where .. come into picture. You may want to do something like this: 
* Canoncalize the accept path
* Check if the canoncalized path contains the path specified in the accept URI's path
* Compute the appropriate connect request URI and set in the connect session

msalavastru:
@jitsni I have moved the logic of computing the path in the HttpProxyServiceHandler and also added validation for using "../" in the request. If you have time, can you please have a look and let me know if it is ok.

I have also started working on two k3po tests. The test "httpProxyPath()"  ( that uses 'http.proxy.path.rpt' )  fails with the client receiving "504 Gateway Timeout" instead of "200 OK", because of an "Early termination of IO connection". Do you have any idea why this is happening?

transport/http/src/main/java/org/kaazing/gateway/transport/http/DefaultHttpSession.java
```diff
@@ -474,6 +474,11 @@ public URI getPathInfo() {
     }
 
     @Override
+    public void setPathInfo(URI pathInfo) {
```
jitsni: Are you using this method ?

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpConnectSession.java
```diff
@@ -21,14 +21,16 @@
...
     
 	void setRequestURI(URI requestURI);
 	
     void setMethod(HttpMethod method);
-    
+
     // setContentLength(int length);
-    
+
+    void setPathInfo(URI pathInfo);
```
jitsni: Are you using this method ?

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -94,10 +95,37 @@ public void initializeSession(IoSession session, ConnectFuture future) {
...
+            if (connectPathEndsInSlash && acceptPathStartsWithSlash) {
+                acceptPath = acceptPath.substring(1);
+            }
+            return connectPath + acceptPath;
+        }
+
+        private boolean validateComputedPath(String path, String connectPath) {
+
+            path = URI.create(path).normalize().toString();
+            if (path.indexOf("/\\.\\./") == 0 || path.indexOf(connectPath) != 0) {
```
jitsni: I think we can reduce the logic. I think validation should include only request path and accept path. Then if validation passes, just use the path to resolve connect URI (I think you can use resolve() method)

service/http.proxy/src/test/scripts/org/kaazing/gateway/service/http/proxy/http.proxy.path.security.rpt
```diff
@@ -0,0 +1,42 @@
...
+
+connect http://localhost:8110/acceptConfigPath/../../example/../jsp/num/file.html
+connected
+
+write method "GET"
+write parameter "name" "value"
+write version "HTTP/1.1"
+
+write header "User-Agent" "curl/7.37.1"
+write header "Host" "localhost:8110"
```
jitsni: "write header host" would do

service/http.proxy/src/test/scripts/org/kaazing/gateway/service/http/proxy/http.proxy.path.rpt
```diff
@@ -0,0 +1,60 @@
...
+write version "HTTP/1.1"
+
+write header "User-Agent" "curl/7.37.1"
+write header "Host" "localhost:8110"
+write header "Accept" "*/*"
+write close
+
+read status "200" "OK"
+read version "HTTP/1.1"
+read header "Content-Length" "17"
```
jitsni: You can just ignore reading most of the headers since the test is testing path.

service/http.proxy/src/test/scripts/org/kaazing/gateway/service/http/proxy/http.proxy.path.rpt
```diff
@@ -0,0 +1,60 @@
...
+read header "Connection" "close"
+read header "Accept" "*/*"
+read closed
+
+write status "200" "OK"
+write version "HTTP/1.1"
+
+write header "Content-Type" "text/html;charset=ISO-8859-1"
+write header "Connection" "close"
+write header content-length
```
jitsni: You want to write those 17 bytes here as the client is expecting Content-Length: 17

msalavastru:
@jitsni I have implemented the changes you suggested and pushed the latest changes. The 2 k3po tests are now passing. Thanks for your input.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -94,10 +95,31 @@ public void initializeSession(IoSession session, ConnectFuture future) {
...
+            URI connectRequestURI = connectSession.getRequestURI().resolve(acceptPathInfo);
+            connectSession.setRequestURI(connectRequestURI);
             processRequestHeaders(acceptSession, connectSession);
         }
 
+        private boolean validateRequestPath() {
+            URI requestURI = acceptSession.getRequestURI();
+            String acceptPath = acceptSession.getServicePath().getPath();
+            String requestPath = requestURI.normalize().getPath();
+            if (requestPath.indexOf("/\\.\\./") == 0 || requestPath.indexOf(acceptPath) != 0) {
```
jitsni: I think the whole if condition can be eliminated (infact indexOf is wrong check as it checks anywhere in the path). Doesn't the following work ?
```
return requestPath.startsWith(acceptPath)
```

server/src/main/java/org/kaazing/gateway/server/context/resolve/GatewayContextResolver.java
```diff
@@ -670,6 +670,7 @@ private DefaultServiceDefaultsContext resolveServiceDefaults(ServiceDefaultsType
                 }
                 serviceRegistry.register(acceptURI, serviceContext);
             }
+
```
jitsni: You may want to remove this file from PR.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -94,10 +95,31 @@ public void initializeSession(IoSession session, ConnectFuture future) {
             HttpConnectSession connectSession = (HttpConnectSession) session;
             connectSession.setVersion(acceptSession.getVersion());
             connectSession.setMethod(acceptSession.getMethod());
-            connectSession.setRequestURI(acceptSession.getRequestURI());
+            if (!validateRequestPath()) {
```
jitsni: Can you see what happens to the future object in this method ? I need to check whether we need to fail the ConnectFuture in case of validation doesn't go through.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -94,10 +95,31 @@ public void initializeSession(IoSession session, ConnectFuture future) {
...
+                acceptSession.setStatus(CLIENT_NOT_FOUND);
+                acceptSession.close(false);
+                return;
+            }
+            String acceptPathInfo = acceptSession.getPathInfo().toString();
+            if (acceptPathInfo.indexOf("/") == 0) {
```
jitsni: Is this if condition needed at all ? If the accept session requstURI doesn't contain /, then it would be wrong at http transport level.

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpConnectProcessor.java
```diff
@@ -71,6 +71,7 @@ protected void finishConnect(DefaultHttpSession session) {
             // create HttpRequestMessage
             HttpRequestMessage httpRequest = new HttpRequestMessage();
             httpRequest.setMethod(session.getMethod());
+
```
jitsni: You may want to remove this new line. It doesn't add any value as httpRequest is still poplulated with other details.

service/http.proxy/src/test/scripts/org/kaazing/gateway/service/http/proxy/http.proxy.path.rpt
```diff
@@ -0,0 +1,54 @@
...
+
+connect http://localhost:8110/acceptConfigPath/example/jsp/num/file.html
+connected
+
+write method "GET"
+write parameter "name" "value"
+write version "HTTP/1.1"
+
+write header "User-Agent" "curl/7.37.1"
+write header "Host" "localhost:8110"
```
jitsni: Just use "write header host"

service/http.proxy/src/test/scripts/org/kaazing/gateway/service/http/proxy/http.proxy.path.rpt
```diff
@@ -0,0 +1,54 @@
...
+read header "Via" "1.1 kaazing"
+read header "User-Agent" "curl/7.37.1"
+read header "Host" "localhost:8080"
+read header "Forwarded" "for=127.0.0.1"
+read header "Accept" "*/*"
+read closed
+
+write status "200" "OK"
+write version "HTTP/1.1"
+write close
```
jitsni: I want you to send some data.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
jitsni: I think here the tcp connection is already established to origin server, i.e connect session is established. Then we are doing the validation of accept path and deciding not to send http request. I think it is better to do the validation in the sessionOpened(), that way gateway doesn't connect to origin server for invalid requests.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -94,7 +108,15 @@ public void initializeSession(IoSession session, ConnectFuture future) {
             HttpConnectSession connectSession = (HttpConnectSession) session;
             connectSession.setVersion(acceptSession.getVersion());
             connectSession.setMethod(acceptSession.getMethod());
-            connectSession.setRequestURI(acceptSession.getRequestURI());
+            String acceptPathInfo = acceptSession.getPathInfo().toString();
+            if (acceptPathInfo.startsWith("/")) {
```
msalavastru: Looking in the rfc2396 ( at section 5.2 steps 5 and 6 ) for resolving a base URI against a reference URI, I see that if the acceptPath begins with "/", then after .resolve() the whole path in connect is removed. 
If only query is appended to the accept in config in the request, then acceptPathInfo will start with "?", that is why I used this validation.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -94,7 +108,15 @@ public void initializeSession(IoSession session, ConnectFuture future) {
...
+                acceptPathInfo = acceptPathInfo.substring(1);
+            }
+            String connectPath = connectSession.getRequestURI().getPath();
+            if (!connectPath.endsWith("/")) {
```
msalavastru: If the last segment from the connect path does not end in "/", then after .resolve() this last segment is removed.

msalavastru:
@jitsni I have updated the code with your suggestions and pushed in the latest changes. I also added 2 line comments above. Please let me know if it is ok.

PR #385, opened 2015-12-30
Uptake latest k3po version (3.0.0-alpha-6)

cmebarrow:
Uptake k3po version 3.0.0-alpha-6
This is a forward port of a similar change made to gateway.server (PR https://github.com/kaazing-private/gateway.server/pull/110)
In addition some improvements are made to the wsn and wseb specification tests to speed up execution time and make them hopefully more reliable.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
         <maven.compiler.testSource>1.8</maven.compiler.testSource>
         <maven.compiler.testTarget>1.8</maven.compiler.testTarget>
         <hazelcast.version>1.9.4.8</hazelcast.version>
-        <k3po.version>3.0.0-alpha-6</k3po.version>
+        <k3po.version>3.0.0-alpha-8</k3po.version>
         <jdom.version>1.1</jdom.version>
         <jmock.version>2.6.0</jmock.version>
         <slf4j.log4j.version>1.7.5</slf4j.log4j.version>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
         <maven.compiler.testSource>1.8</maven.compiler.testSource>
         <maven.compiler.testTarget>1.8</maven.compiler.testTarget>
         <hazelcast.version>1.9.4.8</hazelcast.version>
-        <k3po.version>3.0.0-alpha-3</k3po.version>
+        <k3po.version>3.0.0-alpha-6</k3po.version>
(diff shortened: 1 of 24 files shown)


pom.xml
```diff
@@ -28,7 +28,7 @@
         <maven.compiler.testSource>1.8</maven.compiler.testSource>
         <maven.compiler.testTarget>1.8</maven.compiler.testTarget>
         <hazelcast.version>1.9.4.8</hazelcast.version>
-        <k3po.version>3.0.0-alpha-3</k3po.version>
+        <k3po.version>3.0.0-alpha-6</k3po.version>
```
dpwspoon: Maybe upgrade to alpha 8

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/specification/ws/connector/ControlIT.java
```diff
@@ -126,21 +109,13 @@ public Object invoke(Invocation invocation) throws Throwable {
...
-                oneOf(handler).sessionClosed(with(any(IoSessionEx.class)));
-                will(new CustomAction("Latch countdown") {
-                    @Override
-                    public Object invoke(Invocation invocation) throws Throwable {
-                        latch.countDown();
-                        return null;
-                    }
-                });
+                allowing(handler).exceptionCaught(with(any(IoSessionEx.class)), with(any(Exception.class)));
+                allowing(handler).sessionClosed(with(any(IoSessionEx.class)));
```
dpwspoon: I don't think we need to allow session closed?  It should be deterministic when the tcp connection will be closed based on the chain rule,  as context is the most inner in this case it should always be open at that point.  I would believe

cmebarrow: In cases where the client sends a WsClose frame and the server does not send a WsClose response the client (WsnConnector) goes into a five second wait before abruptly closing the connection and causing exceptionCaught and session closed to be fired. Without the extra expectations the test would fail in this case. If you comment out those expectations and set up debug breakpoint on k3po.finish(), and wait there for 5 seconds before continuing, you will see the test failure. This means that we have to have these expectations to provide spurious test failures in cases where available CPU can fluctuate (i.e. Travis).

PR #392, opened 2016-01-07
Wseb session thread migration

jitsni:
When wseb session is migrated to different thread, transportSession also needs to be migrated.


cmebarrow:
@jitsni WsebSession gets aligned because of socks I presume? Could you explain a bit exactly how that plays out? When does socks realign its transport? Is it when a remote bind gets hooked up to an incoming client connection?

jitsni:
Yes, for socks remote bind. We are aligning socks transport to the connect transport.
See SocksReverseAcceptor#createReverseConnection

If socks is running over wseb and if socks migrates to new io thread, then underneath stack is also migrated. wseb is tricky because it has writer session, reader session, transport session.

transport/wseb/src/main/java/org/kaazing/gateway/transport/wseb/WsebSession.java
```diff
@@ -179,6 +179,14 @@ public WsebSession(int ioLayer,
         transportSession.setHandler(transportHandler);
     }
 
+    protected void setIoAlignment0(Thread ioThread, Executor ioExecutor) {
+        transportSession.setIoAlignment(ioThread, ioExecutor);
+        super.setIoAlignment0(ioThread, ioExecutor);
+
+        // TODO shoud we detachWriter() too so that newly attached writer gets aligned by default ?
```
cmebarrow: @jitsni Yes I think we should do this

cmebarrow: Actually given the writer is the parent we don't need to do this. I do wonder a bit about the reader.

transport/wseb/src/main/java/org/kaazing/gateway/transport/wseb/WsebSession.java
```diff
@@ -179,6 +179,14 @@ public WsebSession(int ioLayer,
...
+        // TODO writer gets aligned now since it is parent. what about reader alignment ?
```
cmebarrow: Do you have a test that fails without this fix? Writer is okay as you point out since it's the parent. For the reader, I think it might be best to do logic similar to that in WsebSession.attachReader, only without calling attachReader0.

cmebarrow:
@jitsni Done reviewing, please see my suggestion concerning aligning the reader.

transport/wseb/src/main/java/org/kaazing/gateway/transport/wseb/WsebSession.java
jitsni: Yes, the tests in https://github.com/kaazing/enterprise.gateway/pull/217 fail without this fix.

PR #393, opened 2016-01-07
Allow 0 or more spaces in parsing of wse accept content-type header

dpwspoon:
Note: this is a forward port


jitsni:
We need generic parameter parsing of http header values. 
For now, I am just merging this one.

Issue #384, opened 2015-12-29
Need better diagnostics when a ProtocolDecoderException occurs or other unexpected transport exception


cmebarrow:
If for example a WebSocket protocol error is detected, the WebSocket connection is closed, but unless log level is set to debug or finer, the only message in the Gateway log file indicating there was an issue is the generic "java.io.IOException: Early termination of IO session", which gives no indication of the underlying cause. We should include the underlying cause in this message when there is one.

To reproduce:
Run any of the failure scenario tests in org.kaazing.gateway.transport.wsn.specification.ws.connector.FragmentationIT, for example, shouldFailWebSocketConnectionWhenServerSendBinaryFrameWithPayloadFragmentedButNotContinued, after tweaking the test to print out log messages (simplest way is to force test failure by adding the line "throw new Exception();" at the end of the test method. This causes trace level of messages to be printed out. But by examining only the INFO level messages one can see that the only one indicating it was an issue is

```
540  transport.wsn.connect [New I/O worker #1] INFO  connect  - [wsn#3 /127.0.0.1:56462] EXCEPTION: java.io.IOException: Early termination of IO session
```

which does not indicate the cause of the problem (which was 

``` if
ProtocolDecoderException: Expecting CONTINUATION frame, but got BINARY frame 
```

)

The full trace level log output from running the above test is available at:
[issue_384_out.txt](https://files.zenhub.io/56943160d4ec3c2c3bf4a1b5)
It is easy to see from this, by scanning for "ProtocolDecoderException", that the only log messages mentioning that exception are DEBUG or TRACE level (exception stack traces are always part of the preceding log message).

cmebarrow:
I would suggest we add a cause exception in these cases to the IOException that we throw, and enhance LoggingFilter to into the cause exception in the EXCEPTION message when available

cmebarrow:
This was fixed by PR https://github.com/kaazing/gateway/pull/396

Archish27:
@cmebarrow Can you please guide how logging level was increased?

PR #396, opened 2016-01-11
Issue 384: provide INFO level diagnostics for exceptions like ProtocolDecoderException

cmebarrow:
**NOTE: not ready for merging yet.**

As noted in issue #384 there is no detail in the Gateway logs  for certain exceptions like ProtocolDecoderException unless one or more loggers are configured at debug or trace level. All one sees in that case is an IOException, with no clue as to the cause. This PR attempts to address this, along the lines of what was suggested in the issue. 

The test case mentioned in issue #384 results in the following log output with the changes in this PR:
[issue_384_with_fix_out.txt](https://files.zenhub.io/569432add4ec3c2c3bf4a1cc)

By scanning that file you can see that the IOException (INFO level message) includes the message from the ProtocolDecoderException as well as its stack trace.


cmebarrow:
The Travis build failure in wse spec test BinaryAsEscapedTextIT.
shouldEchoFrameWithPayloadLength127 is a known issue (see https://github.com/k3po/k3po/pull/280), unrelated to these changes, which I am working on and will fix in another PR.

PR #397, opened 2016-01-12
Added NioSystemProperty that will kill session if idle at tcp level 

dpwspoon:
org.kaazing.gateway.server.transport.tcp.IDLE_TIMEOUT which is set in seconds.


dpwspoon:
Note: this is a forward port.

PR #401, opened 2016-01-18
Modified xsd file in order to support only one <connect> for http.proxy

mgherghe:
Pending: adding test.

### DIFF
--- a/server/src/main/xsd/gateway-config-201409.xsd
+++ b/server/src/main/xsd/gateway-config-201409.xsd
@@ -968,7 +968,7 @@
                              name="accept"
                              type="anyURI"/>
 
-                    <element maxOccurs="unbounded" minOccurs="0"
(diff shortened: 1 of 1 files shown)


server/src/main/xsd/gateway-config-201409.xsd
```diff
@@ -968,7 +968,7 @@
                              name="accept"
                              type="anyURI"/>
 
-                    <element maxOccurs="unbounded" minOccurs="0"
```
jitsni: Should we have minOccurs=1 too ?

dpwspoon: Yes we should @mgherghe can you add another pull request for that

mgherghe: @dpwspoon Added [PR](https://github.com/kaazing/gateway/pull/404) for this.

PR #405, opened 2016-01-20
A minor fix for issue 183

msalavastru:
Fixed the appending of the last "/" to the connect path when there is no accept request path.


jitsni:
Can you also add a test case ?

msalavastru:
@jitsni I have added a k3po test for this scenario.

jitsni:
I ran the current patch through few inputs and they seem to be inconsistent. I think we should have clear semantics and easy to explain algo. I am proposing the following algo (it just assumes accept and connect point to dirs and that is used for resolving. That would be consistent  when path is not specified, we were resolving using root dir or '/'):
```
    static URI algo(String accept, String connect, String request) {
        URI acceptURI = URI.create(accept.endsWith("/") ? accept : accept.concat("/"));
        URI connectURI = URI.create(connect.endsWith("/") ? connect : connect.concat("/"));
        if (!request.startsWith(acceptURI.getPath())) {
            return URI.create("-");   // 404 not found
        }
        URI requestURI = URI.create(request);
        URI relativeURI = acceptURI.relativize(requestURI);
        return connectURI.resolve(relativeURI);
    }

```

With that, the following table is created. columns are (accept path, connect path), and the rows are request paths.
```
          |        /a,      /b/c|        /a,     /b/c/|       /a/,      /b/c|       /a/,     /b/c/|
===================================================================================================
        /a|                    -|                    -|                    -|                    -|
---------------------------------------------------------------------------------------------------
       /a/|                /b/c/|                /b/c/|                /b/c/|                /b/c/|
---------------------------------------------------------------------------------------------------
    /a/foo|             /b/c/foo|             /b/c/foo|             /b/c/foo|             /b/c/foo|
---------------------------------------------------------------------------------------------------
   /a/foo/|            /b/c/foo/|            /b/c/foo/|            /b/c/foo/|            /b/c/foo/|
---------------------------------------------------------------------------------------------------
```

I will run through @robinzimmermann whether it is ok with this mapping.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -108,16 +108,41 @@ public void initializeSession(IoSession session, ConnectFuture future) {
...
+
+        private URI computeConnectPath(URI connectURI) {
+            URI acceptURI = acceptSession.getServicePath();
+            URI requestURI = acceptSession.getRequestURI();
+            URI relativeURI = acceptURI.relativize(requestURI);
+
+            boolean hasExtraPath = relativeURI.getPath().length() > 0;
+            boolean requestEndsInSlash = requestURI.getPath().endsWith("/");
+            String connectPath = connectURI.getPath();
+            if (hasExtraPath || requestEndsInSlash) {
```
msalavastru: - hasExtraPath - checks for path in request ( with or without query and/or fragments )
- requestEndsInSlash - covers the scenario where the request has no path and no query.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -108,16 +108,41 @@ public void initializeSession(IoSession session, ConnectFuture future) {
...
+                connectURI = URI.create(connectPath.endsWith("/") ? connectPath : connectPath.concat("/"));
+                connectURI = connectURI.resolve(relativeURI);
+            } else if (hasQueryOrFragment(relativeURI)) {
```
msalavastru: checks if the request has query or fragment only, but no path.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyService.java
```diff
@@ -38,11 +39,35 @@ public void init(ServiceContext serviceContext) throws Exception {
...
+        Iterator<URI> acceptIterator = acceptURIs.iterator();
+        Iterator<URI> connectIterator = connectURIs.iterator();
+
+        while (acceptIterator.hasNext() && connectIterator.hasNext()) {
+            String acceptPath = acceptIterator.next().getPath();
+            String connectPath = connectIterator.next().getPath();
+            boolean acceptPathEndsInSlash = acceptPath.endsWith("/");
+            boolean connectPathEndsInSlash = connectPath.endsWith("/");
+
+            if (acceptPathEndsInSlash != connectPathEndsInSlash) {
```
msalavastru: Here we have a scenario for which this is not correct. For example with a config such as:

```
    <accept>http://localhost:8018</accept>

    <connect>http://developer.kaazing.com/b/c</connect> 

```
In GatewayContextResolver getCanonicalURI will append "/" to the path in accept, and the above check will throw the exception, as the path in connect has no trailing "/". How do we now if the request adeed a "/" or not so that we can add it or not to the connect. Do we want that for the above config?

msalavastru:
@jitsni I have pushed my latest changes based on the latest requirements as per the table in the [user story](https://github.com/kaazing/roadmap/issues/515#issuecomment-174077775). I have added 3 comments to the changes. Please let me know what you think.

I added k3po tests for all scenarios, except the ones where gateway throws the exception at startup. Waiting for you feedback on the comment above.

Testing with Wireshark, all the scenarios in the table are valid.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
jitsni: Can you try 
```
        private URI computeConnectPath(URI connectURI) {
            String acceptPath = acceptSession.getServicePath().getPath();
            String requestUri = acceptSession.getRequestURI().toString();
            String connectPath = connectURI.getPath();
            return URI.create(connectPath + requestUri.substring(acceptPath.length()));
        }
```

msalavastru:
@jitsni I have implemented the changes that you suggested and updated the PR. The algorithm works as per requirements. I have tested the scenarios from the user story with Wireshark.

What remains to clarify is the gateway startup throwing an error with the scenario described in my comment above regarding one of the accept or connect not having any path and the other having path, but not ending in "/".

jitsni:
I would assume that any http uri has some path. So, http://example.com has / as path whether we specify in the config or not. So let us not do anything special in proxy code.

dpwspoon:
Great work on this everyone.  Thanks @msalavastru for making sure we got the design right, and asking great questions.  Thanks @jitsni for following through and getting the ground work set.

PR #406, opened 2016-01-20
Added input area for choosing which URL to connect to

mgherghe:


dpwspoon:
This is ok.  

We don't yet know hot the product will actually be shipped or what demoes it will include or what they will look like. @nemigaservices is in charge of getting the design for that. Until that point we can't say what the demo should or should not have

PR #407, opened 2016-01-21
Forward Port issue266  from 4.0 to  CE/EE

mjolie:


util/src/main/java/org/kaazing/gateway/util/GL.java
```diff
@@ -20,6 +20,8 @@
 
 public class GL {
 
+    public static final String CLUSTER_LOGGER_NAME = "ha";
```
dpwspoon: @cmebarrow Does this seem correct?

dpwspoon:
LGTM, waiting for feedback from Chris.

util/src/main/java/org/kaazing/gateway/util/GL.java
dpwspoon: @mjolie, suggest moving this to: ClusterContext class.  GL class is an antipattern for logging (inefficient because every call looks up the logger).  Sorry for miss directing you

dpwspoon:
LGTM other then comment

util/src/main/java/org/kaazing/gateway/util/GL.java
mjolie: But all the log statements are already using the GL logger ?   How is a static constant located in this class any different than in other class ?

dpwspoon: That was just Chris Barrows suggestion when I talked to him.  I think the cluster context interface makes more sense anyways as ha really has to do with that

dpwspoon:
@mjolie, it looks like you removed the tests, not moved them.  I think we will still need them

mjolie:
@mjolie was not finished with my commits,   but back to the logging issue, it can not be a class in gateway.server , need to avoid circular dependency

mjolie:
The build problems have nothing to do with my changes

service/http.balancer/pom.xml
```diff
@@ -98,6 +98,24 @@
             <groupId>org.kaazing</groupId>
             <artifactId>k3po.junit</artifactId>
         </dependency>
+        <dependency>
```
dpwspoon: It would be better to move these as a dependency of the maven-plugin:
```xml
            <plugin>
                 <artifactId>k3po-maven-plugin</artifactId>
</dependencies
    <dependency>
             <artifactId>specification.http</artifactId>
             <version>${k3po.version}</version>
          </dependency>
          <dependency>
              <artifactId>specification.httpx</artifactId>
              <version>${k3po.version}</version>
          </dependency>
          <dependency>
              <artifactId>specification.ws</artifactId>
              <version>${k3po.version}</version>
          </dependency>
             </plugin>
```

dpwspoon:
Looks good, one minor nit about the dependency management

service/http.balancer/pom.xml
mjolie: what does that do ?  I and aren't all these projects in git hub now ?  I  checked the other modules and I do not see an example of the dependency under this plug in ??

dpwspoon: You can see this being done in the top level pom: https://github.com/kaazing/gateway/blob/develop/pom.xml.  I included an excerpt at the bottom.  And perhaps you don't need to specify them in this pom at all because of that (you will have to confirm that independently).  

> what does that do ?

The main dependency section of the pom is pulled in while compiling the code and test.  If you put the scope as test, it is only pulled in while compiling the tests.  

The scripts in the specification project are actually only needed at run time, and specifically only for the runtime of the k3po driver which is launched during the plugin.  There isn't really a way to extend the scripts, and there is generally no code in them that you need while compiling the tests.

The dependency section of the plugin is only pulled in when running the plugin.  With k3po, the junit part of the test tells the driver (which is launched by the plugin) to find a script.  The driver needs to search its class path.

So it's just a minor fix saying, eclipse and test compiling does not need that jar and dependencies

```
  <plugin>
                <dependencies>
                    <dependency>
                        <artifactId>specification.http</artifactId>
                        <version>${k3po.version}</version>
                    </dependency>
                    <dependency>
                        <artifactId>specification.httpx</artifactId>
                        <version>${k3po.version}</version>
                    </dependency>
                    <dependency>
                        <artifactId>specification.ws</artifactId>
                        <version>${k3po.version}</version>
                    </dependency>
                    <dependency>
                        <artifactId>specification.tcp</artifactId>
                        <version>${k3po.version}</version>
                    </dependency>
                    <dependency>
                        <artifactId>specification.wse</artifactId>
                        <version>${k3po.version}</version>
                    </dependency>
                    <dependency>
                        <artifactId>specification.x-kaazing-ping-pong</artifactId>
                        <version>${k3po.version}</version>
                    </dependency>
                </dependencies>
                <artifactId>k3po-maven-plugin</artifactId>
                <version>${k3po.version}</version>
                <configuration>
                    <skipTests>${skipITs}</skipTests>
                </configuration>
                <executions>
                    <execution>
                        <goals>
                            <goal>start</goal>
                            <goal>stop</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
```

mjolie: just fyi the gateway.transport.wsn  pom.xml includes some of these and they are not under the plugin and the versions are hard coded:

<dependency>
            <artifactId>specification.ws</artifactId>
            <version>2.0.1</version>
            <scope>test</scope>
        </dependency>
        <dependency>
            <artifactId>specification.x-kaazing-idle-timeout</artifactId>
            <version>2.2.0</version>
            <scope>test</scope>
        </dependency>
        <dependency>
            <artifactId>specification.x-kaazing-ping-pong</artifactId>
            <version>2.2.0</version>
            <scope>test</scope>
        </dependency>

mjolie: @dpwspoon: just fyi the gateway.transport.wsn  pom.xml includes some of these and they are not under the plugin and the versions are hard coded:
``` xml
<dependency>
            <artifactId>specification.ws</artifactId>
            <version>2.0.1</version>
            <scope>test</scope>
        </dependency>
        <dependency>
            <artifactId>specification.x-kaazing-idle-timeout</artifactId>
            <version>2.2.0</version>
            <scope>test</scope>
        </dependency>
        <dependency>
            <artifactId>specification.x-kaazing-ping-pong</artifactId>
            <version>2.2.0</version>
            <scope>test</scope>
        </dependency>
```

dpwspoon: Noted, thanks.  This is a common mistake I try and correct when I see it.  Also the generally in " there is generally no code in them that you need while compiling the tests" is not always true.

dpwspoon:
Latest build passed: https://s3.amazonaws.com/archive.travis-ci.org/jobs/105762465/log.txt.  Not sure why didn't update.

PR #412, opened 2016-01-29
Issue.183 added 2 new scenarios for validating accept and connect trailing slashes

msalavastru:


msalavastru:
@sbadugu I have created a pull request for the 2 new extra scenarios that we agreed upon in our last meeting. Can you please assign someone for code review? Also, are we ok with the error messages issued?

There are now 3 scenarios that are checked in relation to the trailing slashes in accept and connect:

- accept URL has no path ( slash is added by default for empty path), but connect URL has a path with no trailing slash, for example: 

```
    <name>service name</name>
    <accept>http://localhost:8018</accept>
    <connect>http://localhost:8018/b/c</connect> 
```
Here the error message issued is: **Please fix the gateway configuration file for service 'service name'. Using accept URL 'http://localhost:8018/' ( empty path in accept must be a slash ) and connect URL 'http://localhost:8018/b/c'. Accept and connect must both either have trailing slashes, or both must not end in slashes.**

- accept URL has a path with no trailing slash, but connect URL has no path ( slash is added by default for empty path), for example:

```
    <name>service name</name>
    <accept>http://localhost:8018/a</accept>
    <connect>http://localhost:8018</connect> 
```
Here the error message issued is: **Please fix the gateway configuration file for service 'service name'. Using accept URL 'http://localhost:8018/a' and connect URL 'http://localhost:8018/' ( empty path in connect must be a slash ). Accept and connect must both either have trailing slashes, or both must not end in slashes.**

- accept URL and connect URL both have paths, but do not match in trailing slashes ( one with, the other without ), for example:

```
    <accept>http://localhost:8018/a</accept>
    <connect>http://localhost:8018/b/</connect> 
```
```
    <accept>http://localhost:8018/a/</accept>
    <connect>http://localhost:8018/b</connect> 
```
Here the error message issued is: **Please fix the gateway configuration file for service 'service name'. Accept and connect must both either have trailing slashes, or both must not end in slashes.**

sbadugu:
@jitsni Can you please review this PR ?

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyService.java
```diff
@@ -54,16 +62,24 @@ private void checkForTrailingSlashes(ServiceContext serviceContext) {
...
+                        serviceContext.getServiceName(), acceptURI, connectURI));
+            } else if (acceptPathNoTrailingSlash && connectPathIsSlash) {
+                throw new IllegalArgumentException(String.format(CONNECT_PATH_IS_SLASH_ERROR_MESSAGE,
+                        serviceContext.getServiceName(), acceptURI, connectURI));
+            } else if (acceptPathNoTrailingSlash != connectPathNoTrailingSlash) {
                 throw new IllegalArgumentException(
-                        "Please fix the gateway configuration file for service '" + serviceContext.getServiceName()
-                                + "'. Accept and connect must both either have trailing slashes,"
-                                + " or both must not end in slashes.");
+                        String.format(ACCEPT_CONNECT_ERROR_MESSAGE, serviceContext.getServiceName()));
```
jitsni: I would go with

```
    private void checkForTrailingSlashes(ServiceContext serviceContext) {
        Collection<URI> acceptURIs = serviceContext.getAccepts();
        Collection<URI> connectURIs = serviceContext.getConnects();

        assert acceptURIs.size() == 1;
        assert connectURIs.size() == 1;

        URI acceptURI = acceptURIs.iterator().next();
        URI connectURI = connectURIs.iterator().next();
        String acceptPath = acceptURI.getPath();
        String connectPath = connectURI.getPath();

        boolean acceptPathIsSlash = acceptPath.endsWith("/");
        boolean connectPathIsSlash = connectPath.endsWith("/");
        if (acceptPathIsSlash ^ connectPathIsSlash) {
            if (acceptPathIsSlash) {
                URI connectURIWithSlash = connectURI.resolve(connectPath + "/");
                throw new IllegalArgumentException("Configure connect URI as " + connectURIWithSlash + " since accept URI " + acceptURI + " ends with a trailing /");
            } else {
                String noTrailingSlash = connectPath.substring(0, connectPath.length()-1);
                URI connectURIWithNoSlash = connectURI.resolve(noTrailingSlash);
                throw new IllegalArgumentException("Configure connect URI as " + connectURIWithNoSlash + " since accept URI " + acceptURI + " doesn't end with a trailing /");
            }
        }
    }
```

If you want you can add additional ". Both accept and connect URIs either end with trailing / or not." to IllegalArgumentException.

msalavastru: @jitsni I tried using the following config with the above implementation:

```
<accept>http://localhost:8018/a</accept>
<connect>http://localhost:8018</connect> 
```
and the following error message is being displayed
**Configure connect URI as http://localhost:8018/ since accept URI http://localhost:8018/a doesn't end with a trailing /**

In this case adding a trailing slash in connect will not change the triggered error since the path is empty and replaced by a trailing slash. The only change to pass the validation will be to add a / in accept, but the error message does not indicate this.

jitsni:
Good point. Just add one message that " accept uri = "...", connectURI="...". Either both URI should end with / or not." That way they would know what gateway interprets and what to configure.

msalavastru:
@jitsni I have updated the logic with the error message you suggested and pushed the updates.

PR #413, opened 2016-02-01
Httpxe.spec.tests

jitsni:


jitsni:
Do NOT merge yet (Latest k3po requires changes in wse spec tests)

transport/http/src/test/java/org/kaazing/gateway/transport/http/HttpAcceptorRule.java
```diff
@@ -0,0 +1,107 @@
...
+import org.kaazing.gateway.transport.nio.internal.NioSocketAcceptor;
+import org.kaazing.gateway.transport.nio.internal.NioSocketConnector;
+import org.kaazing.gateway.util.scheduler.SchedulerProvider;
+
+import java.net.URI;
+import java.util.Collections;
+
+
+/**
+ * Declaring an instance of this class as a @Rule causes the gateway to be
```
cmebarrow: acceptor not gateway (several places)

transport/http/src/test/java/org/kaazing/gateway/transport/http/HttpAcceptorRule.java
```diff
@@ -0,0 +1,107 @@
...
+
+                tcpConnector = (NioSocketConnector)transportFactory.getTransport("tcp").getConnector();
+                tcpConnector.setResourceAddressFactory(addressFactory);
+                tcpConnector.setBridgeServiceFactory(serviceFactory);
+                tcpConnector.setTcpAcceptor(tcpAcceptor);
+
+                httpAcceptor = (HttpAcceptor)transportFactory.getTransport("http").getAcceptor();
+                httpAcceptor.setBridgeServiceFactory(serviceFactory);
+                httpAcceptor.setResourceAddressFactory(addressFactory);
+                httpAcceptor.setSchedulerProvider(schedulerProvider);
```
cmebarrow: You can use transportFactory.injectResources to simplify the code and avoid having to create the tcpAcceptor and tcpConnector. That method automatically injects resources which are connectors or acceptors:
```
                schedulerProvider = new SchedulerProvider();
                addressFactory = ResourceAddressFactory.newResourceAddressFactory();
                TransportFactory transportFactory = TransportFactory.newTransportFactory(Collections.<String, Object> emptyMap());

                Map<String, Object> resources = new HashMap<String, Object>();
                resources.put("schedulerProvider",  schedulerProvider);
                resources.put("configuration",  new Properties());
                resources.put("bridgeServiceFactory", new BridgeServiceFactory(transportFactory));
                resources.put("resourceAddressFactory", addressFactory);
                transportFactory.injectResources(resources);

                httpAcceptor = (WsnAcceptor)transportFactory.getTransport("http").getAcceptor();

                base.evaluate();
```

cmebarrow: Actually don't do this yet, it may not be working!

transport/http/src/test/java/org/kaazing/gateway/transport/http/specification/httpxe/RequestsIT.java
```diff
@@ -0,0 +1,180 @@
...
+
+import java.net.URI;
+
+import static org.kaazing.test.util.ITUtil.createRuleChain;
+
+/**
+ * Defines how httpxe will deal with http methods.
+ */
+public class RequestsIT {
+
```
cmebarrow: I think we should be verifying that the HttpSession has the correct method, headers, etc. For example you could use a mock object instead of the concrete object for the acceptHandler with a matches expectation on HttpSession. Each test may need its own separate expectations (or at least a special matcher to verify the expected method and headers).

transport/http/src/test/java/org/kaazing/gateway/transport/http/specification/httpxe/OriginSecurityIT.java
```diff
@@ -0,0 +1,121 @@
...
+    @Test
+    @Specification("origin.request.using.referer/request")
+    public void shouldPassWithOnlyRefererAndXoriginRequest() throws Exception {
+        test(HTTP_ADDRESS);
+    }
+
+    @Test
+    @Specification("x.origin.header.not.identical.to.origin.header/request")
+    public void shouldPassWhenXoriginHeaderDiffersFromOriginHeader() throws Exception {
+        test(HTTP_ADDRESS);
```
cmebarrow: Do we need to check that we are using the correct value for the origin? According to the spec in this case we would use the value of the X-Origin-Header. Same applies for the other test methods.

transport/http/src/test/java/org/kaazing/gateway/transport/http/specification/httpxe/OriginSecurityIT.java
```diff
@@ -0,0 +1,121 @@
...
+    public void shouldPassWithAccessControlWithKacParameter() throws Exception {
+        test(HTTP_ADDRESS);
+    }
+
+    @Test
+    @Specification("x.origin.encoded.request.header/request")
+    public void shouldPassWithEncodedXoriginRequest() throws Exception {
+        test(HTTP_ADDRESS);
+    }
+
```
cmebarrow: Do we have any negative tests? Spec says "If the authoritative source origin cannot be determined, then the server MUST response with status code 400 Bad Request"

transport/http/src/test/java/org/kaazing/gateway/transport/http/specification/httpxe/ResponsesIT.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    private void testWrappedResponse(ResourceAddress address, HttpStatus status) throws Exception {
+        final IoHandler acceptHandler = new IoHandlerAdapter<HttpAcceptSession>() {
+            @Override
+            protected void doSessionOpened(HttpAcceptSession session) throws Exception {
+                // Removing the filter that doesn't keep the wrapped responses
+                IoFilterChain filterChain = session.getParent().getFilterChain();
+                if (filterChain.contains(HttpAcceptFilter.CONDITIONAL_WRAPPED_RESPONSE.filterName())) {
+                    filterChain.remove(HttpAcceptFilter.CONDITIONAL_WRAPPED_RESPONSE.filterName());
```
cmebarrow: Why do we have to do this? Clients cannot do it.

transport/http/src/test/java/org/kaazing/gateway/transport/http/specification/httpxe/ResponsesIT.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    @Test
+    @Specification("wrapped.400.response.in.200/request")
+    public void shouldPassWithWrapped400ResponseIn200() throws Exception {
+        testWrappedResponse(HTTPXE_ADDRESS, HttpStatus.CLIENT_BAD_REQUEST);
+    }
+
+    @Test
+    @Specification("unwrapped.404.response/request")
```
cmebarrow: Do we need a test to show we don't attempt to wrap 404?

cmebarrow:
@jitsni I am done reviewing. Ping me when you want me to review again.

transport/http/src/test/java/org/kaazing/gateway/transport/http/HttpAcceptorRule.java
jitsni: fixing it.

transport/http/src/test/java/org/kaazing/gateway/transport/http/HttpAcceptorRule.java
jitsni: Fixing it.

transport/http/src/test/java/org/kaazing/gateway/transport/http/specification/httpxe/RequestsIT.java
jitsni: Most of the tests are for HTTP methods (path encoded etc). Fixing it.

transport/http/src/test/java/org/kaazing/gateway/transport/http/specification/httpxe/ResponsesIT.java
jitsni: Added a test

cmebarrow:
@jitsni Please finish answering my comments so I can merge

transport/http/src/test/java/org/kaazing/gateway/transport/http/specification/httpxe/OriginSecurityIT.java
jitsni: Checking the Origin header in handler

jitsni:
Review this PR along with K3PO PR (https://github.com/k3po/k3po/pull/311)
Do not merge it as it needs K3PO integration.

transport/http/src/test/java/org/kaazing/gateway/transport/http/specification/httpxe/OriginSecurityIT.java
```diff
@@ -0,0 +1,136 @@
...
+    public final TestRule chain = createRuleChain(acceptor, k3po);
+
+    @Test
+    @Specification("request.with.origin.header/request")
+    public void shouldPassWithOriginRequestHeader() throws Exception {
+        test(HTTP_ADDRESS, "http://source.example.com:80");
+    }
+
+    @Test
+    @Specification("bad.request.with.origin.header/request")
```
cmebarrow: Need to rename the script (bad -> unauthorized)

transport/http/src/test/java/org/kaazing/gateway/transport/http/specification/httpxe/OriginSecurityIT.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    @Test
+    @Specification("request.with.origin.header.and.x.origin.header/request")
+    public void shouldPassWithOriginAndXoriginRequests() throws Exception {
+        test(HTTP_ADDRESS, "http://source.example.com:80");
+    }
+
+    @Test
+    @Specification("bad.request.with.origin.header.and.x.origin.header/request")
```
cmebarrow: Ditto

transport/http/src/test/java/org/kaazing/gateway/transport/http/specification/httpxe/OriginSecurityIT.java
```diff
@@ -0,0 +1,136 @@
...
+    public void shouldFailWithOriginAndXoriginRequests() throws Exception {
+        test(HTTP_ADDRESS, "http://source.example.com:80");
```
cmebarrow: test method sets status 200 but script expects 403 so how can this test pass?

transport/http/src/test/java/org/kaazing/gateway/transport/http/specification/httpxe/OriginSecurityIT.java
```diff
@@ -0,0 +1,136 @@
...
+    @Test
+    @Specification("request.with.origin.header/request")
+    public void shouldPassWithOriginRequestHeader() throws Exception {
+        test(HTTP_ADDRESS, "http://source.example.com:80");
+    }
+
+    @Test
+    @Specification("bad.request.with.origin.header/request")
+    public void shouldFailWithOriginRequestHeader() throws Exception {
+        test(HTTP_ADDRESS, "http://unknownsource.example.com:80");
```
cmebarrow: test method sets status 200 but script expects 403 so how can this test pass?

transport/http/src/test/java/org/kaazing/gateway/transport/http/specification/httpxe/OriginSecurityIT.java
```diff
@@ -0,0 +1,136 @@
...
+        test(HTTP_ADDRESS, "http://localhost:8000");
+    }
+
+    @Test
+    @Specification("x.origin.encoded.request.header/request")
+    public void shouldPassWithEncodedXoriginRequest() throws Exception {
+        test(HTTP_ADDRESS, "http://source.example.com:80");
+    }
+
+    private void test(ResourceAddress address, String origin) throws Exception {
```
cmebarrow: You will probably need a variant of this for the failure cases. Consider renaming to shouldPass (then the other can be shouldFail)

transport/http/src/test/java/org/kaazing/gateway/transport/http/specification/httpxe/OriginSecurityIT.java
jitsni: Before reaching the handler, one of the filters return 403.

transport/http/src/test/java/org/kaazing/gateway/transport/http/specification/httpxe/OriginSecurityIT.java
jitsni: Before reaching the handler, one of the filters send 403.

cmebarrow:
@jitsni Looking good except I think we are still missing some httpxe test cases for OriginSecurityIT. Sorry I didn't spot that when reviewing your k3po changes.  You added not authorized ones but we are still not covering bad request cases where the authoritative origin cannot be determined (when we should reply status 400). I think the cases we need are:
- request method is path-encoded, Origin header not present, X-Origin present, X-Origin-http-xxx header not present or xxx part does not match X-Origin
- request method is path-encoded, Origin header not present, X-Origin present, X-Origin-http-xxx header xxx part does not match X-Origin
- X-Origin present and differs from the Origin header, and the request is not path-encoded, and the request is NOT determined to be a same-origin request by comparing value of the Origin header with the Host header
- X-Origin header is present and the Origin header is not present, and the request is not path-encoded, and the request is NOT determined to be a same-origin request by comparing value of the Referer header with the Host header

jitsni:
For me, those cases are unauthorized (whether origin is there, not there, or cannot be determined). That was the reason I didn't add specific tests for them. I can add them. May be spec needs to be changed to say 403 is sent.

PR #415, opened 2016-02-03
Support proxying to application servers that generate dynamic resource addresses

msalavastru:
* `Forwarded`
* `X-Forwarded-Host`
* `X-Forwarded-For`
* `X-Forwarded-Server`
* `X-Forwarded-Proto`


msalavastru:
@jitsni Looking through the code I found that there is a partial implementation for the `Forwarded` header in HttpSubjectSecurityFilter. Is this a good place to implemented the other part of the `Forwarded` header and the `X-Forwarded` headers?

Do we want to allow anonymity to not populate the headers as an option? Do we want to use obfuscated Identifier for this?

msalavastru:
@jitsni This is how the X-forwarded headers are populated in my implementation:

```
X-Forwarded-Server=[192.168.1.11, 192.168.1.12, 192.168.1.13]
X-Forwarded-Proto=[http, http, http]
X-Forwarded-Port=[8001, 8002, 8003]
X-Forwarded-Host=[gateway1, gateway2, gateway3]
```

the initial client request comes from 192.168.1.13 ( same as gateway 3), so the `X-forwarded-For` looks like: 
```
X-Forwarded-For=[192.168.1.13, 192.168.1.11, 192.168.1.12]
```

The forwarded header looks like this:

```
Forwarded=[server=192.168.1.11, server=192.168.1.12, server=192.168.1.13; port=8001, port=8002, port=8003; proto=http, proto=http, proto=http; for=192.168.1.13, for=192.168.1.11, for=192.168.1.12; host=gateway1, host=gateway2, host=gateway3]
```

If we want to allow to disable the usage of forwarded headers, is it ok if we use InternalSystemProperty to set a flag in GATEWAY_OPTS ?

I will also move the logic to the HttpProxyServiceHandler.

msalavastru:
@jitsni Looking at [section 5.7 Example usage](https://tools.ietf.org/html/rfc7239#section-7.5) from rfc 7239, it seems that the Forwarded header should only have 1 `by`, 1 `port`, 1 `host`, 1 `proto` parameter in a scenario where the request goes through multiple proxies. The `for` should be populated for each successive pass ( client, proxy_1... proxy_n-1), and the other parameters should be populated only with the info from the final proxy making the request to the origin server. Is this assumption correct? What do you think?

jitsni:
* Add Forwarded to HttpHeaderDecodingState.COMMA_SEPARATED_HEADERS
* Each proxy service adds a Forwarded header ( since it is comma separated header, all the Forwarded headers can be combined with "," ) if it wants to.

* So in that sense, the posted Forwarded header is wrong.
* For example, if the message is going through two proxy services:
```
  Forwarded: for=for1;proto=proto1;by=by1;host=host1
  Forwarded: for=for2;proto=proto2;by=by2;host=host2
```

 or same as
```
  Forwarded: for=for1;proto=proto1;by=by1;host=host1, for=for2;proto=proto2;by=by2;host=host2
```

Note that some of the parameters may be absent (depends on what a proxy service want to divulge)
* So just think that a proxy service adds a Forwarded header and not worry about combining the parameters across.

msalavastru:
@jitsni I have made the changes that you suggested and pushed them into the PR:

- added the forwarded and x-forwarded headers to HttpHeaderDecodingState.COMMA_SEPARATED_HEADERS
- changed the logic to compose the headers as per your comment above
- added a property to the http.proxy service in xsd and used it's value in HttpProxyServiceHandler

Please let me know what you think.

msalavastru:
@jitsni I have added two k3po tests: one for a standalone proxy, and one for 3 chained proxies, the first one not using forwarded headers.
I just realized that disabling the use of forwarded headers by default affects the k3po tests for http proxy that expect to read a `Forwarded` header.

server/src/main/xsd/gateway-config-201409.xsd
```diff
@@ -1516,6 +1516,24 @@
         </complexContent>
     </complexType>
 
+    <complexType name="httpProxyServicePropertiesType">
```
jitsni: Remember to have these changes in enterprise.gateway schema after this PR

server/src/main/xsd/gateway-config-201409.xsd
```diff
@@ -1516,6 +1516,24 @@
...
+        <complexContent>
+            <restriction base="gateway:ServicePropertiesType">
+                <sequence>
+                    <element name="use-forwarded" maxOccurs="1" minOccurs="0">
```
jitsni: @robinzimmermann @jfallows can you review the config change ?

jfallows:
When `use-forwarded` is `disabled` (default) what is the correct behavior when our proxy encounters a request that already has `Forwarded` or `X-Forwarded-...` headers?

Looking at the code, it seemed that we might be removing those headers from the inbound request before proxying the request to the HTTP server behind our `http.proxy` service.

Perhaps there are really 3 states for `use-forwarded` here?
i.e. `inject`, `ignore`, `exclude` rather than just `enabled` and `disabled`?

jitsni:
My initial thought was enabled will inject a new Forwarded header and disabled doesn't inject a new header (but doesn't remove any existing Forwarded header). 

RFC also suggests that some proxies may *remove* Forwarded headers. So three
options should work :
inject - injects a new header
ignore - doesn't inject a new header (existing Forwarded headers go through)
exclude - removes all the existing Forwarded headers.

msalavastru:
@jitsni I have updated the xsd to use the 3 new proposed states, and updated the corresponding code in the gateway. Can you please have a look and let me know what you think?

If this is ok, I will also need to update the k3po scripts for http proxy that expect to receive a forwarded header.

server/src/main/xsd/gateway-config-201409.xsd
jitsni: We couldn't express well enough in schema for other http.proxy service properties. So we are not using something like httpProxyServicePropertiesType. Remove the schema type and  validate in the code.

jitsni:
Can you resolve the conflicts.

service/http.proxy/src/test/scripts/org/kaazing/gateway/service/http/proxy/http.proxy.use.forwarded.headers.rpt
```diff
@@ -0,0 +1,63 @@
...
+
+accept http://localhost:8080/
+accepted
+connected
+
+read method "GET"
+read header "Via" "1.1 kaazing"
+read header "User-Agent" "curl/7.37.1"
+read header "Host" "localhost:8080"
+read header "Forwarded" "for=127.0.0.1;by=127.0.0.1;proto=http;host=localhost;port=8110"
```
jitsni: I don't see RFC specifying any "port" parameter. host parameter itself includes it.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -234,11 +258,116 @@ private static void processRequestHeaders(HttpAcceptSession acceptSession, HttpC
...
+     * 
+     * @param connectSession
+     */
+    private static void excludeForwardedHeaders(HttpConnectSession connectSession) {
+        connectSession.clearWriteHeaders(HEADER_FORWARDED);
+        connectSession.clearWriteHeaders(HEADER_X_FORWARDED_FOR);
+        connectSession.clearWriteHeaders(HEADER_X_FORWARDED_SERVER);
+        connectSession.clearWriteHeaders(HEADER_X_FORWARDED_HOST);
+        connectSession.clearWriteHeaders(HEADER_X_FORWARDED_PROTO);
+        connectSession.clearWriteHeaders(HEADER_X_FORWARDED_PORT);
```
jitsni: I think you should remove X-Forwarded-Port header and related code.

service/http.proxy/src/test/scripts/org/kaazing/gateway/service/http/proxy/http.proxy.use.forwarded.headers.chained.proxies.rpt
```diff
@@ -0,0 +1,73 @@
...
+connected
+
+read "GET / HTTP/1.1\r\n"
+read "X-Forwarded-Server: 127.0.0.1\r\n"
+read "X-Forwarded-Server: 127.0.0.1\r\n"
+read "X-Forwarded-Proto: http\r\n"
+read "X-Forwarded-Proto: http\r\n"
+read "X-Forwarded-Port: 8112\r\n"
+read "X-Forwarded-Port: 8113\r\n"
+read "X-Forwarded-Host: localhost\r\n"
```
jitsni: X-Forward-Host is like host header, it will have port also

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpHeaders.java
```diff
@@ -32,6 +32,11 @@
     String HEADER_X_CREATE_ENCODING = "X-Create-Encoding";
     String HEADER_X_ACCEPT_COMMANDS = "X-Accept-Commands";
     String HEADER_FORWARDED = "Forwarded";
+    String HEADER_X_FORWARDED_FOR = "X-Forwarded-For";
+    String HEADER_X_FORWARDED_SERVER = "X-Forwarded-Server";
+    String HEADER_X_FORWARDED_HOST = "X-Forwarded-Host";
+    String HEADER_X_FORWARDED_PROTO = "X-Forwarded-Proto";
+    String HEADER_X_FORWARDED_PORT = "X-Forwarded-Port";
```
jitsni: Remove "X-Forwarded-Port" and related functionality.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -93,14 +112,15 @@ private boolean validateRequestPath(DefaultHttpSession acceptSession) {
...
-     * Initializer for connect session. It adds the processed accept session headers
-     * on the connect session
+     * Initializer for connect session. It adds the processed accept session headers on the connect session
      */
     private static class ConnectSessionInitializer implements IoSessionInitializer<ConnectFuture> {
         private final DefaultHttpSession acceptSession;
+        private final String forwardedProperty;
 
-        ConnectSessionInitializer(DefaultHttpSession acceptSession) {
+        ConnectSessionInitializer(DefaultHttpSession acceptSession, String forwardedProperty) {
```
jitsni: In the latest version, ConnectSessionInitializer is not static any more. You can keep useForwarded config info at handler level.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -132,10 +152,11 @@ private URI computeConnectPath(URI connectURI) {
         @Override
         public void operationComplete(ConnectFuture future) {
             if (future.isConnected()) {
-                DefaultHttpSession connectSession = (DefaultHttpSession)future.getSession();
+                DefaultHttpSession connectSession = (DefaultHttpSession) future.getSession();
 
                 if (LOGGER.isTraceEnabled()) {
-                    LOGGER.trace("Connected to " + getConnectURIs().iterator().next() + " ["+acceptSession+"->"+connectSession+"]");
+                    LOGGER.trace("Connected to " + getConnectURIs().iterator().next() + " [" + acceptSession + "->"
```
jitsni: Now that you are changing this line for formatting, s/getConnectURIs().iterator().next()/connectURI. same goes below.
Then I suppose you could put that in one line

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -234,11 +258,116 @@ private static void processRequestHeaders(HttpAcceptSession acceptSession, HttpC
...
+            connectSession.addWriteHeader(HEADER_X_FORWARDED_PROTO, protocol);
+            forwardedSb.append(format("%s=%s;", FORWARDED_PROTO, protocol));
+
+            String host = externalURI.getHost();
+            connectSession.addWriteHeader(HEADER_X_FORWARDED_HOST, host);
+            forwardedSb.append(format("%s=%s;", FORWARDED_HOST, host));
+
+            String port = format("%d", externalURI.getPort());
+            connectSession.addWriteHeader(HEADER_X_FORWARDED_PORT, port);
+            forwardedSb.append(format("%s=%s", FORWARDED_PORT, port));
```
jitsni: I think StringBuilder can be avoided. Can we use just one format(...) for Forwarded header

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -234,11 +258,116 @@ private static void processRequestHeaders(HttpAcceptSession acceptSession, HttpC
...
+            String serverIpAddress = getResourceIpAddress(acceptSession, FORWARDED_BY);
+            if (serverIpAddress != null) {
+                connectSession.addWriteHeader(HEADER_X_FORWARDED_SERVER, serverIpAddress);
+                forwardedSb.append(format("%s=%s;", FORWARDED_BY, serverIpAddress));
+            }
+
+            URI externalURI = acceptSession.getLocalAddress().getExternalURI();
+            String protocol = externalURI.getScheme();
+            connectSession.addWriteHeader(HEADER_X_FORWARDED_PROTO, protocol);
+            forwardedSb.append(format("%s=%s;", FORWARDED_PROTO, protocol));
```
jitsni: Just because, scheme is https doesn't mean the protocol is https. The underlying stack may be configured and replaced differently (may be you should use acceptSession.isSecure() for http or https)

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
msalavastru: @jitsni I have fetched/merged the latest changes from upstream, but this class still remains static.

msalavastru:
@jitsni Thanks for your input. I made all the changes that you suggested, except the  one regarding the constructor for ConnectSessionInitializer for which i added a line comment above. I also updated all the k3po scripts that were expecting to receive a Forwarded header.

jitsni:
Looks OK to me. I need to check in IDE.

jitsni:
I have a concern about HttpSubjectSecurityFilter adding Forwarded header by default and http proxy is trying to deal with it. I talked with John and we thought we could just delete that code in HttpSubjectSecurityFilter that adds Forwarded header. The http proxy can add the Forwarded header according to the config setting. But I want to check with @sanjay-saxena if we are missing any usecases for geo security login module.

dpwspoon:
@robinzimmermann Needs to review the config changes

jitsni:
I think we should do the following:
- HttpSubjectSecurityFilter *doesn't* add Forwarded header at all
- HttpLoginSecurityFilter populates remote address to InetAddressCallbackHandler using Forwarded header (if there is one) else just session's remote address
- Clean up special code that accounts HttpSubjectSecurityFilter changes in this PR
- We should verify enterprise.gateway tests with these changes
- If the InetAddressFilterLoginModule is configured on a gateway after some gateway intermediaries, use-forwarded may need to be configured on those gateways appropriately.

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpLoginSecurityFilter.java
```diff
@@ -457,56 +458,69 @@ private void registerCallbacks(
         callbackHandlerMap.put(AuthenticationTokenCallback.class, authenticationTokenCallbackHandler);
 
         String forwarded = httpRequest.getHeader(HEADER_FORWARDED);
```
msalavastru: @jitsni Here getHeader will return the first element in the List of Forwarded headers added by intermediaries. Here Forwarded headers are not updated with the current info yet (did not go through proxy handler yet). I am not sure the data in the Forwarded headers is ok to be used, what do you think? Maybe we should just use the session's remote address?

jitsni: Two cases:
1. client --> gateway
In this case, there is no Forwarded header, we just use session's remote address. In fact, the request wouldn't even go through http.proxy service.

* client --> gateway1 ---> gateway2
In this case, there is Forwarded header from gateway1. The filter at gateway2 should use Forwarded header. In this case, the request would go through http.proxy service of gateway1 which adds Forwarded header (if configured to inject)

Hence, I am suggesting to use Forwarded header (if there is one), otherwise just use session's remote address(the best guess of client's address)

msalavastru:
@jitsni I have implemented the changes, and tested the InetAddressFilterLoginModule in EE with this changes and I think the behavior is correct:

- having a chain such as client -> gateway1 ->gateaway2, and a Login Module set for gateway2,  if both gateways add forwarded headers, then the LM will have the remote address set to the IP of the client from the forwarded headers. If, for example the gateway1 will not add forwarded headers, then the LM will have the remote address set to the IP of the gateway1, and not the client 
- if no forwarded headers are found the InetAddressFilterLoginModule will have the remote address set based on the ResourceAddress of the session.

I will also run the existing tests in EE to see if there is any impact.

jitsni:
Looks ok to me. @sanjay-saxena any comments about changes in HttpSubjectSecurityFilter, HttpLoginSecurityFilter

When it is ready, let me know.

msalavastru:
@jitsni I ran the tests in EE and they are not impacted by these changes.

jitsni:
Can you resolve the merge conflicts.

msalavastru:
@jitsni I have resolved the merging conflicts.

jitsni:
I think loop detection PR has caused merge conflicts. Can you resolve them again ?

msalavastru:
@jitsni I resolved the conflicts again.

PR #419, opened 2016-02-10
Http.proxy identical <accept> and <connect> with no additional accept/connect options provokes Gateway timeout

irina-mitrea-luxoft:
Added a first version of a fix - TBD: discuss as to whether the proposed approach in GatewayContextResolver is sufficient.

### DIFF
--- a/server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultAcceptOptionsContext.java
+++ b/server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultAcceptOptionsContext.java
@@ -169,6 +169,7 @@ public class DefaultAcceptOptionsContext implements AcceptOptionsContext {
         }
     }
 
+    @Override
     public Map<String, Object> asOptionsMap() {
         Map<String, Object> result = new LinkedHashMap<>();
 
--- a/server/src/main/java/org/kaazing/gateway/server/context/resolve/GatewayContextResolver.java
+++ b/server/src/main/java/org/kaazing/gateway/server/context/resolve/GatewayContextResolver.java
@@ -634,6 +634,13 @@ public class GatewayContextResolver {
             ConnectOptionsContext connectOptionsContext =
                     new DefaultConnectOptionsContext(connectOptions, defaultConnectOptions);
 
+            if (serviceType.equals("http.proxy") && acceptOptions == null && connectOptions == null) {
+                if (acceptURIs.contains(connectURIs.iterator().next())) {
+                    throw new RuntimeException(String.format("Different <accept> and <connect> URIs should"
+                            + " be provided for service of type %s", serviceType));
+                }
+            }
+
             Key encryptionKey = null;
 
             if (serviceRealmContext == null &&
--- a/server/src/main/java/org/kaazing/gateway/server/context/resolve/GatewayContextResolver.java
+++ b/server/src/main/java/org/kaazing/gateway/server/context/resolve/GatewayContextResolver.java
@@ -634,13 +634,6 @@ public class GatewayContextResolver {
             ConnectOptionsContext connectOptionsContext =
                     new DefaultConnectOptionsContext(connectOptions, defaultConnectOptions);
 
-            if (serviceType.equals("http.proxy") && acceptOptions == null && connectOptions == null) {
-                if (acceptURIs.contains(connectURIs.iterator().next())) {
-                    throw new RuntimeException(String.format("Different <accept> and <connect> URIs should"
-                            + " be provided for service %s of type %s", serviceName, serviceType));
-                }
-            }
-
             Key encryptionKey = null;
 
             if (serviceRealmContext == null &&
--- a/server/src/main/java/org/kaazing/gateway/server/context/resolve/GatewayContextResolver.java
+++ b/server/src/main/java/org/kaazing/gateway/server/context/resolve/GatewayContextResolver.java
@@ -637,7 +637,7 @@ public class GatewayContextResolver {
             if (serviceType.equals("http.proxy") && acceptOptions == null && connectOptions == null) {
                 if (acceptURIs.contains(connectURIs.iterator().next())) {
                     throw new RuntimeException(String.format("Different <accept> and <connect> URIs should"
-                            + " be provided for service of type %s", serviceType));
+                            + " be provided for service %s of type %s", serviceName, serviceType));
                 }
             }
 
--- a/service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyService.java
+++ b/service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyService.java
@@ -17,17 +17,22 @@ package org.kaazing.gateway.service.http.proxy;
 
 import static java.lang.String.format;
 
-import org.kaazing.gateway.service.ServiceContext;
-import org.kaazing.gateway.service.proxy.AbstractProxyService;
-
 import java.net.URI;
 import java.util.Collection;
-import java.util.Iterator;
+import java.util.Map;
+import java.util.Map.Entry;
+
+import org.kaazing.gateway.service.AcceptOptionsContext;
+import org.kaazing.gateway.service.ConnectOptionsContext;
+import org.kaazing.gateway.service.ServiceContext;
+import org.kaazing.gateway.service.proxy.AbstractProxyService;
 
 /**
  * Http proxy service
  */
 public class HttpProxyService extends AbstractProxyService<HttpProxyServiceHandler> {
+    private static final String IDENTICAL_ACCEPT_CONNECT_ERROR = "Different <accept> and <connect> URIs should"
+            + " be provided for service %s of type %s";
     private static final String TRAILING_SLASH_ERROR = "Accept URI is '%s' and connect URI is '%s'. Either both URI should end with / or not.";
 
     @Override
@@ -47,6 +52,14 @@ public class HttpProxyService extends AbstractProxyService<HttpProxyServiceHandl
 
         HttpProxyServiceHandler handler = getHandler();
         handler.setConnectURIs(connectURIs);
+
+        if (noOverridingAcceptConnectOptions(serviceContext)) {
+            if (serviceContext.getAccepts().contains(connectURIs.iterator().next())) {
+                throw new RuntimeException(String.format(IDENTICAL_ACCEPT_CONNECT_ERROR,
+                        serviceContext.getServiceName(), serviceContext.getServiceType()));
+            }
+        }
+
         handler.initServiceConnectManager();
     }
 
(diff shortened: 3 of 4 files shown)


irina-mitrea-luxoft:
To be discussed whether the current approach in the GatewayContextResolver is sufficient - a check is performed on whether accept/connect options are present - considering only options which would override the default accept/connect URIs would be a good approach.

server/src/main/java/org/kaazing/gateway/server/context/resolve/GatewayContextResolver.java
```diff
@@ -634,6 +634,13 @@ private DefaultServiceDefaultsContext resolveServiceDefaults(ServiceDefaultsType
             ConnectOptionsContext connectOptionsContext =
                     new DefaultConnectOptionsContext(connectOptions, defaultConnectOptions);
 
+            if (serviceType.equals("http.proxy") && acceptOptions == null && connectOptions == null) {
```
jitsni: Can we do this check in HttpProxyService#init(). Currently gateway server does many things which it shouldn't do, instead should done in specific components.

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultAcceptOptionsContext.java
```diff
@@ -169,6 +169,7 @@ public void addBind(String scheme, String hostPort) {
         }
     }
 
+    @Override
```
irina-mitrea-luxoft: Added also Override annotation with this PR.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyService.java
```diff
@@ -73,4 +86,37 @@ private void checkForTrailingSlashes(ServiceContext serviceContext) {
...
+
+        return bindsEmpty && !hasAcceptTransport && !hasConnectTransport;
+    }
+
+    /**
+     * Method returning whether option list has *.transport(s)
+     * @param map
+     * @return
+     */
+    private boolean hasTransport(Map<String, Object> map) {
```
irina-mitrea-luxoft: This method assumes knowledge of the fact that transport accept options are in *.transport format (which is actually an information to be encapsulated within DefaultAcceptOptionsContext/DefaultConnectOptionsContext).

Should we rather add a method like *hasTransports* in AcceptOptionsContex/ConnectOptionsContext interfaces?

irina-mitrea-luxoft:
@jitsni Hello, Jitu. I have updated the PR as per your comment and also added some comments to some lines of code within the PR.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyService.java
jitsni: It seems like a hack to check ".transport" (but I don't have a good solution)
- if there is some option like "socks.transport" would break this check
- there may be tcp.bind option
- there may be layers of transport options

The best thing is get ResourceAddress for accept and connect and check. But I don't see an easy way to do that here. It would be nice to do the check here while gateway is starting.

May be this can be done during request processing (which is bit late but better late than never). Need to think more.

irina-mitrea-luxoft: Regarding *.bind option:
- for AcceptOptionsContext: I am checking the fact that the getBinds() method returns an empty list - so this is covered
- for ConnectOptionsContext: DefaultConnectOptionsContext seems to not specifically address any *.bind option (DefaultConnectOptionsContext asOptionsMap() method does not return any *.bind). I do see that however the gateway-config-201409.xsd file does allow tcp.bind for connect options. Is there a reason why the *.xsd seems to allow the tcp.bind for connect options, but this option seems to not further be used in the code?

In case the tcp.bind option is not used for connects, the *.bind scenario should be covered by the validation of:
>
boolean bindsEmpty = acceptOptionsContext.getBinds().isEmpty();

irina-mitrea-luxoft: Hello, Jitu. Regarding *.bind option:
- for AcceptOptionsContext: a check is performed on the getBinds() method checking that it returns an empty list - so this should be covered
- for ConnectOptionsContext: DefaultConnectOptionsContext seems to not specifically address any *.bind option (DefaultConnectOptionsContext asOptionsMap() method does not return any *.bind). I do see that however the gateway-config-201409.xsd file does allow tcp.bind for connect options. Is there a reason why the *.xsd seems to allow the tcp.bind for connect options, but this option seems to not further be used in the code?

In case the tcp.bind option is not used for connects, the *.bind scenario should be covered by the validation of:
>
boolean bindsEmpty = acceptOptionsContext.getBinds().isEmpty();

dpwspoon:
I'm skeptical about the approach this fix is going.  For one it seems we are checking a bunch of things to make sure they are not there.  Meaning as new accept and connect options are added we will need to continually update this list.  Second, it seems like there may be use cases where the accept and connect may match (to preserve host headers etc), but the underlying transport would change so they get routed to the right place.   So it could be something that a user would hit.  3 even if we do catch this case it is not clear to me we will fulfill all the requirements of rfc 7230, by catching every case

>
 An intermediary MUST NOT forward a message to itself unless it is
   protected from an infinite request loop.  In general, an intermediary
   ought to recognize its own server names, including any aliases, local
   variations, or literal IP addresses, and respond to such requests
   directly.

Which may happen if someone sets the underlying transports to accept on connect on ip addresses that got routed in a loop that we weren't familiar with.  

Perhaps, there is a more generic catch all we can do to protect us from infinite loops, like look at the via header's/forwarded-for.  It would have the disadvantage of being at runtime and not startup.  But it might be a safer fix.

irina-mitrea-luxoft:
@dpwspoon Hello, David.

The ["Http.proxy identical <accept> and <connect> with no additional accept/connect options provokes Gateway timeout " issue](https://github.com/kaazing/tickets/issues/318), as initially entered, addressed only startup verification (verification which could be done with the data provided at starup, which would be without actual resolution or considering of actual values for *.bind(s) and *.tansport(s)).

We can change the title and description of the issue in order to address all scenarios (i.e. perform address resolution and impose restriction considering all bind/transports values). This will probably mean that the validation will not be able to be done at gateway startup.

irina-mitrea-luxoft:
@dpwspoon @jitsni  Hello David, Jitu. 
Thank you for your suggestion, David. Would it be ok to switch to an approach as per David's [comment](https://github.com/kaazing/gateway/pull/419#issuecomment-183978916):
>
Perhaps, there is a more generic catch all we can do to protect us from infinite loops, like look at the via header's/forwarded-for. It would have the disadvantage of being at runtime and not startup. But it might be a safer fix.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyService.java
```diff
@@ -47,6 +52,14 @@ public void init(ServiceContext serviceContext) throws Exception {
 
         HttpProxyServiceHandler handler = getHandler();
         handler.setConnectURIs(connectURIs);
+
+        if (noOverridingAcceptConnectOptions(serviceContext)) {
```
jitsni: This change is part of some other PR

jitsni:
Currently Via header doesn't have much info. Forwarded headers could be turned off for security purposes.

My thinking is that we should look at ResourceAddress for connect and accept during request processing and reject those requests. Other option is to some how use Max-Forwards header but it is usually used with TRACE/OPTIONS request methods.

irina-mitrea-luxoft:
@jitsni Hello, Jitu. Regarding the proposed approach:
>
My thinking is that we should look at ResourceAddress for connect and accept during request processing and reject those requests.

That sounds like a good approach, right now a timeout is received at request.

Generally speaking, regarding the run-time check: This would mean that we would only reject looping requests but provide no notification to the administrator regarding the fact that their configuration has led to this scenario and could potentially lead to this scenario again? We could probably log an error regarding this, but would that suffice?

Having even some minor startup validation could catch some of the potentially faulty configs.

irina-mitrea-luxoft:
@jitsni Hello, Jitu. Thank you for your feedback. Regarding the proposed approach:
>
My thinking is that we should look at ResourceAddress for connect and accept during request processing and reject those requests.

That sounds like a good approach, right now a timeout is received at request.

Regarding the second proposed approach, usage of the [Max-Forwards](https://tools.ietf.org/html/rfc2616#page-136) header, I see no code specifically in the gateway for it now, so we would have to add the support for it, am I correct?

This seems to be an approach in accordance to the [RFC](https://tools.ietf.org/html/rfc2616#page-136)  but, as you have mentioned above, seems to be related to TRACE and OPTIONS methods. It also would imply run-time validation.

Generally speaking, regarding the run-time check: This would mean that we would only reject looping requests but provide no notification to the administrator regarding the fact that their configuration has led to this scenario and could potentially lead to this scenario again? We could probably log an error regarding this, but would that suffice?

Having even some minor startup validation could catch some of the potentially faulty configs.

jitsni:
@jfallows do you have any comments on this ?

jitsni:
After discussing with John, we thought we could use Via header for detecting cycles (which should work when a service forwards to itself).
* Each *http.proxy service* (not gateway) injects a unique Via header
* A service checks for existing Via header values in a request and detects loop if the values contain the service's unique Via header. That request is closed with 508 status code.

A proxy should also handle Max-Forwards header (if exists) as per RFC.

irina-mitrea-luxoft:
@jitsni @sbadugu Hello, Jitu, thank you for your feedback. As your above comment we will be using the [Via](http://tools.ietf.org/html/rfc7230#section-5.7.1) header. Excerpt from the [RFC 7230](http://tools.ietf.org/html/rfc7230):
>
Via can be used for tracking message forwards, avoiding request loops, and identifying the protocol capabilities of senders along the request/response chain.

The scenario described in the RFC above would be a match for our current scenario.

I have looked over the implementation within HttpProxyServiceHandler and noticed that we already inject the **Via** header, by default hardcoded to the value **1.1 kaazing**. This value will probably be switched to the **protocol version** and **received-by** or **pseudonym**, as per the RFC description:
>
Via = 1#( received-protocol RWS received-by [ RWS comment ] )
     received-protocol = [ protocol-name "/" ] protocol-version
     received-by       = ( uri-host [ ":" port ] ) / pseudonym
     pseudonym         = token
(if the real host is considered to be sensitive information, a sender MAY replace it with a pseudonym)

Can you please confirm whether we should use a pseudonym?

irina-mitrea-luxoft:
@jitsni @sbadugu Also, as per your comment
>
A proxy should also handle Max-Forwards header (if exists) as per RFC.

above we should also introduce [Max-Forwards](http://tools.ietf.org/html/rfc2616#page-136) support within the gateway, am I correct? After a brief analysis I didn't find any implementation regarding this within the gateway, so as per my understanding we should address this within the current PR.

As per the RFC description, this will work for TRACE and OPTIONS methods:
>
The Max-Forwards request-header field provides a mechanism with the TRACE (section 9.8) and OPTIONS (section 9.2) methods to limit the number of proxies or gateways that can forward the request to the  next inbound server.

jitsni:
We will use  pseudonym, something like "1.1 kaazing-guid" for each http.proxy service.

Max-Forwards support is not there in http.proxy. It can be added (but not critical for first release. How often do we see TRACE and OPTIONS requests ?)

irina-mitrea-luxoft:
This PR will be closed and [PR](https://github.com/kaazing/gateway/pull/443) will be used as a reference from now on.

PR #431, opened 2016-02-22
Throw error on specific circumstances when PreferedIPv4 flag is true and the host IP is IPV6

vmaraloiu:


resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/ResourceAddressFactorySpi.java
```diff
@@ -463,5 +471,36 @@ protected void parseNamedOptions0(URI location, ResourceOptions options,
...
+        throw new IllegalArgumentException(error.toString());
+    }
+
+    /**
+     * Verify PreferedIPv4 flag and IPV6 in order to throw a complex error message
+     * @param error
+     * @param address
+     */
+    private String getErrorMessageIfPreferedIPv4(InetAddress address) {
+        boolean isPeferedIPv4 = "true".equals(System.getProperty(JAVA_NET_PREFER_IPV4_STACK));
```
dpwspoon: typo

resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/ResourceAddressFactorySpi.java
```diff
@@ -463,5 +471,36 @@ protected void parseNamedOptions0(URI location, ResourceOptions options,
...
+     * Throw error on specific circumstances:
+     * - no addresses available for binding
+     * - when PreferedIPv4 flag is true and the host IP is IPV6
+     * @param location
+     */
+    private void throwNoAddressesToBindError(URI location) {
+        StringBuffer error = new StringBuffer(format(NO_ADDRESSES_AVAILABLE_FOR_BINDING, location));
+        try {
+            InetAddress address = InetAddress.getByName(location.getHost());
+            error.insert(0, getErrorMessageIfPreferedIPv4(address));
```
dpwspoon: personally I think it may be clearer to read to pull the method call getErrorMessageIfPreferedIPv4 into this block and use normal if block.  The return of `""` making this call a NOOP is unexpected.  Perhaps something like this would be better

```java
if("true".equals(System.getProperty(JAVA_NET_PREFER_IPV4_STACK)) && (address instanceof Inet6Address)){
    error.insert(0, PREFER_IPV4_STACK_IPV6_ADDRESS_EXCEPTION)
}
```

resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/ResourceAddressFactorySpi.java
```diff
@@ -42,7 +45,11 @@
 
 public abstract class ResourceAddressFactorySpi<T extends ResourceAddress> {
 
+    private static final String PREFER_IPV4_STACK_IPV6_ADDRESS_EXCEPTION = "Option java.net.preferIPv4Stack is set to"
+                                                               + " true and an IPv6 address was provided in the config.";
+    private static final String NO_ADDRESSES_AVAILABLE_FOR_BINDING = "No addresses available for binding for URI: %s.";
```
dpwspoon: I personally don't think constants for strings that are used once are necessary and I would normally inline these.  That is completely personal preference and some people would disagree with me.  But if you choose to keep these as constants, anything with a %s or other formatter needs to have _FORMATTER or something else in the name saying it is a formatter and not a string

resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/ResourceAddressFactorySpi.java
```diff
@@ -463,5 +471,36 @@ protected void parseNamedOptions0(URI location, ResourceOptions options,
...
+            error.insert(0, getErrorMessageIfPreferedIPv4(address));
+        } catch (UnknownHostException e) {
+            // InetAddress.getByName(hostAddress) throws an exception (hostAddress may have an
+            // unsupported format, e.g. network interface syntax)
+        }
+        throw new IllegalArgumentException(error.toString());
+    }
+
+    /**
+     * Verify PreferedIPv4 flag and IPV6 in order to throw a complex error message
```
dpwspoon: I normally would not comment on this but as there is already things to fix, the first line of JavaDoc should end with punctuation.  Our disabled checkstyle that we want to enable in the future would throw an error on this.

dpwspoon:
Thanks @vmaraloiu, let me know when you have applied the feedback

vmaraloiu:
Thank you for feedback @dpwspoon, I have added the modifications.

resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/ResourceAddressFactorySpi.java
```diff
@@ -42,7 +45,11 @@
 
 public abstract class ResourceAddressFactorySpi<T extends ResourceAddress> {
 
+    private static final String PREFER_IPV4_STACK_IPV6_ADDRESS_EXCEPTION_FORMATTER = "Option java.net.preferIPv4Stack is set to"
```
dpwspoon: PREFER_IPV4_STACK_IPV6_ADDRESS_EXCEPTION_FORMATTER is not a formatter correct, just a String.  NO_ADDRESSES_AVAILABLE_FOR_BINDING_FORMATTER is a formatter as it has a &s

dpwspoon:
One minor nit,

vmaraloiu:
Hi @dpwspoon, I have added last sugestion from feedback. Also there are some more feedback items added by @jitsni in the [PR](https://github.com/kaazing-private/gateway.server/pull/124) from 4.x. What do you sugest regarding this?

dpwspoon:
This looks good to me.  If there are changes @jitsni suggested I would suggest pulling them into this one as well.  I'll assign this to Jitu so he can give direct feedback

vmaraloiu:
Ported changes from 4.x [PR](https://github.com/kaazing-private/gateway.server/pull/124).

vmaraloiu:
Ported review changes from 4.x [PR](https://github.com/kaazing-private/gateway.server/pull/130).

PR #432, opened 2016-02-25
Adding code to rewrite cookie domain, cookie path, location headers

jitsni:


server/src/main/xsd/gateway-config-201409.xsd
```diff
@@ -1516,6 +1516,114 @@
         </complexContent>
     </complexType>
 
+    <complexType name="httpProxyServicePropertiesType">
+        <complexContent>
+            <restriction base="gateway:ServicePropertiesType">
+                <sequence>
```
jitsni: I tried using <all> to specify these elements in any order, but it doesn't allow multiple elements like <location-mapping>. So I am using <sequence>. We can live with it or there are two options:
* Remove the changes in schema and use free form for specifying properties (jms follows that)
* Or need to create intermediate elements like <location-mappings> which contains <location-mapping>. 

@jfallows @robinzimmermann Please review the schema change

dpwspoon: I'm ok with free form and would prefer that to having a strict sequence given the low quality exception that is thrown if you get the order wrong.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyService.java
```diff
@@ -18,11 +18,15 @@
 import static java.lang.String.format;
 
 import org.kaazing.gateway.service.ServiceContext;
+import org.kaazing.gateway.service.ServiceProperties;
 import org.kaazing.gateway.service.proxy.AbstractProxyService;
 
 import java.net.URI;
 import java.util.Collection;
+import java.util.HashMap;
```
dpwspoon: nit, hmm imports change without code that reflects that.  Maybe they are unused now

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -36,35 +37,82 @@
...
 import static org.kaazing.gateway.transport.http.HttpHeaders.HEADER_UPGRADE;
 import static org.kaazing.gateway.transport.http.HttpHeaders.HEADER_VIA;
-import static org.kaazing.gateway.transport.http.HttpStatus.INFO_SWITCHING_PROTOCOLS;
 import static org.kaazing.gateway.transport.http.HttpStatus.CLIENT_NOT_FOUND;
+import static org.kaazing.gateway.transport.http.HttpStatus.INFO_SWITCHING_PROTOCOLS;
 
 class HttpProxyServiceHandler extends AbstractProxyAcceptHandler {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger("http.proxy");
+    private static final Logger LOGGER = LoggerFactory.getLogger("service.http.proxy");
```
dpwspoon: This looks good.  Is this the convention we are following in general for logging names on services?

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -178,14 +226,92 @@ private void processResponseHeaders(IoSession session) {
...
+
+        private String processCookieDomain(String cookie, Map<String, String> cookieDomainMap) {
+            String lowerCookie = cookie.toLowerCase();
+            if (lowerCookie.contains("domain=")) {
+                return cookieDomainMap.entrySet().stream()
+                        .filter(e -> lowerCookie.contains("domain="+e.getKey()))
+                        .findFirst()
+                        .map(e -> {
+                            int index = lowerCookie.indexOf("domain="+e.getKey());
+                            return cookie.substring(0, index+7)+e.getValue()+cookie.substring(index+7+e.getKey().length());
```
dpwspoon: The 7 letters is from `http://`? Could the scheme be `https://` or something different then 7 letters long?

dpwspoon: nevermind I see.  its domain

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
jitsni: broadcast, directory are doing it. Some services (for e.g echo) are not doing.

server/src/main/xsd/gateway-config-201409.xsd
dpwspoon: Follow up from Robin, allow any ordering.  Any validation we can do on top of that is bonus

dpwspoon:
@jitsni  Let me know when you've updated

jitsni:
Updated PR (primary change is no schema types for service properties)

PR #436, opened 2016-03-07
Fix line endings on DuplicateBindTest.java

Anisotrop:
File was committed with LF line endings which makes core.autocrlf change it and further it appears as modified after repository cloning. Committed now with CRLF.


jitsni:
I don't think it is problem (as most of us create files with LF endings). But merging it

PR #440, opened 2016-03-08
Added gatewayContext to startingGateway and stoppingGateway methods from GatewayObserverApi

mgherghe:


dpwspoon:
I tested this locally

PR #442, opened 2016-03-09
Fix ipv6 accept options

Anisotrop:
The previous method that parsed the host and port from a String was failing for IPv6. Added some unit tests.


jitsni:
can you resolve the conflicts ?

PR #443, opened 2016-03-09
Connection closed when http.proxy loops encountered

irina-mitrea-luxoft:
For background on this pull request, [PR](https://github.com/kaazing/gateway/pull/419) can be used as reference.

All Via header *.rpt scripts would need to be changed if this approach is to be used, since Via headers now hold authority information within them. We should also decide whether we should use a hash based on the authority value instead of directly using the authority value.


irina-mitrea-luxoft:
@jitsni Hello, Jitu, I have created a PR with the first set of changes on loop checks using Via headers. I have also added ITests on this. All existing K3po *.rpts containing the Via header should also be updated. I will do this change as soon as I have feedback on the current PR so that we know this is the desired approach. Thank you, Irina.

service/http.proxy/src/test/java/org/kaazing/gateway/service/http/proxy/HttpProxyChainedProxiesInLoopIT.java
```diff
@@ -0,0 +1,73 @@
...
+                    .done();
+            // @formatter:on
+            init(configuration);
+        }
+    };
+
+    @Rule
+    public TestRule chain = createRuleChain(gateway, robot);
+
+    @Specification({"http.proxy.linked.proxy.request.loop"})
```
jitsni: { } is not required if you specify single value in annotation

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -152,6 +154,25 @@ public void operationComplete(ConnectFuture future) {
             }
         }
 
+        /**
+         * Helper method performing loop detection
+         * @param connectSession
```
jitsni: Either add javadoc for parameters or remove the @param (My IDE highlights it  and doesn't give the file a green check !)

service/http.proxy/src/test/scripts/org/kaazing/gateway/service/http/proxy/http.proxy.linked.proxy.reply.rpt
```diff
@@ -0,0 +1,48 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#
+# Origin server
+#
+
+accept tcp://localhost:8030
+accepted
```
jitsni: I think you should create http tests(instead of tcp, i.e. accept http://localhost:8030) which makes it much easier.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -152,6 +154,25 @@ public void operationComplete(ConnectFuture future) {
...
+         * @param viaHeaders
+         */
+        private void performLoopDetection(DefaultHttpSession connectSession, List<String> viaHeaders) {
```
jitsni: I think this can be done on the accept session itself (just after validateRequestPath) that way no need to connect to the next server. You could also simplify the logic
```
if (viaHeaders != null && viaHeaders.stream().anyMatch(h -> h.equals(viaHeader))) {
...
```

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -232,8 +253,9 @@ private static void processRequestHeaders(HttpAcceptSession acceptSession, HttpC
             }
         }
 
-        // Add Via: 1.1 kaazing header
-        connectSession.addWriteHeader(HEADER_VIA, VIA_HEADER_VALUE);
+        // Add Via: 1.1 kaazing + authority header
+        String authority = acceptSession.getLocalAddress().getExternalURI().getAuthority();
+        connectSession.addWriteHeader(HEADER_VIA, VIA_HEADER_VALUE + SEPARATOR + authority);
```
jitsni: I would suggest create via header value using some guid in the constructor of HttpProxyServiceHandler.

```
    private final String viaHeader;

    HttpProxyServiceHandler() {
        viaHeader = String.format("1.1 kaazing-%s", UUID.randomUUID());
    }
```

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpStatus.java
```diff
@@ -67,7 +67,8 @@
     SERVER_BAD_GATEWAY(502, "Bad Gateway"),
     SERVER_SERVICE_UNAVAILABLE(503, "Service Unavailable"),
     SERVER_GATEWAY_TIMEOUT(504, "Gateway Timeout"),
-    SERVER_VERSION_NOT_SUPPORTED(505, "Version Not Supported");
+    SERVER_VERSION_NOT_SUPPORTED(505, "Version Not Supported"),
+    SERVER_LOOP_DETECTED(508, "Loop detected");
```
jitsni: Seems like 508 is only for webdav.

service/http.proxy/src/test/scripts/org/kaazing/gateway/service/http/proxy/http.proxy.linked.proxy.reply.rpt
irina-mitrea-luxoft: @jitsni Hellom, Jitu. I had initially opted for tcp since http k3po transport does not do an exact count match on headers (such as Via header). However, I have now done this change as per your suggestion as it does simplify the scripts.

irina-mitrea-luxoft: @jitsni Hello, Jitu. I had initially opted for tcp since http k3po transport does not do an exact count match on headers (such as Via header). However, I have now done this change as per your suggestion as it does simplify the scripts.

irina-mitrea-luxoft:
@jitsni Hello, Jitu, thank you for your feedback, I have checked the changes as per the code review. Please give me your feedback on this. When we are ok with the current approach, I can start also modifying exisiting k3po  tests which may use directly that "1.1 kaazing" Via header. Thank you, Irina.

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpStatus.java
irina-mitrea-luxoft: @jitsni Hello, Jitu. Would you suggest using a different error code and not [508](https://tools.ietf.org/html/rfc5842#section-7.2)?

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -92,15 +97,32 @@ private boolean validateRequestPath(DefaultHttpSession acceptSession) {
         return requestPath.startsWith(acceptPath);
     }
 
+
+    /**
+     * Helper method performing loop detection
+     * @param acceptSession - session parameter
+     */
+    private void performLoopDetection(DefaultHttpSession acceptSession) {
```
jitsni: May be it needs to return boolean so that no connect is performed when loop is detected.

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -92,15 +97,32 @@ private boolean validateRequestPath(DefaultHttpSession acceptSession) {
...
     /*
      * Initializer for connect session. It adds the processed accept session headers
      * on the connect session
      */
     private static class ConnectSessionInitializer implements IoSessionInitializer<ConnectFuture> {
         private final DefaultHttpSession acceptSession;
+        private final String viaHeader;
 
-        ConnectSessionInitializer(DefaultHttpSession acceptSession) {
+        ConnectSessionInitializer(DefaultHttpSession acceptSession, String viaHeader) {
```
jitsni: It seems like ConnectSessionInitializer non-static class (that way you don't have to pass viaHeader parameter. similarly useForwarded parameter etc)

irina-mitrea-luxoft:
Added changes as per code review; still pending: solving conflicts with upstream, modifying "1.1 kaazing" Via header already existing tests if needed - was unable to test last changes to PR locally due to an environment issue.

irina-mitrea-luxoft:
@jitsni Hello, Jitu, added the changes with the items which were pending:
- Via headers in *.rpt scripts
- Merged changes from upstream and fixed conflicts

Please provide your feedback on the current PR, all remaining issues should be solved now, there are still some tests which are failing for me but they should be related to the fact that the build fails on Windows 7 for some tests and not specifically related to my changes.
Thank you, Irina.

irina-mitrea-luxoft:
@jitsni Hello, Jitu, @msalavastru 's changes on [PR](https://github.com/kaazing/gateway/pull/415) are impacted by the current PR as follows:
- Via headers now have "1.1 kaazing-uuid" format
- ConnectSessionInitializer is not static anymore so it can access it's outer class instance attributes
The information above is relevant for deciding the merge order of the pull requests. Thank you, Irina.

PR #445, opened 2016-03-09
Gateway returns incorrect error code when payload is too long (1002 instead of 1009)

Anisotrop:
Implement a new close message type triggerd by a different exception. Set the response code to value 1009 (0xf1) when closing the session, if the message length exceeds the maximum length configured in accept-options/ws.maximum.message.size. Fix a trace message to include the actual close frame sent.

### DIFF
--- a/transport/ws/src/main/java/org/kaazing/gateway/transport/ws/WsCloseMessage.java
+++ b/transport/ws/src/main/java/org/kaazing/gateway/transport/ws/WsCloseMessage.java
@@ -26,6 +26,7 @@ import org.kaazing.gateway.util.Utils;
 public class WsCloseMessage extends WsMessage {
     public static final WsCloseMessage NORMAL_CLOSE = new WsCloseMessage(1000, null);
     public static final WsCloseMessage PROTOCOL_ERROR = new WsCloseMessage(1002, null);
+    public static final WsCloseMessage MESSAGE_TOO_LONG_ERROR = new WsCloseMessage(1009, null);
     public static final WsCloseMessage UNEXPECTED_CONDITION = new WsCloseMessage(1011, null);
 
     private final int status;
--- a/transport/ws/src/main/java/org/kaazing/gateway/transport/ws/bridge/filter/WsFrameDecoder.java
+++ b/transport/ws/src/main/java/org/kaazing/gateway/transport/ws/bridge/filter/WsFrameDecoder.java
@@ -32,6 +32,7 @@ import org.kaazing.gateway.transport.ws.WsPingMessage;
 import org.kaazing.gateway.transport.ws.WsPongMessage;
 import org.kaazing.gateway.transport.ws.WsTextMessage;
 import org.kaazing.gateway.transport.ws.bridge.filter.WsFrameEncodingSupport.Opcode;
+import org.kaazing.gateway.transport.ws.util.WSMessageTooLongException;
 import org.kaazing.mina.core.buffer.IoBufferAllocatorEx;
 import org.kaazing.mina.core.buffer.IoBufferEx;
 import org.kaazing.mina.filter.codec.CumulativeProtocolDecoderEx;
@@ -255,10 +256,10 @@ public class WsFrameDecoder extends CumulativeProtocolDecoderEx {
         buf.position(start);
     }
 
-    private void validateMessageSize(long messageSize) throws ProtocolDecoderException {
+    private void validateMessageSize(long messageSize) throws WSMessageTooLongException {
         // note: negative size indicates large unsigned value, larger than Long.MAX_VALUE
         if (maxMessageSize > 0 && (messageSize < 0 || messageSize > maxMessageSize)) {
-            throw new ProtocolDecoderException(String.format("Incoming message size %d bytes exceeds permitted maximum of %d bytes", messageSize, maxMessageSize));
+            throw new WSMessageTooLongException(String.format("Incoming message size %d bytes exceeds permitted maximum of %d bytes", messageSize, maxMessageSize));
         }
     }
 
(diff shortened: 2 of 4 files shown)


Anisotrop:
Some ITs fail. Investigating wheter it is due to the changes in this PR or not.

Anisotrop:
Successfully ran integration tests on the development machine by increasing timeouts of FramingTextMessagesIT.

jitsni:
Looks ok to me. Is there a test that uses ws.maximum.message.size config option along with this PR ?

cmebarrow:
@jitsni The tests are being added by @NicoletaOita (see https://github.com/kaazing/QA-Roadmap/issues/60)

Anisotrop:
@jitsni, @cmebarrow  I have merged the changes of Nicoleta on a local branch and have run the tests.
LimitsMaxSize126IT has passed.
LimitsSessionTimeoutIT has passed.

LimitsMaxSizeDefaultIT has failed with stacktrace:
```java
org.apache.mina.filter.codec.ProtocolDecoderException: RSV1 is set (Hexdump: 90 63 [...])
	at org.kaazing.gateway.transport.ws.bridge.filter.WsFrameDecoder.validateRSV(WsFrameDecoder.java:201)
        at org.kaazing.gateway.transport.ws.bridge.filter.WsFrameDecoder.doDecode(WsFrameDecoder.java:61)
	at org.kaazing.mina.filter.codec.CumulativeProtocolDecoderEx.decode(CumulativeProtocolDecoderEx.java:209)
[...]
```

The validation of the message doesn't seem to raise an exception. When analyzing the code it appeared that the validation of maximum size is currently done at frame level by incrementing the message size with the size of each frame for each CONTINUATION opcode. In this case the each frame has opcode BINARY, frameSize 65536 and the maximumMessageSize is 131072.

jitsni:
The current logic to calculate the candidate size in WsFrameDecoder looks ok to me.

I think that you want to check the scripts.

cmebarrow:
@jitsni @Anisotrop Yes, the script is wrong, it's my mistake. It's setting length to 65536 instead of 131073. @NicoletaOita could you please fix these scripts (the 4 should.fail.xxx.payload.length.131073 ones) as part of your PR https://github.com/k3po/k3po/pull/316? In the request scripts the line 
```
write [0x82 0xff 0x00 0x00 0x00 0x00 0x00 0x01 0x00 0x00] ${writeMask}
```
needs to be replaced with
```
write [0x82 0xff 0x00 0x00 0x00 0x00 0x00 0x02 0x00 0x01] ${writeMask}
```
(because 131073 in hexadecimal is 0x02 0x00 0x01)
and of course a similar change needs to be made in the response scripts

NicoletaOita:
@cmebarrow @Anisotrop I've modified the scripts with the correct hexadecimal value in my [k3po PR] (https://github.com/k3po/k3po/pull/316).

Anisotrop:
@cmebarrow I have used the latest version of the script and the test passes but had to replace in my setup because of a parse exception. I replaced:
```java
property location 'http://localhost:8080/echo'
``` 
with 
```java
property location "http://localhost:8080/echo"
```
.
Also the test will trigger an exception in the method validateMessageSize and not validateFrameSize:

```java
    private void validateMessageSize(long messageSize) throws WSMessageTooLongException {
        // note: negative size indicates large unsigned value, larger than Long.MAX_VALUE
        if (maxMessageSize > 0 && (messageSize < 0 || messageSize > maxMessageSize)) {
            throw new WSMessageTooLongException(String.format(
                    "Incoming message size %d bytes exceeds permitted maximum of %d bytes", messageSize, maxMessageSize));
        }
    }

    private void validateFrameSize(long frameSize, Opcode opcode) throws ProtocolDecoderException {
        // Control frame payload size cannot be greater than 125 bytes
        if (frameSize > 125) {
            switch (opcode) {
            case PING:
            case PONG:
            case CLOSE:
                throw new ProtocolDecoderException("Invalid " + opcode + " frame payload length = " + frameSize);
            default:
                break;
            }
        }

        // If there is data in CLOSE frame, first 2-bytes represent close status code
        if (opcode == Opcode.CLOSE && frameSize == 1) {
            throw new ProtocolDecoderException("Invalid CLOSE frame payload length = " + frameSize);
        }
}
```

Some "garbage" also seems to be read after the message payload size validation fails. There seems to be a buffer with capacity: 2048 position: 0 remaining: 2048 that is decoded and I haven't yet determined if it is a loop or a normal decoding process. Nevertheless the opcodes that are read seem to have random values.

cmebarrow:
@Anisotrop Costin, Jitu and I think these changes are fine so Jitu has merged. About your observation about the "garbage", please explain how you saw this (so we can try it ourselves too if needed). Do you think it's an issue we need to resolve? Is there a functional issue?

Anisotrop:
@cmebarrow I have created a branch on my repository that should allow to reproduce the observed behavior. The branch is [investigate_ws_decode_frame](https://github.com/Anisotrop/gateway/tree/investigate_ws_decode_frame) and it merges into the latest develop @NicoletaOita's branch related to this issue and also modifies WSFrameDecoder.java to write the execution steps into an external file. When executing the test LimitsMaxSizeDefaultIT I have obtained output similar to the attached one.
[fix_370.txt](https://github.com/kaazing/gateway/files/179459/fix_370.txt)

cmebarrow:
@Anisotrop Costin, you don't need to add all those diagnostics to WsFrameDecoder.java, all you have to do is thrown an exception from the test method after k3po.finish():
```
    public void shouldRefuseBinaryFrameWithPayloadLengthExceeding128KiB() throws Exception {
        k3po.finish();
        throw new Exception("TEMP");
    }
```
This causes trace level of messages to be printed out (thanks to the MethodExecutionTrace rule).
I did that (using Nicoleta's PR which I have now merged) and saw the exceptions you are talking about in the log messages. I am preparing a fix.

PR #447, opened 2016-03-10
Added test for web socket compliance (maximum message size and maxmimum lifetime).

NicoletaOita:
Added test for web socket compliance (maximum message size and maxmimum lifetime).


transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/specification/ws/acceptor/LimitsIT.java
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    @Test
+    @Specification({
+        "should.fail.max.lifetime.exceeded/handshake.request.and.frame"
+        })
+    public void shouldFailSendingMessageAfterSessionLifetimeIsUp() throws Exception {
+        k3po.start();
+        k3po.awaitBarrier("HANDSHAKE_COMPLETE");
+        Thread.sleep(3000);
```
cmebarrow: You should not need the sleep here since `k3po.awaitBarrier("CLOSE_RECEIVED")` will block until the barrier is notified by the request script, and that will only happen when the WebSocket session times out. You should add code to verify that the timing is about correct (stash System.currentTimeMillis instead of your Thread.sleep call and call it against after k3po.awaitBarrier("CLOSE_RECEIVED") and check the difference is at least say half of TEST_LIFETIME (converted to milliseconds). Don't make the test too straight because timing vary a lot between platforms.

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/specification/ws/acceptor/LimitsIT.java
```diff
@@ -0,0 +1,162 @@
+/**
```
cmebarrow: Given my comment on the k3po PR asking you to make the test cases on use http://localhost:8080/echo, you will need to split this class into multiple ones, one per required different gateway configuration. I would recommend splitting into classes with names like "LimitsMaxSize126IT",  "LimitsMaxSizeDefaultIT", and "LimitsSessionTimeoutIT"

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/specification/ws/acceptor/LimitsIT.java
```diff
@@ -0,0 +1,162 @@
...
+
+    private GatewayRule gateway = new GatewayRule() {
+        {
+            // @formatter:off
+            GatewayConfiguration configuration =
+                    new GatewayConfigurationBuilder()
+                        .service()
+                            .accept(URI.create(WS_PROXY_SERVICE_ACCEPT_LIMIT))
+                            .connect(URI.create(PROXY_SERVICE_CONNECT_LIMIT))
+                            .type("proxy")
```
cmebarrow: We don't need to test against both proxy and echo services since these tests are purely testing the WebSocket transport. They are not testing services. Just use echo service for simplicity.

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/specification/ws/acceptor/LimitsIT.java
```diff
@@ -0,0 +1,162 @@
...
+    private static String PROXY_SERVICE_CONNECT_LIMIT = "tcp://localhost:" + PROXY_SERVICE_CONNECT_PORT_LIMIT;
+
+    private static String WS_ECHO_SERVICE_ACCEPT_DEFAULT = "ws://localhost:8081/echo";
+    private static String WS_PROXY_SERVICE_ACCEPT_DEFAULT = "ws://localhost:8011/proxy";
+    private static int PROXY_SERVICE_CONNECT_PORT_DEFAULT = 8124;
+    private static String PROXY_SERVICE_CONNECT_DEFAULT = "tcp://localhost:" + PROXY_SERVICE_CONNECT_PORT_DEFAULT;
+    
+    
+    private static int DEFAULT_WEBSOCKET_MAXIMUM_MESSAGE_SIZE = 128*1024;
+    private static int LIMITED_WEBSOCKET_MAXIMUM_MESSAGE_SIZE_125 = 125;
```
cmebarrow: For readability I would recommend removing this constant and just putting "125" where you are using it. We don't need to slavishly define constant variable for every constant value we use. But I certainly agree it's testified in the case of the default. This is a minor point however so feel free to disagree and leave it.

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/specification/ws/acceptor/LimitsIT.java
```diff
@@ -0,0 +1,162 @@
...
+                                .allowOrigin("*")
+                            .done()
+                            .acceptOption("ws.maximum.message.size", Integer.toString(LIMITED_WEBSOCKET_MAXIMUM_MESSAGE_SIZE_125))
+                        .done()
+
+                        .service()
+                            .accept(URI.create(WS_PROXY_SERVICE_ACCEPT_DEFAULT))
+                            .connect(URI.create(PROXY_SERVICE_CONNECT_DEFAULT))
+                            .type("proxy")
+                            .acceptOption("ws.maximum.message.size", Integer.toString(DEFAULT_WEBSOCKET_MAXIMUM_MESSAGE_SIZE))
```
cmebarrow: The purpose of the default tests is to verify the gateway does the check by default ( without being told), so you should remove this line (`acceptOption("ws.maximum.message.size",...`)

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/specification/ws/acceptor/LimitsSessionTimeoutIT.java
```diff
@@ -0,0 +1,92 @@
...
+    };
+
+    @Rule
+    public TestRule chain = createRuleChain(gateway, k3po);
+
+    @Test
+    @Specification({
+        "should.fail.max.lifetime.exceeded/handshake.request.and.frame"
+        })
+    public void shouldFailSendingMessageAfterSessionLifetimeIsUp() throws Exception {
```
cmebarrow: A more accurate name would be shouldCloseWebSocketConnectionAfterSessionLifetimeIsUp

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/specification/ws/acceptor/LimitsSessionTimeoutIT.java
```diff
@@ -0,0 +1,92 @@
...
+        k3po.start();
+        k3po.awaitBarrier("HANDSHAKE_COMPLETE");
+
+        long startTime = System.currentTimeMillis();
+        k3po.awaitBarrier("CLOSE_RECEIVED");
+        long endTime = System.currentTimeMillis();
+
+        long timePassed = endTime - startTime;
+        //make sure the connection does not close immediately
+        assertTrue(timePassed >= TimeUnit.SECONDS.toMillis(TEST_LIFETIME)/2);
```
cmebarrow: Yes, this looks good

cmebarrow:
@NicoletaOita Done reviewing. Just one minor comment about a method name. Once you fix that and once the k3po PR is merged and release I will merge this. Are all the tests passing with @Anisotrop's fix for https://github.com/kaazing/gateway/issues/445?

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/specification/ws/acceptor/LimitsSessionTimeoutIT.java
NicoletaOita: Yes, the connection closes after ~3.1 seconds.

NicoletaOita:
@cmebarrow Done with the last changes, also fixed the scripts for default size, as suggested.

cmebarrow:
@NicoletaOita This looks good. I have merged your k3po PR (https://github.com/k3po/k3po/pull/316). Once that has built I will release k3po, then you can update the pom file to pick up that version and I wil be able to merge this PR.

cmebarrow:
@NicoletaOita I have scheduled k3po release build 3.0.0-alpha-24 on [Jenkins](http://jenkins.kaazing.wan/view/release_build/job/k3po-release/). Once that is done, you will need to put that as the k3po version in the gateway pom.xml file. Then you should be held to build your  branch and it should pick up the new scripts (and scripts updates) that you did.

cmebarrow:
@NicoletaOita k3po 3.0.0-alpha-24 release has completed so you can go ahead and set that as the k3po version in properties near the start of the gateway pom.xml. Then I will be able to merge this PR.

NicoletaOita:
@cmebarrow Thank you! I've modified the k3po version in the gateway pom file.

NicoletaOita:
@cmebarrow Thanks for merging the k3po PR. I've updated the k3po version in the gateway pom file and rebuilt my branch, so I think now you should be able to merge this PR too.

cmebarrow:
@NicoletaOita  The Travis build failed in your new size exceeded tests because Costin's fix (https://github.com/kaazing/gateway/pull/445) was not yet in. Jitu has merged it now, so I have restarted the Travis build to see if it passes now.

PR #448, opened 2016-03-10
Fixed pipe error message.

vmaraloiu:


jitsni:
Can you also resolve conflicts.

resource.address/pipe/src/main/java/org/kaazing/gateway/resource/address/pipe/PipeResourceAddressFactorySpi.java
```diff
@@ -28,7 +28,7 @@
 
     private static final String PROTOCOL_NAME = "pipe";
 
-    private static final String PIPE_PATH_ERROR_MESSAGE = "Using pipe://%s instead of pipe://%s%s "
+    private static final String PIPE_PATH_ERROR_MESSAGE = "Use pipe://%s instead of pipe://%s%s "
             + "because paths are ignored for pipe:// URIs. See 'pipe://' in the documentation for more details.";
```
jitsni: I think you should just say named pipes shouldn't contain path. Something like
private static final String PIPE_PATH_ERROR_MESSAGE = "Use pipe://%s instead of pipe://%s%s because named pipe URIs shouldn't contain paths."

I think "see pipe:// in the documentation for more details" is superfluous.

vmaraloiu:
Implemented the code review changes.

PR #451, opened 2016-03-14
Forward feature #560 from 4.x to 5.x: Implement the directives for Cache-Control header in Directory Service

mgherghe:


jitsni:
This PR seems to break the build. See
https://github.com/kaazing/gateway/issues/468

PR #455, opened 2016-03-17
Anisotrop fix ipv6 accept options

Anisotrop:
Resolved conflicts with develop. The regular expression seemed to miss an IPv6 adress between brackets (because of the mandatory @ sign). Proposing a simplification to try to extract the host and port by generic expression. If the expression doesn't match the value could be a port. Otherwise is host only and it should be rejected. Some of the tests don't cover the changed method since they fail earlier.
* shouldRequireExplicitPortOnIPv6FromString
* shouldRequireExplicitPortOnIPv6

Exception is thrown at 	
```java
org.kaazing.gateway.resource.address.URLUtils.modifyURIAuthority(URLUtils.java:61)
	at org.kaazing.gateway.resource.address.uri.URIUtils.modifyURIAuthority(URIUtils.java:305)
	at org.kaazing.gateway.resource.address.tcp.TcpResourceAddressFactorySpi.newResourceAddresses0(TcpResourceAddressFactorySpi.java:203)
	at org.kaazing.gateway.resource.address.ResourceAddressFactorySpi.newResourceAddress(ResourceAddressFactorySpi.java:130)
	at org.kaazing.gateway.resource.address.ResourceAddressFactorySpi.newResourceAddress(ResourceAddressFactorySpi.java:72)
	at org.kaazing.gateway.resource.address.ResourceAddressFactorySpi.newResourceAddress(ResourceAddressFactorySpi.java:68)
	at org.kaazing.gateway.resource.address.tcp.TcpResourceAddressFactorySpiTest.shouldRequireExplicitPortOnIPv6(TcpResourceAddressFactorySpiTest.java:100)
[...]
```

### DIFF
--- a/resource.address/tcp/src/main/java/org/kaazing/gateway/resource/address/tcp/TcpResourceAddressFactorySpi.java
+++ b/resource.address/tcp/src/main/java/org/kaazing/gateway/resource/address/tcp/TcpResourceAddressFactorySpi.java
@@ -104,20 +104,7 @@ public class TcpResourceAddressFactorySpi extends ResourceAddressFactorySpi<TcpR
         if (bindAddress instanceof InetSocketAddress) {
             return (InetSocketAddress) bindAddress;
         } else if (bindAddress instanceof String) {
-            String address = (String) bindAddress;
-            // host and port
-            Pattern pattern = Pattern.compile("(.*):([0-9]*)");
-            Matcher matcher = pattern.matcher((String) bindAddress);
-            if (matcher.find()) {
-                return new InetSocketAddress(matcher.group(1), parseInt(matcher.group(2)));
-            }
-            // port only
-            try {
-                return new InetSocketAddress(parseInt(address));
-            } catch (NumberFormatException e) {
-                throw new IllegalArgumentException(
-                        String.format("%s port is mandatory for: %s", BIND_ADDRESS.name(), address));
-            }
+            return ResolutionUtils.parseBindAddress((String) bindAddress);
         }
         throw new IllegalArgumentException(BIND_ADDRESS.name());
     }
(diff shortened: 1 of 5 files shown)


resource.address/tcp/src/main/java/org/kaazing/gateway/resource/address/tcp/TcpResourceAddressFactorySpi.java
```diff
@@ -103,27 +103,22 @@ protected void parseNamedOptions0(String location, ResourceOptions options,
...
-                int port = parseInt(bindParts[1]);
-                return new InetSocketAddress(hostname, port);
-            }
-            // otherwise (more than one ":" separator encountered)
-            Pattern pattern = Pattern.compile("(\\[{1}@[a-zA-Z0-9 :]*\\]{1}):([0-9]*)");
+        } else if (bindAddress instanceof String) {
+            String address = (String) bindAddress;
+            // host and port
+            Pattern pattern = Pattern.compile("(.*):([0-9]*)");
             Matcher matcher = pattern.matcher((String) bindAddress);
```
jitsni: You can get rid of cast using "address" variable

resource.address/tcp/src/main/java/org/kaazing/gateway/resource/address/tcp/TcpResourceAddressFactorySpi.java
```diff
@@ -103,27 +103,22 @@ protected void parseNamedOptions0(String location, ResourceOptions options,
...
-                String hostname = bindParts[0];
-                int port = parseInt(bindParts[1]);
-                return new InetSocketAddress(hostname, port);
-            }
-            // otherwise (more than one ":" separator encountered)
-            Pattern pattern = Pattern.compile("(\\[{1}@[a-zA-Z0-9 :]*\\]{1}):([0-9]*)");
+        } else if (bindAddress instanceof String) {
+            String address = (String) bindAddress;
+            // host and port
+            Pattern pattern = Pattern.compile("(.*):([0-9]*)");
```
jitsni: Seems like "[::1" doesn't throw any exception
"192.168.1.1:" doesn't throw IllegalArgumentException as line no:112 throws number format exception

I suggest just catch Exception for most of the code and throw IllegalArgumentException with a generic message.

Anisotrop:
@jitsni I have updated the pull request, by re-including network interfaces support, moving the method in a general purpose class and adding unit tests. Can you please review ?

PR #462, opened 2016-03-21
Removed double creation of TcpExtensionFactory

dpwspoon:


transport/nio/src/main/java/org/kaazing/gateway/transport/nio/internal/NioSocketAcceptor.java
```diff
@@ -123,12 +123,12 @@ public NioWorker call() throws Exception {
...
 
-    NioSocketAcceptor(Properties configuration, TcpExtensionFactory extensionFactory) {
+    public NioSocketAcceptor(Properties configuration, TcpExtensionFactory extensionFactory) {
         super(configuration, LoggerFactory.getLogger(LOGGER_NAME));
         this.extensionFactory = extensionFactory;
     }
 
-    public NioSocketAcceptor(Properties configuration) {
+    NioSocketAcceptor(Properties configuration) {
         super(configuration, LoggerFactory.getLogger(LOGGER_NAME));
```
jitsni: just call the above constructor by doing
this(configuration, TcpExtensionFactory.newInstance())

PR #464, opened 2016-03-22
added jacoco(code coverage) plugin to all modules and updated travis …

AdrianCozma:
…file for SonarQube


service/update.check/pom.xml
```diff
@@ -88,6 +81,8 @@
                 <configuration>
                     <additionalClasspathElements>
                         <additionalClasspathElement>${basedir}/src/test/resources/gateway.server-5.0.0.8.jar</additionalClasspathElement>
+                        <argLine>${surefireArgLine}</argLine>
+                        <skipTests>${skipTests}</skipTests>
                     </additionalClasspathElements>
                 </configuration>
             </plugin>
```
AdrianCozma: @dpwspoon there is a warning in the Travis build related to this plugin not having a version specified, not sure how to get around it(just add a version? shouldn't it inherit the version?); also why is the special surefire configuration needed for this plugin?

mina.core/core/pom.xml
```diff
@@ -33,5 +33,14 @@
             <artifactId>easymockclassextension</artifactId>
         </dependency>
     </dependencies>
+	
+    <build>
```
dpwspoon: If this is truly in every pom (like it looks to me), lets just have it in the top level pom.  If it is not and only in the vast majority, lets have it in the top level pom and then use a [property](http://eclemma.org/jacoco/trunk/doc/prepare-agent-mojo.html#skip) to deactivate it in the poms it is not used in.  I think that would be cleaner.  Do you agree

service/update.check/pom.xml
dpwspoon: The plugin version probably is not setup in kaazing/community which is where it would inherit the version from.

dpwspoon: This module has its own configuration of surefire because it needs the additionalClasspathElement for one of its tests

AdrianCozma:
@dpwspoon I've remade the pull request, you were right it is cleaner to just have it in the root pom

Issue #463, opened 2016-03-22
wsn transport is reacting incorrectly to case where maximum ws message size is exceeded


cmebarrow:
As @Anisotrop remarked in comments in https://github.com/kaazing/gateway/pull/445, the server (wsn transport) continues to read incoming data after detecting that the incoming WebSocket frame size exceeds the configured limit. This results in further protocol decoder errors and is also in violation of RFC 6455, which states (in section [7.1.7 Fail the WebSocket connection](http://tools.ietf.org/html/rfc6455#section-7.1.7)
```
An endpoint MUST NOT continue to attempt to
   process data (including a responding Close frame) from the remote
   endpoint after being instructed to _Fail the WebSocket Connection_.
```
and section [10.4.  Implementation-Specific Limits](http://tools.ietf.org/html/rfc6455#section-10.4) implies that when the limit is exceeded the server should fail the WebSocket connection.

To reproduce:
- Alter test LimitsMaxSizeDefaultIT by adding a throw instruction as follows:
```
    public void shouldRefuseBinaryFrameWithPayloadLengthExceeding128KiB() throws Exception {
        k3po.finish();
        throw new Exception("TEMP");
    }
```
- Run the test from Eclipse, or the command line: 
`mvn verify -DskipTests  -Dit.test=LimitsMaxSizeDefaultIT `
The throw causes trace level of messages to be printed out (thanks to the MethodExecutionTrace rule). These show numerous protocol decoder exceptions after the detection of the message size exceeded exception, indicating that the server is continuing to read incoming data in violation of the spec.

sbadugu:
@cmebarrow It seems like you already have the fix for this bug based on your [comment](https://github.com/kaazing/gateway/pull/445#issuecomment-199603427). You can post the PR here and close this issue, when you are back :)

cmebarrow:
Fix is in my branch gateway_issue_463

cmebarrow:
This was fixed by PR https://github.com/kaazing/gateway/pull/466 which was merged on 28-Mar-16

PR #466, opened 2016-03-22
Issue #463: in transport/wsn, added suspendRead in WsCloseFilter.filt…

cmebarrow:
…erClose for max size exceeded case to make sure we don't try to decode the remaining data (which results in lots of protocol decoder exceptions). Enhanced LimitsMaxSizeDefaultIT to check log messages to ensure there are no longer other protocol decoder exceptions being thrown. Also updated all wsn k3po dependencies to inherit from k3po.version in root pom.


cmebarrow:
@Anisotrop Please review these changes emerge if you think they're okay. They address the issue you raised in https://github.com/kaazing/gateway/pull/445

cmebarrow:
@Anisotrop I just added a fix to WsxAcceptorLoggingIT to fix an occasional failure (not related to this PR but it prevented the build of this PR from passing)

Anisotrop:
Merging as per @cmebarrow comments.

PR #461, opened 2016-03-18
Amqp open handshake

ilyaanisimov-kaazing:

PR #467, opened 2016-03-23
AMQP 0.9.1 spec tests against GW instance

ilyaanisimov-kaazing:


ilyaanisimov-kaazing:
PR #461 was created instead

PR #470, opened 2016-03-24
Avoid possible NPE in WsnAcceptor.ioBridgeHandler.sessionClosed

cmebarrow:
(forward port of https://github.com/kaazing-private/gateway.server/pull/89)

### DIFF
--- a/transport/wsn/src/main/java/org/kaazing/gateway/transport/wsn/WsnAcceptor.java
+++ b/transport/wsn/src/main/java/org/kaazing/gateway/transport/wsn/WsnAcceptor.java
@@ -684,8 +684,8 @@ public class WsnAcceptor extends AbstractBridgeAcceptor<WsnSession, WsnBindings.
         @Override
         protected void doSessionClosed(IoSessionEx session) throws Exception {
             WsnSession wsnSession = SESSION_KEY.remove(session);
-            boolean isWsx = !wsnSession.getLocalAddress().getOption(CODEC_REQUIRED);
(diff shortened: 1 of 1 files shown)


cmebarrow:
@jitsni Easy code review for you :)

transport/wsn/src/main/java/org/kaazing/gateway/transport/wsn/WsnAcceptor.java
```diff
@@ -684,8 +684,8 @@ protected void doExceptionCaught(IoSessionEx session, Throwable cause) throws Ex
         @Override
         protected void doSessionClosed(IoSessionEx session) throws Exception {
             WsnSession wsnSession = SESSION_KEY.remove(session);
-            boolean isWsx = !wsnSession.getLocalAddress().getOption(CODEC_REQUIRED);
             if (wsnSession != null && !wsnSession.isClosing()) {
+                boolean isWsx = !wsnSession.getLocalAddress().getOption(CODEC_REQUIRED);
                 if (isWsx) {
```
jitsni: I am not sure whether if (isWsx) block is needed. Can we not just reset the wsnSession ? Why do we need to do something specific for wsx layer ?

For now, I can merge as it is

cmebarrow: I think it may be because we don't want to report a separate IOException in this case (we just want one for ws transport)

PR #477, opened 2016-03-29
Add additional logging to the heartbeat at trace level for future dia…

dpwspoon:
…gnostics of what the heartbeat is doing. Catch any exception when the heartbeat tries to connect, log it, and proceed allowing the heartbeat to be rescheduled. An example of one of these exceptions is being unable to resolve a DNS name due to network connectivity issues thrown from the TcpResourceAddressFactorySpi. When this exception bubbles up to the heartbeat the heartbeat was not getting rescheduled. That is now fixed such that when network connectivity is restored the heartbeat correctly re-binds the service in question.


jitsni:
I have already created a similar PR !!
https://github.com/kaazing/gateway/pull/471

PR #478, opened 2016-03-29
Adding a log message when a session is closed due to Connection: close

jitsni:
Adding a log message when a session is closed due to Connection: close header in HTTP request

### DIFF
--- a/transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpPersistenceFilter.java
+++ b/transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpPersistenceFilter.java
@@ -37,11 +37,11 @@ import org.slf4j.LoggerFactory;
 
 public class HttpPersistenceFilter extends HttpFilterAdapter<IoSessionEx> {
 
-	private static final String CONNECTION_CLOSE = "close";
-	private static final String HEADER_CONNECTION = "Connection";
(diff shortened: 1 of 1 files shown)


transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpPersistenceFilter.java
```diff
@@ -125,58 +125,61 @@ public void operationComplete(WriteFuture future) {
...
+        if (httpContent.isComplete()) {
+            if (session.containsAttribute(CONNECTION_CLOSE_KEY)) {
+                writeRequest.getFuture().addListener(new IoFutureListener<WriteFuture>() {
+                    @Override
+                    public void operationComplete(WriteFuture future) {
+                        IoSession session = future.getSession();
+                        if (logger.isTraceEnabled()) {
+                            logger.trace(String.format("Closing session %s because of Connection: close header in HTTP request", session));
+                        }
+
```
jitsni: These three lines are the only changes (others are replacing tabs)

cmebarrow:
@jitsni Looks fine to me but Travis build gave 9 failures in directory service. I have restarted it to see if they are random.

jitsni:
There is a pending PR https://github.com/kaazing-private/gateway.server/pull/135 in 4.x that addresses directory service failures. That need to be ported to 5.x also.

Issue #315, opened 2015-09-04
WSN Connector: Connection should fail when the negotiated protocol is not in the list of supported protocols


sanjay-saxena:
When Gateway is acting as a client, if the negotiated protocol in handshake response does not match with any of the protocols in the list of supported protocols that were included in the handshake request, then the connection must fail.

This can be reproduced using `OpeningHandshakeIT.shouldFailConnectionWhenResponseHeaderSecWebSocketProtocolNotNegotiated()` test in transport/wsn module under `org.kaazing.gateway.transport.wsn.specification.ws.connector` package.

Issue #472, opened 2016-03-25
Gateway sends response with invalid Sec-WebSocket-Protocol header


robinzimmermann:
**Gateway with version:**

5.0

**Can the bug be reproduced in a Kaazing demo out-of-the-box?**  

No

**Steps to reproduce:**  

Install ActiveMQ, a recent version that supports MQTT. You can tell, because ```ACTIVEMQ_HOME/conf/activemq.xml``` contains an MQTT `<transportConnector>`:

```
<transportConnectors>
    ...
    <transportConnector name="mqtt" uri="mqtt://0.0.0.0:1883?maximumConnections=1000&amp;wireFormat.maxFrameSize=104857600"/>
    ...
</transportConnectors>
```

Install Gateway 5.0.

Configure the gateway using the **gateway-config.xml** file at the end of this post.

Place the **mqtt.html** file at the end of this post into `GATEWAY_HOME/web/base`.

Download the [Paho MQTT Javascript library](https://raw.githubusercontent.com/eclipse/paho.mqtt.javascript/master/src/mqttws31.js) and place it in `GATEWAY_HOME/web/base`.

Open mqtt.html in a browser: http://localhost:8000/mqtt.html. This will connect, send a message, receive an echoed response, and write it to the screen.

**Observed behavior:**  

**Expected behavior:**  

**Gateway Java version**  

**Server OS with version:**  

**Back-End System with version**  

**Client Technology:**  

**Browser with Version:**  

## gateway-config.xml

```
<?xml version="1.0" encoding="UTF-8" ?>
<gateway-config xmlns="http://xmlns.kaazing.org/2014/09/gateway">

  <!-- =================================================================== -->

  <properties>
    <property>
      <name>gateway.host</name>
      <value>localhost</value>
    </property>
    <property>
      <name>gateway.port</name>
      <value>8000</value>
    </property>
  </properties>

  <!-- =================================================================== -->

  <service>
    <name>base-directory</name>
    <description>Directory service for base files</description>
    <accept>http://${gateway.host}:${gateway.port}/</accept>

    <type>directory</type>

    <properties>
      <directory>/base</directory>
      <welcome-file>index.html</welcome-file>
      <error-pages-directory>/error-pages</error-pages-directory>
      <options>indexes</options>
    </properties>

    <accept-options>
        <tcp.bind>${gateway.port}</tcp.bind>
    </accept-options>
  </service>


  <!-- =================================================================== -->

  <service>
    <name>mqtt-proxy</name>
[...]

PR #479, opened 2016-03-30
Validation of Sec-WebSocket-Protocol on the connector side

jitsni:
Validation of Sec-WebSocket-Protocol on the connector side
Since the header is comma separated value list, it is stored as multiple values

This fixes https://github.com/kaazing/gateway/issues/315


jitsni:
It needs k3po PR https://github.com/k3po/k3po/pull/322

dpwspoon:
I believe K3po has been released: 3.0.0-alpha-27

jitsni:
Yes, I integrated the latest k3po in this branch, gateway builds fines. But enterprise.gateway has some NPEs which I need to investigate.

jitsni:
Now it is ready for review and merge

dpwspoon:
I think we should revert this for now as we don't have an API to set the expected next protocol on the connector yet... And some proxies (ours included) art strict to the spec.  Once we have that API I think we should add this back in

dpwspoon:
Reasoning for reverting this is here: https://github.com/kaazing/gateway/issues/472#issuecomment-201508244

Issue #316, opened 2015-09-05
WSN Connector: Writing a BINARY(0x02) frame results in a TEXT(0x01) frame to be written out


sanjay-saxena:
When Gateway is acting as a client, writing a BINARY(0x02) frame results in a TEXT(0x01) frame to be written out. It is quite possible that `WsnConnectorRule` may need some changes to address this problem. Currently, all the tests that require Gateway to send a BINARY frame use the following hack:

``` java
    private static String TEXT_FILTER_NAME = WsnProtocol.NAME + "#text";

        WsnSession wsnConnectSession = (WsnSession) connectFuture.getSession();
        // ### Temporary hack till the issue related to Connector writing out TEXT frame instead of BINARY is resolved.
        if (wsnConnectSession != null) {
            IoFilterChain parentFilterChain = wsnConnectSession.getParent().getFilterChain();
            if (parentFilterChain.contains(TEXT_FILTER_NAME)) {
                parentFilterChain.remove(TEXT_FILTER_NAME);
            }
        }
```

The following tests in transport/wsn module in `org.kaazing.gateway.transport.wsn.specification.ws.connector` package use the aforementioned hack:

0. `FragmentationIT.shouldEchoServerSendBinaryFrameWithEmptyPayloadFragmentedAndInjectedPingPong()`
0. `FragmentationIT.shouldEchoServerSendBinaryFrameWithPayloadFragmented()`
0. `FragmentationIT.shouldEchoServerSendBinaryFrameWithPayloadFragmentedAndInjectedPingPong()`
0. `FragmentationITshouldEchoServerSendBinaryFrameWithPayloadFragmentedWithSomeEmptyFragments()`
0. `FragmentationIT.shouldEchoServerSendBinaryFrameWithPayloadNotFragmented()`

0. `BaseFramingIT.shouldEchoBinaryFrameWithPayloadLength0()`
0. `BaseFramingITshouldEchoBinaryFrameWithPayloadLength125()`
0. `BaseFramingIT.shouldEchoBinaryFrameWithPayloadLength126()`
0. `BaseFramingIT.shouldEchoBinaryFrameWithPayloadLength127()`
0. `BaseFramingIT.shouldEchoBinaryFrameWithPayloadLength128()`
0. `BaseFramingIT.shouldEchoBinaryFrameWithPayloadLength65535()`
0. `BaseFramingIT.shouldEchoBinaryFrameWithPayloadLength65536()`

PR #480, opened 2016-03-30
WsnConnector doesn't send text frame by default.

jitsni:
If X-Frame-Type header is used to specify encoding. If it was not there, it was assumed as text encoding. That is removed so that so that binary frames are not written as text frames. This fixes https://github.com/kaazing/gateway/issues/316

### DIFF
--- a/transport/wsn/src/main/java/org/kaazing/gateway/transport/wsn/WsnConnector.java
+++ b/transport/wsn/src/main/java/org/kaazing/gateway/transport/wsn/WsnConnector.java
@@ -147,21 +147,20 @@ public class WsnConnector extends AbstractBridgeConnector<WsnSession> {
     public void addBridgeFilters(IoFilterChain filterChain) {
         IoSession session = filterChain.getSession();
         Encoding encoding = (Encoding) session.getAttribute(ENCODING_KEY);
-
-        switch (encoding) {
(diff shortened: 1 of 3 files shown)


transport/wsn/src/main/java/org/kaazing/gateway/transport/wsn/WsnConnector.java
```diff
@@ -581,8 +580,6 @@ public WsnSession call() throws Exception {
                         parent.setAttribute(ENCODING_KEY, Encoding.BINARY);
                     } else if ("base64".equals(frameType)) {
                         parent.setAttribute(ENCODING_KEY, Encoding.BASE64);
-                    } else {
-                        parent.setAttribute(ENCODING_KEY, Encoding.TEXT);
```
jitsni: Not sure why connector was sending text messages by default. I don't know if there are any backward compatible issues here by this change

cmebarrow: I think this is historical. X-Frame-Type header is set by WsnAcceptor in doUpgrade76, so only when using old draft-76 WebSocket protocol, to indicate whether the entire conversation will be text or binary I think.

PR #483, opened 2016-03-31
Fix for HttpDirectoryServiceIT failure

mgherghe:


service/http.directory/src/main/java/org/kaazing/gateway/service/http/directory/HttpDirectoryServiceHandler.java
```diff
@@ -104,6 +104,10 @@ void setPatterns(List<PatternCacheControl> patterns) {
         this.patterns = patterns;
     }
 
+    void emptyUrlCacheControlMap() {
+        urlCacheControlMap.clear();
+    }
```
jitsni: Cleaning up the resources is good, but I wonder why this is affecting the tests. Then when I see that urlCacheControlMap is static. So two questsions:
* I wonder why we need to keep it as static.
* urlCacheControlMap is accessed from multiple threads(say two clients are acessing the same directory service) , but currently it uses HashMap and is not thread-safe.

mgherghe: The urlCacheControlMap is static so that it caches the resolved header for each url. If it weren't static, this would have to be resolved for every url, at each request.

jitsni: I don't understand. Can you elaborate bit here. There is only one HttpDirectoryServiceHandler instance per service. Then I don't see any need for static unless you want to cache across multiple services. All the requests for a service are handled at one handler. So static map is not required.  What am I missing ?

Also, you haven't commented on thread-safety issue.

mgherghe:
@jitsni I created [PR](https://github.com/kaazing/gateway/pull/521) to address your above comments.

PR #488, opened 2016-04-02
Ilya amqp open handshake clean

cmebarrow:
This PR has the same content as Ilya's PR https://github.com/kaazing/gateway/pull/476 except for two small changes to pom.xml files to pick up the correct k3po version for the amqp specification scripts.


cmebarrow:
Build failed because the latest released k3po is not yet available in the maven repository. Need to try later.

PR #498, opened 2016-04-06
Adding a log message when binding is not found and 404 is sent

jitsni:


jitsni:
I want this changeset to be in release branch too

PR #500, opened 2016-04-06
Added on.client.message property to broadcast service, which when set…

dpwspoon:
… to "broadcast" broadcasts the client messages. 

Also did some xsd clean up that wasn't caught yet


dpwspoon:
Now tracked here: https://github.com/kaazing/gateway/pull/501

PR #501, opened 2016-04-06
Added on.client.message property to broadcast service, which when set…

dpwspoon:
… to "broadcast" broadcasts the client messages.

Also did some xsd clean up that wasn't caught yet

### DIFF
--- a/service/broadcast/src/main/java/org/kaazing/gateway/service/broadcast/BroadcastServiceHandler.java
+++ b/service/broadcast/src/main/java/org/kaazing/gateway/service/broadcast/BroadcastServiceHandler.java
@@ -15,6 +15,8 @@
  */
 package org.kaazing.gateway.service.broadcast;
 
+import static org.kaazing.gateway.service.broadcast.BroadcastService.OnClientMessage.BROADCAST;
+
 import java.nio.channels.ClosedChannelException;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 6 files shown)


service/broadcast/src/main/java/org/kaazing/gateway/service/broadcast/BroadcastServiceHandler.java
```diff
@@ -59,6 +63,13 @@ IoHandler getListenHandler() {
     }
 
     @Override
+    public void messageReceived(IoSession session, Object message) throws Exception {
+        if (onClientMessage == BROADCAST) {
+            handler.messageReceived(session, message);
```
jitsni: Two questions:
- It broadcasts to all the clients (including itself)
- Do you see any thread alignment issues ?

service/broadcast/src/test/java/org/kaazing/gateway/service/broadcast/BroadcastServiceIT.java
```diff
@@ -63,8 +69,20 @@ public void tcpConnectionToBackendSendAndClose() throws Exception {
 
 	@Specification("broadcast.frontend.tcp.connect.and.close")
```
jitsni: indentation for the method need to be fixed

service/broadcast/src/main/java/org/kaazing/gateway/service/broadcast/BroadcastServiceHandler.java
dpwspoon: > It broadcasts to all the clients (including itself)

Yes it does

PR #503, opened 2016-04-06
Reverting the change where ssl tests are run in forked mode.

jitsni:
Needed for some tests which deal with Security.setProperty config


dpwspoon:
CCing @AdrianCozma

PR #518, opened 2016-04-13
added certificate to travis and enabled management tests for travis

AdrianCozma:


AdrianCozma:
@dpwspoon the build failed in management but for a different test(`testServiceDefaultsWithSslCiphers`) and doesn't seem related to my changes in this PR

Is it possible that the cause for the failure is because in the `travis.yml` I added `sudo: required`?(this changes the build machine from a container to a VM as far as I know)

PR #521, opened 2016-04-13
Removed static from urlCacheControlMap and made it thread safe

mgherghe:
This was only tested by reloading the demo page several times and checking that the urlCacheControlMap is properly populated.
All tests from the gateway.service.http.directory module passed.

### DIFF
--- a/service/http.directory/src/main/java/org/kaazing/gateway/service/http/directory/HttpDirectoryServiceHandler.java
+++ b/service/http.directory/src/main/java/org/kaazing/gateway/service/http/directory/HttpDirectoryServiceHandler.java
@@ -22,12 +22,12 @@ import java.text.DateFormat;
 import java.text.DecimalFormat;
 import java.text.SimpleDateFormat;
 import java.util.Date;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.TimeZone;
 import java.util.TreeMap;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.kaazing.gateway.service.ServiceContext;
 import org.kaazing.gateway.service.http.directory.cachecontrol.CacheControlHandler;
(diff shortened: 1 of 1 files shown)


dpwspoon:
Do we want this on the release branch?

service/http.directory/src/main/java/org/kaazing/gateway/service/http/directory/HttpDirectoryServiceHandler.java
```diff
@@ -53,7 +53,7 @@
     private boolean indexes;
 
     private List<PatternCacheControl> patterns;
-    private static Map<String, CacheControlHandler> urlCacheControlMap = new HashMap<String, CacheControlHandler>();
+    private Map<String, CacheControlHandler> urlCacheControlMap = new ConcurrentHashMap<String, CacheControlHandler>();
```
jitsni: OK, static is gone. But just using ConcurrentHashMap itself doesn't give thread-safety for composite operations (for e.g contains and put). You may want to try something like the following. The idea is to compute CacheControlHandler atomically (when it is already in the map or not). See if it works :

```
    private void addCacheControl(HttpAcceptSession session, File requestFile, String requestPath) {
        CacheControlHandler cacheControlHandler = urlCacheControlMap.computeIfAbsent(requestPath, path ->
             patterns.stream()
                     .filter(patternCacheControl -> PatternMatcherUtils.caseInsensitiveMatch(path, patternCacheControl.getPattern()))
                     .findFirst()
                     .map(patternCacheControl -> new CacheControlHandler(requestFile, patternCacheControl))
                     .orElse(null)
        );

        if (cacheControlHandler != null) {
            addCacheControlHeader(session, requestFile, cacheControlHandler);
        }
}
```

I don't see any operations removing from the map (except clearing the whole thing). Don't the entries in the map be removed once they are expired ?

jitsni:
Also, do you want this in the release ?

service/http.directory/src/main/java/org/kaazing/gateway/service/http/directory/HttpDirectoryServiceHandler.java
mgherghe: The map is used to match the URL of a requested file with a pattern from the configuration file. That is only done once and needs to be cached, so there is no need to remove the entries. 
The max-age and any other information that may change is computed at the request time.

mgherghe:
Recreated [PR](https://github.com/kaazing/gateway/pull/526) against `release/5.1.0` branch.

PR #526, opened 2016-04-14
Removed static from urlCacheControlMap and made it thread safe

mgherghe:


mgherghe:
This should also be ported to gateway 4.x.

mgherghe:
Created [PR](https://github.com/kaazing-private/gateway.server/pull/142) for gateway 4.x

PR #540, opened 2016-04-20
Updated k3po version

dpwspoon:


dpwspoon:
https://github.com/kaazing/tickets/issues/448 has been filed for the bug exposed here

PR #545, opened 2016-04-21
Fix OpeningHandshakeIT

Anisotrop:
Method sessionOpened might not be called before connectFuture.awaitUninterruptibly() exits.


dpwspoon:
I've confirmed with Jitu that we believe there is no guarantee on a connect future and a session open.  So this is the correct fix.

PR #548, opened 2016-04-22
Thread unsafe processing in ProtocolCodecFilter during wseb realignment

jitsni:
https://github.com/kaazing/tickets/issues/450

ProtocolDecoderOutput#flush() is being called from two threads during realignment
1. The thread does the decoding and calls flush()
2. Session is realigned during flush (as it call messageReceived on the upper layer)
   REGISTER_EVENT is sent by setIoAlignment, and that calls flush again

So, synchronizing the access to flush(). Also, old thread doesn't continue
processing the decoding and allows new thread to take over decoding/processing


jfallows:
Looks great, @jitsni, including the break-early logic to let the initial thread unwind quickly during realignment. 😄

PR #377, opened 2015-12-15
Forward port to 5.x: (tickets#180) Avoid ClassCastException in LoggingFilter using socks transport #88

msalavastru:

PR #549, opened 2016-04-25
Fix for issue 377 - tcp or http bind on service with accepts matching Enterprise shield socks host and port cause binding conflict 

msalavastru:
This is a partial fix for [issue #377](https://github.com/kaazing/tickets/issues/377). Although this fixes the issue, some other tests fails because of this change, so it is still in progress.


msalavastru:
Closing this as it comes from develop branch.

PR #550, opened 2016-04-25
Fix for issue 377 - tcp or http bind on service with accepts matching Enterprise shield socks host and port cause binding conflict

msalavastru:
This is a partial fix for [issue #377](https://github.com/kaazing/tickets/issues/377). Although this fixes the issue, some other tests fails because of this change, so it is still in progress.

### DIFF
--- a/resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/Comparators.java
+++ b/resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/Comparators.java
@@ -35,6 +35,7 @@ public final class Comparators {
     private static final Comparator<String> STRING_COMPARATOR = compareComparable(String.class);
     private static final Comparator<ResourceAddress> ORIGIN_COMPARATOR = compareNonNull(new ResourceOriginComparator());
     private static final Comparator<ResourceAddress> LOCATION_COMPARATOR = compareNonNull(new ResourceLocationComparator());
+    private static final Comparator<ResourceAddress> NON_NULL_LOCATION_ALTERNATES_COMPARATOR = compareNonNull(new LocationAlternatesComparator());
     private static final Comparator<ResourceAddress> LOCATION_ALTERNATES_COMPARATOR = compareAlternates(new ResourceLocationComparator());
     private static final Comparator<ResourceAddress> PROTOCOL_STACK_COMPARATOR = compareNonNull(new ResourceProtocolStackComparator());
     private static final Comparator<ResourceAddress> ORIGIN_AND_PROTOCOL_STACK_COMPARATOR = new ResourceOriginAndProtocolStackComparator();
@@ -356,7 +357,7 @@ public final class Comparators {
         @Override
         public int compare(ResourceAddress addr1, ResourceAddress addr2) {
 
-            int compareAlternates = compareAlternatesLocation(addr1, addr2);
+            int compareAlternates = NON_NULL_LOCATION_ALTERNATES_COMPARATOR.compare(addr1, addr2);
             if (compareAlternates != 0) {
                 return compareAlternates;
             }
@@ -364,41 +365,24 @@ public final class Comparators {
             ResourceAddress transport1 = addr1.getOption(TRANSPORT);
             ResourceAddress transport2 = addr2.getOption(TRANSPORT);
 
-            int compareTransportAlternate = compareAlternatesLocation(transport1, transport2);
-            if (compareTransportAlternate != 0) {
-                return compareTransportAlternate;
+            int compareTransportAlternates = NON_NULL_LOCATION_ALTERNATES_COMPARATOR.compare(transport1, transport2);
+            if (compareTransportAlternates != 0) {
+                return compareTransportAlternates;
             }
 
             return 0;
         }
+    }
 
-        private int compareAlternatesLocation(ResourceAddress addr1, ResourceAddress addr2) {
-            // TODO wrap in non null comparator
-            if (addr1 == null) {
-                return (addr2 == null) ? 0 : -1;
-            } else if (addr2 == null) {
-                return 1;
-            }
+    private static final class LocationAlternatesComparator implements Comparator<ResourceAddress>{
+        
+        @Override
+        public int compare(ResourceAddress addr1, ResourceAddress addr2) {
+
             ResourceAddress alternate1 = addr1.getOption(ALTERNATE);
             ResourceAddress alternate2 = addr2.getOption(ALTERNATE);
-            InetAddress iaddr1 = null;
-            InetAddress iaddr2 = null;
-            if (alternate1 != null) {
-                String alternate1Resource = alternate1.getResource().toString();
-                String alternate1Host = URIUtils.getHost(alternate1Resource);
-                try {
-                    iaddr1 = InetAddress.getByName(alternate1Host);
-                } catch (UnknownHostException e) {
-                }
-            }
-            if (alternate2 != null) {
-                String alternate2Resource = alternate2.getResource().toString();
-                String alternate2Host = URIUtils.getHost(alternate2Resource);
-                try {
-                    iaddr2 = InetAddress.getByName(alternate2Host);
-                } catch (UnknownHostException e) {
-                }
-            }
+            InetAddress iaddr1 = createIntetAddressFromAlternate(alternate1);
+            InetAddress iaddr2 = createIntetAddressFromAlternate(alternate2);
 
             if (isIPv6(iaddr1) && alternate2 == null || isIPv6(iaddr2) && alternate1 == null) {
                 URI resource1 = addr1.getResource();
@@ -412,6 +396,18 @@ public final class Comparators {
             }
 
             return 0;
+        }
+
+        private InetAddress createIntetAddressFromAlternate(ResourceAddress alternate) {
+            if (alternate != null) {
+                String alternateResource = alternate.getResource().toString();
+                String alternateHost = URIUtils.getHost(alternateResource);
+                try {
+                    return InetAddress.getByName(alternateHost);
+                } catch (UnknownHostException e) {
+                }
+            }
+            return null;
         }
 
         private boolean isIPv6(InetAddress iaddr) {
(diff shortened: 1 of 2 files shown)


resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/Comparators.java
```diff
@@ -338,6 +346,49 @@ public int compare(ResourceAddress addr1, ResourceAddress addr2) {
...
+            URI resource2 = addr2.getResource();
+
+            String alternate1IPv6 = alternate1 != null ? resource1.toString() : null;
+            String alternate2IPv6 = alternate2 != null ? resource2.toString() : null;
+
+            boolean alt1HasIPv6 = alternate1IPv6 != null ? alternate1IPv6.contains("[0:0:0:0:0:0:0:1]") : false;
+            boolean alt2HasIPv6 = alternate2IPv6 != null ? alternate2IPv6.contains("[0:0:0:0:0:0:0:1]") : false;
+
+            if (alt1HasIPv6 && alternate2 == null || alt2HasIPv6 && alternate1 == null) {
+                return resource1.compareTo(resource2);
```
msalavastru: This is for comparing two TcpResourceAddresses such as: 
```
[tcp://127.0.0.1:65535 (tcp://127.0.0.1:65535)]
```
with
```
[tcp://127.0.0.1:65535 (tcp://localhost:65535) socks/5+]
[tcp://[0:0:0:0:0:0:0:1]:65535 (tcp://localhost:65535) socks/5+]
```
Here we have alternate on the second one as IPv6, and no alternate on the first one, but the two should be identical, so we do the comparison based only on resourceURI.

resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/Comparators.java
```diff
@@ -325,6 +328,11 @@ public int compare(ResourceAddress addr1, ResourceAddress addr2) {
                 return compareLocation;
             }
 
+            int compareAlternateAndTransportLocation = ALTERNATE_AND_TRANSPORT_LOCATION_COMPARATOR.compare(addr1, addr2);
+            if(compareAlternateAndTransportLocation != 0) {
+               return compareAlternateAndTransportLocation;
```
msalavastru: Here we compare the alternates of the transport for ssl (which are TcpResourceAddresses, with or without alternates), and the alternates of TcpResourceAddresses. This is somewhat inline with what equals from ResourceAddress does (if one has alternate and the other doesn't), and also check alternates agains each other using the AlternateResourceAddressComparator).

dpwspoon:
Note, as stated in https://github.com/kaazing/tickets/issues/377#issuecomment-214393652 this causes 
>
some other tests to fail (such as those in MultipleOnPremiseGatewaysAndSingleCloudGatewayTest from enterprise.gateway.transport.socks)

and is a work in progress

resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/Comparators.java
```diff
@@ -338,6 +346,48 @@ public int compare(ResourceAddress addr1, ResourceAddress addr2) {
...
+            String alternate1IPv6 = alternate1 != null ? alternate1.getResource().toString() : null;
+            String alternate2IPv6 = alternate2 != null ? alternate2.getResource().toString() : null;
+
+            boolean alt1HasIPv6 = alternate1IPv6 != null ? alternate1IPv6.contains("[0:0:0:0:0:0:0:1]") : false;
+            boolean alt2HasIPv6 = alternate2IPv6 != null ? alternate2IPv6.contains("[0:0:0:0:0:0:0:1]") : false;
+
+            if (alt1HasIPv6 && alternate2 == null || alt2HasIPv6 && alternate1 == null) {
+                URI resource1 = addr1.getResource();
+                URI resource2 = addr2.getResource();
+                return resource1.compareTo(resource2);
```
msalavastru: This is for comparing two TcpResourceAddresses such as:

[tcp://127.0.0.1:65535 (tcp://127.0.0.1:65535)]
with

[tcp://127.0.0.1:65535 (tcp://localhost:65535) socks/5+]
[tcp://[0:0:0:0:0:0:0:1]:65535 (tcp://localhost:65535) socks/5+]
Here we have alternate on the second one as IPv6, and no alternate on the first one, but the two should be identical, I think, so we do the comparison based only on resourceURI.

resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/Comparators.java
```diff
@@ -338,6 +351,77 @@ public int compare(ResourceAddress addr1, ResourceAddress addr2) {
...
+            if (alternate2 != null) {
+                String alternate2Resource = alternate2.getResource().toString();
+                String alternate2Host = URIUtils.getHost(alternate2Resource);
+                try {
+                    iaddr2 = InetAddress.getByName(alternate2Host);
+                } catch (UnknownHostException e) {
+                }
+            }
+
+            if (isIPv6(iaddr1) && alternate2 == null || isIPv6(iaddr2) && alternate1 == null) {
```
msalavastru: See previous commit for explanation as to why this check is done.

jitsni:
Is there a testcase I can run in IDE and see the addresses that are being compared ?

msalavastru:
@jitsni Here is a [PR](https://github.com/kaazing/enterprise.gateway/pull/306) with the test that I created in EE to check the fix. It uses a similar configuration as the one mentioned by @dpwspoon in the bug. I also added the keystore.db with the certificate for localhost in that PR.
The test was also attached as a comment in the bug [here](https://github.com/kaazing/tickets/issues/377#issuecomment-214393652).

Using the above test, the comparators will check:
```
[ssl://localhost:443 (ssl://localhost:443) http/1.1]
  [tcp://127.0.0.1:443 (tcp://localhost:443) ssl]
  [tcp://[0:0:0:0:0:0:0:1]:443 (tcp://localhost:443) ssl]
```
vs
```
[ssl://localhost:443 (ssl://localhost:443) http/1.1]
  [tcp://127.0.0.1:443 (tcp://localhost:443) ssl]
  [tcp://192.168.1.11:443 (tcp://localhost:443) ssl]
  [tcp://192.168.1.12:443 (tcp://localhost:443) ssl]
  [tcp://192.168.1.13:443 (tcp://localhost:443) ssl]
```
the result being that they are not equal.

Running another test, that already existed in EE `shouldBuildSocksSessionsUsingAReverseAcceptorClient` from `SocksReverseTransportClientTest` will cause the comparators to check:
```
[tcp://127.0.0.1:4444 (tcp://127.0.0.1:4444)]
```
vs
```
[tcp://127.0.0.1:4444 (tcp://localhost:4444) socks/5+]
[tcp://[0:0:0:0:0:0:0:1]:4444 (tcp://localhost:4444) socks/5+]
```
the result being that they are equal.

jitsni:
For some reason, PR is not updated properly here. I see different diff in IDE/commandline.
It needs more work
- I don't want IPv6 etc in the Comparators
- Since it is changing all the NextBindings comparator, we need to see where all it will affect. For e.g. socks uses NextBindings

jitsni:
For some reason, the PR is not updated here (diff using git commands is different). Here is the feedback.
* We don't want all the ipv6 related things in Comparators
* I think we want to use compareAlternates(), something like the following
```
    private static final Comparator<ResourceAddress> LOCATION_AND_TRANSPORT_PROTOCOL_STACK_COMPARATOR = compareAlternates(new ResourceLocationAndTransportProtocolStackComparator());
```
But that is also running into some issues. Perhaps, you can explore that approach.
* It is bit risky to changes comparators without testing/thinking it thoroughly
* The test can use 8443 and 8080 ports (otherwise, it cannot be run easily)

msalavastru:
@jitsni I have removed the IPv6 check, and made changes to the Comparators.

I have added a comparator for externalURI for the case where the ResourceAddress is constructed from the NioSocketChannelIOSession (`ResourceAddress localAddress = asResourceAddress(session.getLocalAddress())`) and the externalURI will be IP rather than hostname. In this case I no longer compare the alternates, so I can get the resourceaddress that already exists in the NexProtocolBindings:

```
[tcp://127.0.0.1:4444 (tcp://127.0.0.1:4444)]

compared to

[tcp://127.0.0.1:4444 (tcp://localhost:4444) socks/5+]
[tcp://[0:0:0:0:0:0:0:1]:4444 (tcp://localhost:4444) socks/5+]
```
The tests seem to pass with this change, but I will investigate further on the impact of the change.

resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/Comparators.java
```diff
@@ -327,15 +330,35 @@ public int compare(ResourceAddress addr1, ResourceAddress addr2) {
...
-            
+
             int compareTransport = PROTOCOL_STACK_COMPARATOR.compare(transport1, transport2);
             if (compareTransport != 0) {
                 return compareTransport;
             }
-            
+
+            int compareExternalURI = EXTERNAL_URI_COMPARATOR.compare(addr1, addr2);
+            if (compareExternalURI == 0) {
```
jitsni: We shouldn't use external URI in comparisons (for e.g dns might map foo.com and bar.com to the same ip).

resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/Comparators.java
```diff
@@ -29,6 +30,8 @@
     private static final Comparator<String> STRING_COMPARATOR = compareComparable(String.class);
     private static final Comparator<ResourceAddress> ORIGIN_COMPARATOR = compareNonNull(new ResourceOriginComparator());
     private static final Comparator<ResourceAddress> LOCATION_COMPARATOR = compareNonNull(new ResourceLocationComparator());
+    private static final Comparator<ResourceAddress> LOCATION_ALTERNATES_COMPARATOR = compareAlternates(new ResourceLocationComparator());
```
jitsni: This looks good to me. I think we should explore in this direction.

jitsni:
See my changes in my branch issue377.address.alternates in both gateway and enterprise.gateway project. I think some tests are failing (It works for ssl etc, but need to see if this approach works for tcp as tcp does physical binds)

msalavastru:
I have updated with the changes from code review. On Travis, all the tests are passing.

jitsni:
* I see these are failing in my local build (since now there is only one binding). I think these tests needs to be fixed. I see that Sslv3Test is also failing.

  NioSocketAcceptorBindingsTest.bindFollowedByTwoUnbindsSucceeds:204 expected:<1> but was:<3>
  NioSocketAcceptorBindingsTest.referenceCountingWorksForOneOrMoreBinds:164 expected:<1> but was:<3>

There may be some maven confusion since the release branch already uses baked versions instead of SNAPSHOT versions so travis may not be picking up the right versions.

* It is also good to add a test may be ssl transport or so (by creating an ssl resource address with alternates, 1. using comparator to test equality 2. SslAcceptor to bind those addresses)

PR #551, opened 2016-04-25
Fixed HttpBalancerService to add bindings properly

sanjay-saxena:


service/http.balancer/src/main/java/org/kaazing/gateway/service/http/balancer/HttpBalancerService.java
```diff
@@ -179,7 +179,8 @@ public void destroy() throws Exception {
                                                     TransportFactory transportFactory) throws Exception {
         List<String> httpURIs = new ArrayList<>(uris.size());
         for (String uri : uris) {
-            Protocol protocol = transportFactory.getProtocol(uri);
+            String schemeFromAcceptURI = uri.substring(0, uri.indexOf(':'));
```
jitsni: May be you can use getScheme(uri) like in toHttpBalancerURIs() method

sanjay-saxena: I thought about it but eventually decided not to because `getScheme()` creates an `java.net.URI` object. I wanted to avoid creating a redundant instance of `java.net.URI` object and contribute to the garbage as I thought that the intent was to move away from using `java.net.URI`.

PR #553, opened 2016-04-27
Fix for sporadic failure of HttpDirectoryServiceIT.testPostLargeData

mgherghe:


jitsni:
Looks good to be. I wonder how this test is passing on other platforms.

PR #558, opened 2016-05-04
Fix webservice url for community/enterprise.

Anisotrop:
Product name doesn't change between community and enterprise editions, it is the product edition that reads the entry "Kaazing-Product" from the manifest file which is "Community.Gateway" and "Enterprise.Gateway" respectively. This is done in class org.kaazing.gateway.server.impl.VersionUtils. Added test to show the different urls in the two cases. Luckily the implementation allows for simple fields change, since static methods can't be mocked with jMock.


sbadugu:
@Anisotrop Please create a new PR on develop branch. We will not deliver this feature in 5.x.

PR #562, opened 2016-05-06
ignored sseIe8HttpxeConnectAndGetData test for Jenkins builds

AdrianCozma:


transport/sse/src/test/java/org/kaazing/gateway/transport/sse/SseSameOriginIT.java
```diff
@@ -63,6 +64,9 @@ public void sseIe8HttpxeConnect() throws Exception {
     @Specification("sse.connect.and.get.data.via.ie8.httpxe")
     @Test
     public void sseIe8HttpxeConnectAndGetData() throws Exception {
+        Assume.assumeFalse(
```
dpwspoon: We should ignore this on all systems, as it reproduces everywhere

PR #564, opened 2016-05-09
Added test file that checks the server's pom.xml has consistent manif…

Anisotrop:
…estEntries with update.check dependency.


server/src/test/java/org/kaazing/gateway/server/SettingsTest.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.xpath.XPath;
+import javax.xml.xpath.XPathConstants;
+import javax.xml.xpath.XPathExpression;
+import javax.xml.xpath.XPathFactory;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Created by cochiriac on 5/9/16.
```
dpwspoon: I don't think we need this comment...

server/src/test/java/org/kaazing/gateway/server/SettingsTest.java
```diff
@@ -0,0 +1,55 @@
...
+    @Test
+    public void shouldHaveCommunityProductEditionAndTitle() {
+
+        try {
+            DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+            DocumentBuilder builder = factory.newDocumentBuilder();
+            Document doc = builder.parse("../server/pom.xml");
+            XPathFactory xPathfactory = XPathFactory.newInstance();
+            XPath xpath = xPathfactory.newXPath();
+            XPathExpression expr = xpath.compile("/project/build/pluginManagement/plugins/plugin[2]/configuration/archive/manifestEntries/Implementation-Title/text()");
```
dpwspoon: This seems to be relying on the plugin being setup correctly, not that the actual result of compilation is correct.  Did you try putting this in a SettingsIT and then reading from the manifest that is in the jar directly?

server/src/test/java/org/kaazing/gateway/server/SettingsIT.java
```diff
@@ -0,0 +1,51 @@
...
+    public void shouldHaveCommunityProductEditionAndTitle() throws IOException {
+        File artefact = null;
+        File target = new File("../server/target");
+        for (File entry : target.listFiles()) {
+            if (entry.getName().startsWith("gateway.server") &&
+                    entry.getName().endsWith(".jar")) {
+                artefact = entry;
+            }
+        }
+        assertNotNull("Artefact for gateway.server not found. Please run this test as an integration test.", artefact);
```
dpwspoon: Nit: artifact typo

server/src/test/java/org/kaazing/gateway/server/SettingsIT.java
```diff
@@ -0,0 +1,51 @@
...
+public class SettingsIT {
+
+    /**
+     * Checks in the projects's pom.xml that the manifest entries will be generated as expected by others (update.check)
+     */
+    @Test
+    public void shouldHaveCommunityProductEditionAndTitle() throws IOException {
+        File artefact = null;
+        File target = new File("../server/target");
+        for (File entry : target.listFiles()) {
```
dpwspoon: This is ok, but it would be better to get the path from the classpath.  For example I'm not sure how this might work in eclipse or multi module builds.. it might be okay but classpath will always work.

Alternatively you can use this Utility class that is already written for you: https://github.com/kaazing/gateway/blob/71d1188b365bcfadf12320c4e6d7b354d9f887d6/server.api/src/main/java/org/kaazing/gateway/server/impl/VersionUtils.java.  (Classpath logic is written in there)

Anisotrop:
Using a new approach that tries to calculate the jar's location based on the root location of the test-classes in the current module.

PR #568, opened 2016-05-10
Fix for issue 385

msalavastru:
This is a fix for [issue 385](https://github.com/kaazing/tickets/issues/385).


dpwspoon:
This is a candidate to go on the release branch?

irina-mitrea-luxoft:
@sbadugu Hello, Satish, can you please help us with your feedback on the above comment posted by David? The PR was initially created on the develop branch in order to not introduce new items in the release branch unless of a critical nature.

sbadugu:
Thanks for raising the concern @irina-mitrea-luxoft ! We are not certifying `http.proxy` feature for this release, so that is the reason why I initially suggested to create a PR on develop branch and not on release branch.

However, after speaking to @dpwspoon, it seems that this bug is affecting `repository.kaazing.com/maven` which is used by HSBC. So, if @jitsni gives approval to add this change to `release` branch then @msalavastru can create PR on release branch for this fix. 

Just to be clear, this is just an exception and all the other future bug fixes for http.proxy will be on develop branch, unless specified.

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpResponseDecodingState.java
```diff
@@ -143,6 +143,9 @@ protected DecodingState finishDecode(IoBuffer readData, ProtocolDecoderOutput ou
                             httpResponse
                                     .setContent(new HttpContentMessage(allocator.wrap(allocator.allocate(0)), false));
                             out.write(httpResponse);
+                            if (status == HttpStatus.REDIRECT_NOT_MODIFIED) {
+                                return null;
+                            }
```
jitsni: This seems like a wrong check here. Can you paste the HTTP response here ?

jitsni: RFC 7232 says "A 304 response cannot contain a message-body; it is always terminated by the first empty line after the header fields."

So, let us not do any content processing. I suppose something like the following should work :

```
            if (status == HttpStatus.REDIRECT_NOT_MODIFIED) {
                httpResponse.setContent(new HttpContentMessage(allocator.wrap(allocator.allocate(0)), true));
                out.write(httpResponse);
                return null;
            } else if (httpSession != null && httpSession.getMethod() == HttpMethod.HEAD) {
            ...
```

msalavastru:
Thank you @jitsni . I have updated the PR with the suggested change. Please let me know if I should move the PR to release.

jitsni:
Can we create a test for this ?  May be origin server sends 304 and gateway sends it to client.

sbadugu:
@msalavastru Please create the PR for release branch and create tests as mentioned above. Jitu has confirmed it's safe to have this change in EE 5.x release. Thanks !

dpwspoon:
I've tested this against the internal setup and local directory setup.  This fix looks good to me.  We should still add a test if possible (Integration or unit).

msalavastru:
Closing this as it was moved and merged on the release branch.

PR #570, opened 2016-05-11
Removing IdleTimeoutFilter in wseb 

jitsni:
Removing IdleTimeoutFilter in wseb as it sends PONG (0x8A) and 
some clients don't understand it. WsebDownStreamHandler sends
NOOP command when writer is idle, so it works out fine.

### DIFF
--- a/transport/wseb/src/main/java/org/kaazing/gateway/transport/wseb/WsebAcceptor.java
+++ b/transport/wseb/src/main/java/org/kaazing/gateway/transport/wseb/WsebAcceptor.java
@@ -628,7 +628,14 @@ public class WsebAcceptor extends AbstractBridgeAcceptor<WsebSession, Binding> {
                     wsebSession.setPingEnabled(pingEnabled);
                     IoSessionEx extensionsSession = wsebSession.getTransportSession();
                     IoFilterChain extensionsFilterChain = extensionsSession.getFilterChain();
-                    WsUtils.addExtensionFilters(negotiated, extensionHelper, extensionsFilterChain, false);
+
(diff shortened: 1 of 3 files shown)


jitsni:
@jfallows @cmebarrow can you review it. thanks.

dpwspoon:
This looks good to me.  Nice test @jitsni.

jitsni:
@sanjay-saxena during the review with @jfallows and @cmebarrow one concern came up
- Do any of the *newer* clients rely on PONGs ? since now PONGs will not be sent by wseb as there is no IdleTimeoutFilter. It is the same behaviour as 4.x gateway but *newer clients* may be relying on PONGs (instead of NOOPs) for implementing x-kaazing-idle-timeout extension

Going forward we modify the semantics of X-Accept-Commands to specify ping, pong. The IdleTimeoutFilter is added only when the client indicates it accepts PONGs. We will update x-kaazing-idle-timeout extension spec accordingly.

sanjay-saxena:
@jitsni -- Not sure what you mean by *newer* clients. For all practical purposes, we only have 4.x based clients.

If the concern is about `netx` based Java client, then it doesn't have any WSE implementation.

cmebarrow:
@jitsni Judging from Sanjay's comment we should be ok. The changes look fine but I would suggest you add the change to make the skipping of the idetimeout filter conditional on the X-Accept-Commands header not containing both ping and pong. Or you could file a separate bug to address that later.

jitsni:
I filed an issue https://github.com/kaazing/gateway/issues/572 to include IdleTimeoutFilter conditionally.

jfallows:
LGTM.

PR #573, opened 2016-05-13
Adding a better log message when WsCheckAliveFilter closes ws session

jitsni:


transport/ws/src/main/java/org/kaazing/gateway/transport/ws/bridge/filter/WsCheckAliveFilter.java
```diff
@@ -45,6 +45,7 @@
 import org.kaazing.gateway.resource.address.ws.WsResourceAddress;
 import org.kaazing.gateway.transport.AbstractBridgeSession;
 import org.kaazing.gateway.transport.IoFilterAdapter;
+import org.kaazing.gateway.transport.LoggingFilter;
 import org.kaazing.gateway.transport.ws.AbstractWsBridgeSession;
```
dpwspoon: I don't think you don't need this import

transport/wseb/src/main/java/org/kaazing/gateway/transport/wseb/WsebSession.java
```diff
@@ -1110,6 +1110,12 @@ WsebSession getWsebSession() {
         Logger getLogger() {
             return wsebSession.getLogger();
         }
+
+        @Override
+        public String toString() {
+            return String.format("[wseb#%s transport]", wsebSession.getId());
+
```
dpwspoon: nit extra new line

PR #575, opened 2016-05-16
Add official support to kill on slow handshake

mgherghe:


mgherghe:
@dpwspoon I looked into how to implement this feature for the http layer. I tried to follow the example from the [WsnAcceptor](https://github.com/kaazing/gateway/blob/develop/transport/wsn/src/main/java/org/kaazing/gateway/transport/wsn/WsnAcceptor.java), where a `wsLocalAddress` is created in the `getWsLocalAddress`(line 899). 
I tried to implement this in the [HttpAcceptor](https://github.com/kaazing/gateway/blob/develop/transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpAcceptor.java), firstly in the `addBridgeFilters` method from the `bridgeHandler`, but the `session` object from there doesn't seem to be of the proper type to get the local address as a ResourceAddress object. 
The `doSessionOpened` method (line 300), from the `httpResourcesHandler` seemed to have the same session type as in the `WsnAcceptor`. I placed a breakpoint in that method, but it doesn't seem to stop there when loading the demo page.
What is the best approach to implement this for the http level?

dpwspoon:
@mgherghe I didn't get a chance to look into the code, I'll double check the design tomorrow but it seems like the right start.  All sessions in the Gateway should be IOSessionEx, and it should be safe to caste to that (put an `assert` on it to double check though).  

I believe the HttpResourceHandler (again haven't looked at the code yet) is for what we use to call "http resources", which are dynamic resources that get served, I think things like the session path on the upstream and downstream URLs of WSE...  And thus it probably would not get called when doing a directory service request.

dpwspoon:
@mgherghe For http, one way to solve it would be to add 2 filters in the HttpAcceptor.addBridgeFilters; one to start the timeout, and one after the codec/decoder filter that cancels the timeout.  The HttpAcceptor appears to be a singleton.  In the bindInternal method of the acceptor you will have the HttpResourceAddress with the handshake timeout, and the transportAddress that it binds to.  I would imagine that the transportAddress of the incoming session would be available to bindInternal.  Thus, by keeping a mapping of the transportAddress Addresses to timeouts you should be able to look up timeout in the addBridgeFilters method.

mgherghe:
@dpwspoon I pushed my latest changes based on your suggestions for http. Which would be the best k3po scenario to test this ?

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpAcceptor.java
```diff
@@ -268,6 +276,8 @@ void bindInternal(final ResourceAddress address, IoHandler handler,
         //
         final ResourceAddress transportAddress = address.getTransport();
 
+        handshakeTimeoutByTransportAddress.put(transportAddress, address.getOption(HANDSHAKE_TIMEOUT).longValue());
```
dpwspoon: Because resources addresses at HTTP might bind on the same transport address, the conflict resolution for when HTTP handshake timeouts are set may have to be done here...

dpwspoon:
I would think the minimal requirements would be a http slow handshake where the timeout kicks in.  And a case where it doesn't kick in.  You will probably need to write the slow handshake timeout one in tcp, as there is not a way to partially send the request headers.  For the case that it doesn't kick in we should make sure that the request headers are received then it is done, but the content can be incomplete (double check that this is what was agreed on in the spec)...

mgherghe:
@dpwspoon I added a robot script to test the http handshake timeout filter. If I look in the Eclipse
[log](https://github.com/kaazing/gateway/files/293733/httpHandshakeTimeoutKillsSlowSession.txt) after running the test, I can see that the http filter is activated and closes the session as expected, but the test throws an InterruptedException. I tried putting a try catch in the unit test but it doesn't seem to catch the exception and finish the test properly. Also, after running this test, the k3po robot gets stuck if I try to rerun the test, so it needs to be restarted (probably why many of the tests from the Travis build failed).
What would be the best approach to fix this ?
Attaching also the [k3po log](https://github.com/kaazing/gateway/files/293770/k3po.txt) for this test.

Also, I changed the way to shut down the ScheduledExecutorService from using shutdown() to shutdownNow() method, since the first one took longer to terminate and the task still executed even when it was not supposed to. This change caused the TcpHandshakeTimeoutIT.serverCloseAfterOpenWithNoData to fail. I looked into the k3po script for this test and it seems that it awaits for a notification from a barrier before closing the connection. With the previous approach, this probably happened after reaching the tcp.idle.timeout value, but now this is not the case. Should this test still be kept in the TcpHandshakeTimeoutIT?

transport/http/src/test/java/org/kaazing/gateway/transport/http/HttpHandshakeTimeoutIT.java
```diff
@@ -0,0 +1,94 @@
...
+
+public class HttpHandshakeTimeoutIT {
+
+    private K3poRule k3po = new K3poRule();
+
+    private List<String> expectedPatterns = new ArrayList<>(
+            Arrays.asList(new String[]{"Closing http session .* because handshake timeout of .* milliseconds is exceeded"}));
+
+    private TestRule checkLogMessageRule = new TestRule() {
+        @Override
```
dpwspoon: Is this checkLogMessageRule needed

transport/http/src/test/scripts/org/kaazing/gateway/transport/http/test.http.handshake.timeout.rpt
```diff
@@ -0,0 +1,30 @@
...
+read await BARRIER
+write "GET /clientaccesspolicy.xml HTTP/1.1\r\n"
+write "Accept-Language: en-US\r\n"
+write "User-Agent: Mozilla/5.0 (compatible; MSIE 10.0; Windows NT 6.1; WOW64; Trident/6.0)\r\n"
+write "Host: localhost:8001\r\n"
+write "\r\n"
+
+read "HTTP/1.1 404 Not Found\r\n"
+
+close
```
dpwspoon: Based on the log I think the sever may be closing the connection here, which is sort of expected.  I'm not even sure if a 404 is the correct error code for a slow handshake timeout.  See http://www.restapitutorial.com/httpstatuscodes.html. A 408 seems more appropriate

transport/http/src/test/scripts/org/kaazing/gateway/transport/http/test.http.handshake.timeout.rpt
```diff
@@ -0,0 +1,30 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+connect tcp://localhost:8001
+connected
+
```
dpwspoon: It is unclear to me whether the http layer will kick in right away here.  I would suggest putting the read await BARRIER lower then the first line "GET /clientaccess" line.  And then after that I think there is no need to write out the rest of the headers, as it will race the close from the server

dpwspoon:
>@dpwspoon I added a robot script to test the http handshake timeout filter. If I look in the Eclipse
log after running the test, I can see that the http filter is activated and closes the session as expected, but the test throws an InterruptedException. I tried putting a try catch in the unit test but it doesn't seem to catch the exception and finish the test properly. Also, after running this test, the k3po robot gets stuck if I try to rerun the test, so it needs to be restarted (probably why many of the tests from the Travis build failed).
What would be the best approach to fix this ?
Attaching also the k3po log for this test.

I see several ways the test can race given the way it is written.  Try thinking about the state machine when drafting the tests where both the server under test and robot are two threads.  If there is ever a state where they both will give an action to the other side at the same time then there is a race and you will see failures.  I've added comments to where the more obvious races are.  We can have a call to cover this if that would help?

>Also, I changed the way to shut down the ScheduledExecutorService from using shutdown() to shutdownNow() method, since the first one took longer to terminate and the task still executed even when it was not supposed to. This change caused the TcpHandshakeTimeoutIT.serverCloseAfterOpenWithNoData to fail. I looked into the k3po script for this test and it seems that it awaits for a notification from a barrier before closing the connection. With the previous approach, this probably happened after reaching the tcp.idle.timeout value, but now this is not the case. Should this test still be kept in the TcpHandshakeTimeoutIT?

I don't understand how the change to shutdownNow caused this test to fail.  This is the one test testing the functionality of the handshake time out so it definitely needs to pass.  I don't see anything that seems to be wrong with the tests as it is written (perhaps I am mistaken).  Does the shutdownNow get called when you run this test (I would think that it shouldn't be as the test is not supposed to send any data)?  or is the test sporadically failing?  Either our implementation here is wrong, or our test is wrong

dpwspoon:
It looks like the presence of the barrier is throwing an exception, as the gateway is closing the connection prior to the barrier being triggered.

From: https://github.com/k3po/k3po/wiki/Scripting-Language

> Require that the barrier named “<barrier>” have been notified before the next input event in the script occurs (normally read). If that event occurs before the barrier has been notified this will result in a script failure.

I don't think you need a barrier at all.  The script could be as simple as
```
connect tcp://localhost:8001
connected

write "GET /clientaccesspolicy.xml HTTP/1.1\r\n"
closed
```
And if the server does not close the connection then the closed event will not be hit.

mgherghe:
I changed the k3po script as per your suggestion and now the tests pass. I tried adding 
`read "HTTP1/1 408` at the end of the k3po script, but the gateway doesn't send this, so the test failed in that case.
I was looking into how to implement this feature for the ws layer. I noticed that the WsnAcceptor contains several BridgeHandlers and it is unclear to me where the filters to start and stop the timer for the ws handshake timeout should be added.

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpAcceptor.java
mgherghe: Does this mean that if a certain resource address already exists in the map, then the smallest value should be kept ? 
Also, if there is a case when the user set value is higher than the default value (of 10 seconds), which value should be kept ?

mgherghe:
@dpwspoon I pushed my changes for the wsn layer. 
I added the filters in the `doUpgrade()` method of the `wsnHttpRFC6455BridgeHandler`.
I am not sure if the way the k3po script is written is the proper way to test if a slow wsn connection will be killed, since the test currently fails. 
Your feedback would be helpful.

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/internal/NioHandshakeFilter.java
```diff
@@ -59,8 +59,8 @@ public void onPostAdd(IoFilterChain parent, String name, NextFilter nextFilter)
     }
 
     @Override
-    protected void doSessionOpened(NextFilter nextFilter, IoSessionEx session) throws Exception {
+    protected void doMessageReceived(NextFilter nextFilter, IoSessionEx session, Object message) throws Exception {
         taskExecutor.shutdownNow();
-        nextFilter.sessionOpened(session);
+        nextFilter.messageReceived(session, message);
```
dpwspoon: Is it ok if taskExecutor is called multiple times?  I assume it is...

Consider having this filter remove itself from the filter chain.  Probably not needed, but I'm more just curious

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpAcceptor.java
dpwspoon: Check the requirements doc, But I believe it takes the lowest user defined value.  Meaning if the user defines a value higher then the default then you will to the user defined value.  

I would suggest, not defaulting the value in the resource address, but rather keeping it null.  Then at conflict resolution time you will know whether the user set it... : -)

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpStartHandshakeTimerFilter.java
```diff
@@ -0,0 +1,60 @@
...
+                    session));
+        }
+
+        taskExecutor.schedule(() -> {
+            if (logger.isDebugEnabled()) {
+                logger.debug(String.format(
+                        "Closing http session %s because handshake timeout of %d milliseconds is exceeded", session,
+                        handshakeTimeout));
+            }
+            session.close(true);
```
dpwspoon: before session.close maybe you can write the 408 error message?

dpwspoon:
If you think of wsn, the handshake is really just at the http layer, so I would expect in the code that the future is being completed nearly instantaneously? So there probably isn't much to do...

For wse, again it is http.

For these two, we should consider propagating the ws handshake timeout to http if it isn't set... Thought? Ping @robinzimmermann 

For wsx the handshake is the second request response, see:

https://github.com/k3po/k3po/blob/develop/specification/httpx/image00.png

Also, take a look at TLS :-).

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpAcceptor.java
mgherghe: And where should the default value be set then? In case there is no user defined value in the config.

mgherghe:
>For these two, we should consider propagating the ws handshake timeout to http if it isn't set... 

In the requirements document, it is stated that all the layers should have a default value. This would mean that there isn't a case where the http isn't set.

mgherghe:
@dpwspoon While trying to start the gateway locally, I noticed that I get a NPE when trying to connect to the configured echo service. 

>2016-06-14 14:28:52,200 [New I/O worker #1] INFO  [tcp#1 0:0:0:0:0:0:0:1:53661] EXCEPTION: java.lang.NullPointerException
java.lang.NullPointerException
	at org.kaazing.gateway.transport.http.HttpAcceptor.addBridgeFilters(HttpAcceptor.java:549)

After debugging, I saw that when starting the gateway, the value for the `transportAddress` in `HttpAcceptor` is `[tcp://[0:0:0:0:0:0:0:1]:8000 (tcp://localhost:8000) http/1.1]` 
as opposed to the value existing in the map, which is
`[tcp://127.0.0.1:8000 (tcp://localhost:8000) http/1.1]`
`[tcp://[0:0:0:0:0:0:0:1]:8000 (tcp://localhost:8000) http/1.1]`.
I could not figure out why this value is different when starting the gateway, but it works fine with the existing k3po scripts. Your feedback would be helpful.

dpwspoon:
It is likely that k3po is not using the ipv6 address to connect, perhaps your test is.   I'm not sure why the address is different.  Perhaps the LocalAddress is not being correctly converted to a ResourceAddress...?  Maybe at the tcp layer it might be better to deal with InetAddresses as the key to the map anyways because there is that inherit casting... (try using externalURI)

mgherghe:
@dpwspoon The value with both ipv4 and ipv6 is from when the k3po script is run. 
I am getting the `transportAddress` with `ResourceAddress transportAddress = LOCAL_ADDRESS.get(session);`
However, when I am trying to connect to a service after starting the gateway, this value seems to be the `alternate` value from the addresses used in the `bindInternal` method, that are kept in the map. 
When running a test with a k3po script, the value for the `transportAddress` is not the alternate..
One option to not get a NPE when trying to connect to the gateway would be to keep the alternate value in the map and also use it when trying to add the filters, but I am not sure this is the best approach.
I looked into using the `externalUri`, but if I have several accepts with just different paths, they would have the same value for the `externalUri`. 

I also ran the `HttpHandshakeTimeoutIT.httpHandshakeTimeoutKillsSlowSession` test in debug mode in its current format, and even though the test passes, it does not add any handshake filters (not the TCP one, neither the http ones). When it had a barrier in the k3po script, it added the filters properly.

dpwspoon:
>  if I have several accepts with just different paths, they would have the same value for the externalUri.

There is no such thing as paths at tcp level, so I think this should be ok... There is only ip and port.  Did you mean different domain names?  If so, the domain names resolving to different IP addresses should be ok

I'm also ok with you populating the map with all the alternates,  that would work as well

> I also ran the HttpHandshakeTimeoutIT.httpHandshakeTimeoutKillsSlowSession test in debug mode in its current format, and even though the test passes, it does not add any handshake filters (not the TCP one, neither the http ones). When it had a barrier in the k3po script, it added the filters properly.

I'm not sure how that could happen?  As I don't know what would kill the session?  Did you wire shark it, if there is no response from HTTP then the filter must of kicked in.  If it didn't can you debug the path it went down to see why it didn't kick in?

mgherghe:
Relevant quote from the conversation with David on Slack:
> The idea with the conflict resolution was that at configuration time/startup you know what values were user set, but you don’t have a way to retain that information through the resource address to where they are actually being used in the transports.  To get around this, you could convey the information of whether they were user set or not by defaulting the values until latter in the process when the transport is being setup, and instead keeping them as null if they are explicitly set in the config . 
 Also, I would suggest we initially introduce this as a Feature switch in the first iterations,  See the EarlyAccessFeatures classes Chris as added.

mgherghe:
I did some more debugging with the `WsnHandshakeTimeoutIT`, set a conditional breakpoint in `HttpAcceptor`, line 280, with the condition: 
`address.getExternalURI().equals("http://localhost:8001/")`. 
I am attaching a [print screen](https://cloud.githubusercontent.com/assets/11552767/16405856/6016669a-3d13-11e6-889e-0c58d370d943.png) with the comparison between the stack traces for the resource address with the value: 
`[http://localhost:8001/ (http://localhost:8001/) ws/rfc6455]
 [tcp://127.0.0.1:8001 (tcp://localhost:8001) http/1.1]
 [tcp://[0:0:0:0:0:0:0:1]:8001 (tcp://localhost:8001) http/1.1]`

In the case from the left, the handshake.timeout value is correctly set to 6000 millis but in the case from the right it is overridden with the default value of 10000.
It is not clear to me why this happens. 
I am assuming that if I configure the default value as you suggested above, this would not be a problem anymore. I will do this now, but I wanted to understand why the value is being changed.
Any information on this would be most useful.

dpwspoon:
Take a look at the call stack, and set an expression for `address.getOption(HANDSHAKE_TIMEOUT).longValue()`.  You can see that the 10 is being set when ws is binding http which is binging ws which is binding http.  This is the WSX case and the values for ws are perhaps not getting passed between layers.

I'm not sure if the alternates will work for this case or not, or if the resources addresses with the comparators will work in the map.

image

mgherghe:
@dpwspoon I committed my changes with the conflict resolution for the http layer.
For the moment, I ignored the `WsnHandshakeTimeoutIT` test, since if I use the [test.wsn.handshake.timeout](https://github.com/mgherghe/gateway/blob/protocol_handshake_timeout/transport/wsn/src/test/scripts/org/kaazing/gateway/transport/wsn/test.wsn.handshake.timeout.rpt) as it is now, the test fails (attaching [log](https://github.com/kaazing/gateway/files/341804/fail.txt)).
If I remove the line with 'closed' from the k3po script, the test passes, but it throws some exceptions and the sessions are not closed by the handshake filters. Attaching [log](https://github.com/kaazing/gateway/files/341815/success.txt).
What would be the best approach to fix this ?

dpwspoon:
I don't see any byte traffic on the wireshark going, nor in the gateway log.  It's interesting that the following is in the log
```
1092 transport.http.accept [New I/O worker #1] TRACE accept  - Setting handshake timeout of 6000 milliseconds on http session (#00000001: kaazing niosocketchannel, server, /127.0.0.1:54644 => /127.0.0.1:8001) 
```

What I think is happening is 1) that you don't have an `write close` in the http script.  This is the equivilant to a flush.  So the http headers never make it on the wire.  2) Because http is the only thing bound on the server socket we can route it immediately to the http layer, that is printing the log message above, I suspect if an ambiguous protocol (perhaps another http binding, but for sure TLS) was bound on the same tls port, then the empty tcp connection would not get routed to the http layer and this log message would not appear.  3) The TCP connection is then closed by the http layers timeout, and the script returns saying it never really completed the writes.

I would suggest a fix would be to write the request in tcp robot script.

Also, a positive test to show that it doesn't kick in when websocket handshake completed should be added ;-)

dpwspoon:
> If I remove the line with 'closed' from the k3po script, the test passes, but it throws some exceptions and the sessions are not closed by the handshake filters. Attaching log.

When I remove the closed the test passes.  Because the test passes the log is not written to the console.  How did you get the log?

Also, what you are seeing in the log doesn't match up with what I'm seeing on the wire from wire shark as there isn't a 404 so something odd it going on here.  Hopefully the above response is helpful enough.

mgherghe:
@dpwspoon 
As per your suggestion, I changed the robot script to connect to TCP and this fixed the problem. 

> When I remove the closed the test passes. Because the test passes the log is not written to the console. How did you get the log?

For testing purposes, I modified the `MethodExecutionTrace.succeeded()` method and added the line: `MemoryAppender.printAllMessages();` 
but I did not commit this change.

mgherghe:
@dpwspoon I had a question regarding your suggestion to initially introduce this as a Feature switch. The filters for killing the slow session handshakes are added on several layers and in different methods that are called several times throughout the functioning of the gateway. One approach would be to check if the feature is enabled within each layer, but that would imply too many checks per layer. Do you have any suggestion for a better approach ?

dpwspoon:
Does a handshake Timeout of 0 mean there is no timeout value (I assume we need something like this).  If this is not in the requirements doc we should add it.  If this is the case one option would be to set all the values to 0 at start up (See Accept / Connect Options Context).  Another option would be to do it at each layer where the filter gets added.

mgherghe:
@dpwspoon I implemented this as an Early Access feature and pushed my changes. I also solved some merging conflicts with the gateway's develop branch. 

I also added the filters for closing slow wsx handshakes. I saw that the `WsxExtensionsNegotiationIT` uses the `should.negotiate.extensions.in.wrapped.request` script to test the wsx handshake. I think I should use a similar approach, but I'm not sure exactly what the http request that is wrapped in a ws frame should look like.

dpwspoon:
This [spec describes httpx](https://github.com/k3po/k3po/blob/develop/specification/httpx/SPEC.md) and has examples of how httpx looks.  httpx is wsx essentially, and shows the wrapped response

management/src/main/java/org/kaazing/gateway/management/config/ServiceConfigurationBeanImpl.java
```diff
@@ -189,6 +189,26 @@ public String getAcceptOptions() {
                         jsonOptions.put("http.transport", httpTransport);
                     }
 
+                    Long tcpHandshakeTimeout = (Long) acceptOptions.remove("tcp.handshake.timeout");
+                    if (tcpHandshakeTimeout != null) {
+                        jsonOptions.put("tcp.handshake.timeout", tcpHandshakeTimeout);
+                    }
+
+                    Long sslHandshakeTimeout = (Long) acceptOptions.remove("ssl.handshake.timeout");
```
dpwspoon: please accept ssl.handshake.timeout or tls.handshake.timeout for this value

management/src/main/java/org/kaazing/gateway/management/config/ServiceConfigurationBeanImpl.java
```diff
@@ -326,6 +346,26 @@ public String getConnectOptions() {
...
                     }
 
+                    Long tcpHandshakeTimeout = (Long) connectOptions.remove("tcp.handshake.timeout");
+                    if (tcpHandshakeTimeout != null) {
+                        jsonOptions.put("tcp.handshake.timeout", tcpHandshakeTimeout);
+                    }
+
+                    Long sslHandshakeTimeout = (Long) connectOptions.remove("ssl.handshake.timeout");
+                    if (sslHandshakeTimeout != null) {
+                        jsonOptions.put("ssl.handshake.timeout", sslHandshakeTimeout);
```
dpwspoon: ditto

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultAcceptOptionsContext.java
```diff
@@ -92,6 +99,21 @@ public DefaultAcceptOptionsContext(ServiceAcceptOptionsType acceptOptions, Servi
         this.options = Collections.unmodifiableMap(options);
     }
 
+    private void checkHandshakeTimeoutEnabled(Map<String, String> options) {
+        if (options.containsKey("tcp.handshake.timeout") || options.containsKey("ssl.handshake.timeout")
```
dpwspoon: consider using constants for these strings

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpAcceptor.java
```diff
@@ -269,6 +280,19 @@ void bindInternal(final ResourceAddress address, IoHandler handler,
         //
         final ResourceAddress transportAddress = address.getTransport();
 
+        final Long handshakeTimeout = address.getOption(HANDSHAKE_TIMEOUT);
+        if (handshakeTimeout != null) {
+            Long newMappedValue = handshakeTimeoutByTransportAddress.computeIfPresent(transportAddress, (key, value) -> {
+                Long handshakeTimeoutValue = Math.min(handshakeTimeout, value);
```
dpwspoon: consider using static imports on your Math call

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpStartHandshakeTimerFilter.java
```diff
@@ -0,0 +1,60 @@
...
+    private final Long handshakeTimeout;
+    private ScheduledExecutorService taskExecutor;
+
+    public HttpStartHandshakeTimerFilter(Logger logger, Long handshakeTimeout, ScheduledExecutorService taskExecutor) {
+        this.logger = logger;
+        this.handshakeTimeout = handshakeTimeout;
+        this.taskExecutor = taskExecutor;
+    }
+
+    @Override
```
dpwspoon: I expected this to be on sessionedOpened, but this might be okay.  Is there a reason you did it on PostAdd, and when is PostAdd called?

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpAcceptor.java
```diff
@@ -530,9 +559,19 @@ private void fireContentReceived(DefaultHttpSession session, HttpContentMessage
 
     @Override
     public void addBridgeFilters(IoFilterChain chain) {
-        IoSession transport = chain.getSession();
+        IoSession session = chain.getSession();
+
+        SocketAddress localAddress = session.getLocalAddress();
+
+        ResourceAddress transportAddress = LOCAL_ADDRESS.get(session);
+        ScheduledExecutorService taskExecutor = Executors.newScheduledThreadPool(1);
```
dpwspoon: I don't think we want to instantiate a ScheduledExecutorService for every session.  Instead get it via `@Resource` annotation for the ScheduleProvider

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpStartHandshakeTimerFilter.java
```diff
@@ -0,0 +1,60 @@
...
+    public void onPostAdd(IoFilterChain parent, String name, NextFilter nextFilter) throws Exception {
+        IoSession session = parent.getSession();
+        if (logger.isTraceEnabled()) {
+            logger.trace(String.format("Setting handshake timeout of %d milliseconds on http session %s ", handshakeTimeout,
+                    session));
+        }
+
+        taskExecutor.schedule(() -> {
```
dpwspoon: you can keep track of the scheduled task by adding it as a key to the HttpSession.class.  This way you don't have to pass the executor around in the constructor and between other filters.

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpStopHandshakeTimerFilter.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.kaazing.gateway.transport.IoFilterAdapter;
+import org.kaazing.mina.core.session.IoSessionEx;
+
+/**
+ * Closes the connection if the http handshake is not successful in the given amount of time
+ *
+ */
+public class HttpStopHandshakeTimerFilter extends IoFilterAdapter<IoSessionEx> {
+
```
dpwspoon: If the other filter is being called on postAdd and this is hitting on sessionOpened, then I would expect they could be combined into on filter with the logic moving from above into here

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/internal/NioIdleFilter.java
```diff
@@ -1,63 +0,0 @@
-/**
```
dpwspoon: We will need these changes in until you are feature complete for your whole story.... Sorry

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/internal/NioHandshakeFilter.java
```diff
@@ -0,0 +1,69 @@
...
+public class NioHandshakeFilter extends IoFilterAdapter<IoSessionEx> {
+
+    private final Logger logger;
+    private final Long handshakeTimeout;
+    private ScheduledExecutorService taskExecutor;
+
+    public NioHandshakeFilter(Logger logger, Long handshakeTimeout) {
+        this.logger = logger;
+        this.handshakeTimeout = handshakeTimeout;
+        taskExecutor = Executors.newScheduledThreadPool(1);
```
dpwspoon: I don't think we will want to create a thread pool per session.  This would have high overhead, see `@Resource` ScheduleProvider to get the one we create for these type of tasks

transport/wsn/src/main/java/org/kaazing/gateway/transport/wsn/WsnAcceptor.java
```diff
@@ -1092,6 +1093,14 @@ private void doUpgrade(final HttpAcceptSession session) throws Exception {
 
                     final ResourceAddress wsLocalAddress = localAddress;
 
+                    ScheduledExecutorService taskExecutor = Executors.newScheduledThreadPool(1);
+
+                    Long handshakeTimeout = 10000L;//localAddress.getOption(HANDSHAKE_TIMEOUT).longValue();
+                    if (handshakeTimeout != null && handshakeTimeout > 0) {
+                        session.getFilterChain().addLast("wsnStartHandshakeTimerFilter",
+                                new WsxStartHandshakeTimerFilter(logger, handshakeTimeout, taskExecutor));
```
dpwspoon: Same here, don't create the taskExecutor, you can then pass the scheduled task around as a session key on the WsxSession

transport/wsn/src/main/java/org/kaazing/gateway/transport/wsn/WsnAcceptor.java
```diff
@@ -1092,6 +1093,14 @@ private void doUpgrade(final HttpAcceptSession session) throws Exception {
 
                     final ResourceAddress wsLocalAddress = localAddress;
 
+                    ScheduledExecutorService taskExecutor = Executors.newScheduledThreadPool(1);
+
+                    Long handshakeTimeout = 10000L;//localAddress.getOption(HANDSHAKE_TIMEOUT).longValue();
+                    if (handshakeTimeout != null && handshakeTimeout > 0) {
+                        session.getFilterChain().addLast("wsnStartHandshakeTimerFilter",
```
dpwspoon: nit, wsn in name is wrong here

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpStartHandshakeTimerFilter.java
mgherghe: The PostAdd method is "invoked after the filter is added to the specified parent" (from the JavaDoc). I implemented this in the PostAdd method because the timer for the handshake should be initialized before the session is created.

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpStartHandshakeTimerFilter.java
mgherghe: What do you mean by `adding it as a key to the HttpSession.class`? Could you provide more details on this or an existing example from the code ?

management/src/main/java/org/kaazing/gateway/management/config/ServiceConfigurationBeanImpl.java
mgherghe: I was thinking of removing the ssl changes for now, since I split the initial story in several stories, and there is a separate story for the ssl implementation.

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpStartHandshakeTimerFilter.java
dpwspoon: You can add it as an attribute to the session.  The attributes are key values, thus the key should be a constant.  Typically we keep those on the Session classes or the Resource addresses.  Here I think it makes more sense to put it on the session

`session.getAttributes().put(HttpSession.HANDSHAKE_TIMEOUT_TASK, timeoutTast)` or something like that

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpStartHandshakeTimerFilter.java
dpwspoon: So in this case it might be ok.  But the filters can be added to the change anytime during the lifecycle including prior to the session even starting (minutes prior) or after a session has been opened.  So tying it to the lifecycle of the session make more sense as you want the timer to start when you know the session is starting to go up the pipeline

PR #587, opened 2016-05-26
Adding TLSv1.2 ciphers that are enabled by java 8 sun provider 

jitsni:


util/pom.xml
```diff
@@ -88,6 +87,7 @@
                     </dependency>
                 </dependencies>
             </plugin>
+-->
```
jitsni: It is commented out now. We need to create a story to upgrade checkstyle that understands java 8 syntax

PR #588, opened 2016-05-27
fixed travis file bug that wasn't building all branches

AdrianCozma:


AdrianCozma:
@dpwspoon I found this issue while looking at the Travis logs to investigate the sonarqube ticket, I've made/updated all the projects that will be in SonarQube but I think we should merge the after https://github.com/kaazing/tickets/issues/523 is fixed and the public artifactory released, this way we can check the sonar analysis works properly from Travis

PR #594, opened 2016-06-03
Fix for issue 466

msalavastru:
This is a PR to fix the [issue 466](https://github.com/kaazing/tickets/issues/466).
I moved the PR here from 4.x ([4.x PR](https://github.com/kaazing-private/gateway.server/pull/137)).
I no longer get OOM, but I am still looking into why the gateway suspends the reads on the sessions without resuming.

### DIFF
--- a/service/proxy/src/main/java/org/kaazing/gateway/service/proxy/AbstractProxyHandler.java
+++ b/service/proxy/src/main/java/org/kaazing/gateway/service/proxy/AbstractProxyHandler.java
@@ -20,6 +20,7 @@ import static java.lang.String.format;
 import java.io.IOException;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -33,6 +34,7 @@ import org.apache.mina.core.write.WriteRequest;
 import org.kaazing.gateway.service.ServiceContext;
 import org.kaazing.mina.core.buffer.IoBufferAllocatorEx;
 import org.kaazing.mina.core.buffer.IoBufferEx;
+import org.kaazing.mina.core.session.IoSessionEx;
 import org.kaazing.mina.filter.util.WriteRequestFilterEx;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


jitsni:
I am not convinced about this change. Here the WriteFuture is completed so it means it has written to session. Are you saying that WriteFuture is fired without actually written (but queued) to the underlying layer ?

@cmebarrow @jfallows do you have any comments on this PR ?

daniel-smith-kaazing:
Please see my comments in the original ticket #466. This does not reproduce with Gateway 5.0.1 as released.

cmebarrow:
@jitsni My understanding of the change is that it is intended to handle a case where the IoSessions on each side of the attached session manager are running on different I/O threads. @msalavastru Is that understanding correct?

msalavastru:
@cmebarrow Yes you are correct. The two sessions are almost always running in different threads, and because of this the read/suspend read logic sometimes has problems (as explained [here](https://github.com/kaazing-private/gateway.server/pull/137#issuecomment-222491679)). But even if I realign the threads when the future fires operationComplete, after some time the sourceSession stops reading, throttling the client down to 0Kb/sec. It looks like at some point the queued messages are written just above the resume read threshold (32000), not going below and therefore not resuming the read anymore.
Scenario is like this:
- write message, scheduledWriteBytes remaining : 36864, messages remaining 17
- write message, scheduledWriteBytes remaining : 34816, messages remaining 16
- write message, scheduledWriteBytes remaining : 32768, messages remaining 15
- in normal operation it would right one more message, going down to 30720 < 32000 and resume the read, but after some time this does not happen. I do not see any stuck threads...

Looking into this. @jitsni Any ideeas what might cause this?

jitsni:
```
int newScheduledWriteBytes = scheduledWriteBytes.addAndGet(-bytesWritten);
```
The above has nothing to do with alignment. So I would think this should be done regardless of the sessions aligned or not. The message is written and just decrement it first in the operationComplete().

There seems to be complicated logic which probably not necessary. There is a very rare chance that the two while loops may not terminate (not saying that is what happening)
```
               while (readSuspended.compareAndSet(false, true)) {
                    sourceSession.suspendRead();
                }
```
```
                        while (readSuspended.compareAndSet(true, false)) {
                            sourceSession.resumeRead();
                        }
```

I think you can some more logging messages so that you can see why draining is not happening (due to some conditional logic). Can you add readSuspended state to the logging messages.

jitsni:
@msalavastru can you try with my branch https://github.com/jitsni/gateway/tree/proxy.handler.oom. If this fixes then we can come up with a better concurrency construct to suspend and resume source session.

daniel-smith-kaazing:
@msalavastru @jitsni @cmebarrow - I could not reproduce this. I attempted at @sbadugu 's request starting Friday 6/3/2015 using EE Gateway 5.0.1 as released. After 4 days running the client at nearly 2 gbps with -Xmx512m. You can see detailed notes in #466. He asked me that to try to determine if a fix was worthwhile. It doesn't seem to be. I was a bit surprised everyone is still working on it.

sbadugu:
@msalavastru We will close this PR for now and revisit later if we reproduce the issue again.

PR #598, opened 2016-06-07
Fix for issue 80:  Multiple Log4J in distribution

msalavastru:


dpwspoon:
See comments on ticket

PR #603, opened 2016-06-18
Updated mina.netty version to resolve security scan (note: we don't a…

dpwspoon:
…ctually use the class exposed in the security bug)

### DIFF
--- a/mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioSelector.java
+++ b/mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioSelector.java
@@ -90,6 +90,11 @@ abstract class AbstractNioSelector implements NioSelector {
     protected volatile Thread thread;
 
     /**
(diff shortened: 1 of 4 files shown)


mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioSelector.java
```diff
@@ -90,6 +90,11 @@
     protected volatile Thread thread;
 
     /**
+     * Count down to 0 when the I/O thread starts and {@link #thread} is set to non-null.
+     */
+    final CountDownLatch startupLatch = new CountDownLatch(1);
```
jitsni: do we need this change ?

dpwspoon: Yes, Netty was accessing this field directly from other classes.  This is a code diff from updating the new version

PR #606, opened 2016-06-23
Fix for issue #518: Http proxy is adding its own unauthorized content to http response that are unauthorized

msalavastru:
This is a fix for [issue 518](https://github.com/kaazing/tickets/issues/518).

### DIFF
--- a/service/http.proxy/src/test/java/org/kaazing/gateway/service/http/proxy/HttpProxyContentFilterIT.java
+++ b/service/http.proxy/src/test/java/org/kaazing/gateway/service/http/proxy/HttpProxyContentFilterIT.java
@@ -1,0 +1,61 @@
+/**
+ * Copyright 2007-2016, Kaazing Corporation. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


msalavastru:
@jitsni Looks like this change is causing problems for tests such as [HttpProxyRequestIT#getMethodStatus401](https://github.com/msalavastru/gateway/blob/develop/service/http.proxy/src/test/scripts/org/kaazing/gateway/service/http/proxy/http.proxy.get.method.status.401.rpt) that do not send Content-Length at all, but send chunked content. The test fails because it receives injected content.

It is not clear to me when do we want to inject content. 

- if Content-Length is present do not inject
- if Content-Length is not present, we probably should not inject also, right?

jitsni:
Looks good. I would reorganize something like the following:

```
        HttpStatus httpStatus = httpResponse.getStatus();
        boolean noContent = content == null || (content.length() == 0 && content.isComplete());

        if (contentLength == null && contentAutomaticallyInjectable(httpStatus) && !httpResponse.isContentExcluded() &&
                noContent) {
            IoBufferAllocatorEx<?> allocator = session.getBufferAllocator();
        ...
```

When there is Content-Length header, just don't inject any new content (it doesn't matter whether HttpContent is null or not).

msalavastru:
@jitsni Thanks for your suggestion, I have updated the PR with these changes.

jitsni:
Good work. Pushing it. @dpwspoon you should test with this fix.

PR #610, opened 2016-06-24
Gateway Code Cleanup

vishalsatish:
<ul>
<li>used SonarLint Eclipse plugin to identify issues</li>
<li>see kaazing/roadmap#780</li>
</ul>


dpwspoon:
Let's break this down in to smaller PRs that we can tackle piecemeal.  

Some notes:

- Let's keep the xml files with 2 spaces / not reformat them.  Please add check to checkstyle plugin if this is occurring/sonar lint.

- The TODO documentation, there is no need to add this as this is a general todo and exposed APIs are currently in separate modules

PR #604, opened 2016-06-21
Login Module passes when there is a realm and login module

justinma246:
Login module passes without require role when there is a realm. If there is a require role when there exists a realm, the authorization constraint is removed. Also, namespace of files is updated to 2016/06.
(Many files are edited because the imports had to be changed from nov2015 to june2016)

Ticket: #825 (https://github.com/kaazing/roadmap/issues/825)

PR #628, opened 2016-07-11
Login Module fix

justinma246:
https://github.com/kaazing/gateway/pull/604


justinma246:
@jitsni

jitsni:
travis build is failing and seems to be related this changeset.

justinma246:
@jitsni I made some changes, thanks for the heads up!

server/src/main/java/org/kaazing/gateway/server/config/parse/translate/June2016Validator.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.kaazing.gateway.server.config.parse.translate.june2016.RemoveRequireUser;
+
+public class June2016Validator extends GatewayConfigTranslatorPipeline {
+
+    public June2016Validator() {
+        super();
+        
+        // Set the namespace
+        addTranslator(new RemoveRequireUser());
```
dpwspoon: I don't get this design reason... Why are removing require user in the 2016 Validator

dpwspoon: We should be removing the realm on the old namespace only if there is a realm and no authorization constraint.  Meaning the translator is wrong and this should be added to a different Translator Pipeline

PR #631, opened 2016-07-14
Pokemon exceptions

DoruM:
Qualified exceptions in gateway configuration parser.
Cleaned up exception code.


dpwspoon:
This generally looks good to me.  @DoruM the branch needs to be caught up.  

@jitsni are you okay with the multiple exception types being thrown for gateway launch?

server/src/main/java/org/kaazing/gateway/server/config/parse/GatewayConfigNamespace.java
```diff
@@ -111,12 +105,8 @@ public String toURI() {
                 uri = NS_SEPTEMBER_2014_URI;
                 break;
 
-            case NOVEMBER_2015:
-                uri = NS_NOVEMBER_2015_URI;
-                break;
-                
             case CURRENT_NS:
-                uri = NS_JUNE_2016_URI;
+                uri = NS_NOVEMBER_2015_URI;
```
jitsni: 2016 is the current namespace, right ?

server/src/main/java/org/kaazing/gateway/server/Launcher.java
```diff
@@ -45,18 +49,19 @@ public Launcher(GatewayObserver gatewayListener) {
         this.gatewayListener = gatewayListener;
     }
 
-    public void init(GatewayContext context) throws Exception {
+    public void init(GatewayContext context) throws URISyntaxException , IOException  {
```
jitsni: I don't think I want these specific exceptions in the contract. May be LauncherException a runtime exception with a specific cause

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/auth/BasicLoginModuleTestRoleIT.java
```diff
@@ -1,83 +0,0 @@
...
-import org.junit.Rule;
-import org.junit.Test;
-import org.junit.rules.TestRule;
-import org.kaazing.gateway.server.test.GatewayRule;
-import org.kaazing.gateway.server.test.config.GatewayConfiguration;
-import org.kaazing.gateway.server.test.config.builder.GatewayConfigurationBuilder;
-import org.kaazing.k3po.junit.annotation.Specification;
-import org.kaazing.k3po.junit.rules.K3poRule;
-
-public class BasicLoginModuleTestRoleIT {
```
jitsni: why is this test deleted ?

service/spi/src/main/java/org/kaazing/gateway/service/ServiceContext.java
```diff
@@ -75,9 +79,9 @@
 
     File getTempDirectory();
 
-    void init() throws Exception;
+    void init() ;
 
-    void start() throws Exception;
+    void start() throws MalformedURLException, RemoteException, URISyntaxException, IOException;
```
jitsni: same feedback as above. Also you will break enterprise.gateway or itests if they have a dependency on this file

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/auth/BasicLoginModuleTestRoleIT.java
DoruM: Hi,

This was deleted by @dpwspoon in [this commit](https://github.com/kaazing/gateway/pull/631/commits/e00d24bf9cc6b123ec0b6776d473297e6978d786) with the comment: 
Revert "Login Module fix"

Regards,
Doru

server/src/main/java/org/kaazing/gateway/server/config/parse/GatewayConfigNamespace.java
DoruM: Hi,

This was deleted by @dpwspoon in this commit with the comment: 
Revert "Login Module fix". I have simply included his change. I assume that he reverted all the changes done by that module. 

Regards,
Doru

server/src/main/java/org/kaazing/gateway/server/Launcher.java
DoruM: I think you are correct. I will incorporate this fix.

dpwspoon:
Apologies.  I merged this to soon.  Reverting.

DoruM:
Hi David, Jitendra,

I will resume here the comments by Jitendra:

1. Changes are not compatible with enterprise.gateway
2. A more general exception, like LauncherException should be implemented

I have a few problems regarding the "enterprise.gateway" merge. (Item 1 above)

"Breaking the enterprise" edition is not an issue, or is an issue bigger than this particular merge.  In truth, the current "enterprise.gateway" develop branch is not compatible with the "gateway" develop branch. 

In reality, enterprise.gateway uses release 5.1.1, and not the develop snapshot, as I believe it would be normal. As such, for example, the enterprise code references classes like org.kaazing.gateway.server.config.parse.translate.November2015Validator which are no longer used. So it won't be affected.

This to me seems like a serious issue and there should be a separate story or ticket which would address this specific kind of merge. 

I will add the LauncherException (item 2) required by @jitsni but I will not do any changes to the enterprise.gateway. 

Regards,
Doru

@jitsni , @dpwspoon

service/spi/src/main/java/org/kaazing/gateway/service/ServiceContext.java
DoruM: Not quite. See comment at the end of the thread.

PR #635, opened 2016-07-18
Checking if 'Connection' header is present when Gateway connects as a client

DoruM:
Up to now, if connection header was not present, the connection would still be established.

Additionally, I used HttpHeaders.HEADER_UPGRADE and HttpHeaders.HEADER_CONNECTION instead of using strings.


transport/wsn/src/main/java/org/kaazing/gateway/transport/wsn/WsnConnector.java
```diff
@@ -527,7 +528,7 @@ public void operationComplete(CloseFuture future) {
         }
 
         private void doUpgrade(final HttpConnectSession httpSession) {
-            String upgradeHeader = httpSession.getReadHeader("Upgrade");
+            String upgradeHeader = httpSession.getReadHeader(HttpHeaders.HEADER_UPGRADE);
```
dpwspoon: Consider using static imports in the future

PR #639, opened 2016-07-19
Hazel cast upgrade

mjolie:


management/src/test/java/org/kaazing/gateway/management/jmx/JmxRoundTripLatencyIT.java
```diff
@@ -39,7 +39,7 @@
 import org.kaazing.gateway.server.test.config.builder.GatewayConfigurationBuilder;
 import org.kaazing.k3po.junit.annotation.Specification;
 import org.kaazing.k3po.junit.rules.K3poRule;
-
```
dpwspoon: Why is this ignored.  I expect it should work...

management/src/test/java/org/kaazing/gateway/management/jmx/JmxSessionPrincipalIT.java
```diff
@@ -125,6 +125,7 @@
         "wse.session.with.user.principal.joe",
         "wsn.session.with.user.principal.ann" })
     @Test
+    @Ignore
     public void shouldKillSessionsByUserPrincipal() throws Exception {
```
dpwspoon: same question

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultClusterContext.java
```diff
@@ -125,6 +125,12 @@ public DefaultClusterContext(String name,
         this.clusterMembers.addAll(members);
         this.schedulerProvider = schedulerProvider;
         this.connectOptions = connectOptions;
+
+        if (!GL.isTraceEnabled(GL.CLUSTER_LOGGER_NAME)) {
+            // Note: must use Logger.getLogger, not LogManager.getLogger
```
dpwspoon: Will this affect the logging that users of our cluster feature expect.  Perhaps we can reflect the same log level instead of turn on vs off...?

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultClusterContext.java
```diff
@@ -313,7 +313,7 @@ private Config initializeHazelcastConfig() throws Exception {
             // Explicitly enable the interface so that Hazelcast will pick the one specified
             if (!useAnyAddress) {
                 networkConfig.getInterfaces().setEnabled(true);
-                hazelCastConfig.setProperty(GroupProperties.PROP_SOCKET_BIND_ANY, "false");
+                hazelCastConfig.setProperty("hazelcast.socket.bind.any", "false");
```
dpwspoon: suggest pulling out to constant?

server/src/test/java/org/kaazing/gateway/server/cluster/StandaloneClusterContextTest.java
```diff
@@ -28,106 +31,157 @@
...
         imap.put("test1", value1);
-        assertEquals(listener1.addedCount, 1);
-        assertEquals(listener1.udpatedCount, 0);
-        assertEquals(listener1.removedCount, 0);
-        assertEquals(listener2.addedCount, 1);
-        assertEquals(listener2.udpatedCount, 0);
-        assertEquals(listener2.removedCount, 0);
-        assertEquals(imap.get("test1"), value1);
+        try {
+            Thread.sleep(1000);  //pause long enough for listeners to receive events
```
dpwspoon: Sleeps are bad, especially when we automate the heck out of testing and run on all the different platforms, some of which will have very low CPU.  Is there a way to do this with a latch that is notified when the value gets set, or something similar.  And then set a timeout for the test for the case the latch is never called

management/src/test/java/org/kaazing/gateway/management/jmx/JmxRoundTripLatencyIT.java
mjolie: These test are consistently failing for me on windows with the same error mentioned here:   @Ignore("https://github.com/kaazing/tickets/issues/448")
    public void shouldKillSessionsByRolePrincipal() throws Exception 
I will revert the ignore as see if they pass on unix.

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultClusterContext.java
mjolie: we were always disabling this in the past with no way of enabling it without making a code change, I wanted to make it so it is enabled as part of HA trace.

server/src/test/java/org/kaazing/gateway/server/cluster/StandaloneClusterContextTest.java
mjolie: Yes, I can use a latch here, will make change

server/src/test/java/org/kaazing/gateway/server/cluster/StandaloneClusterContextTest.java
```diff
@@ -20,13 +20,19 @@
...
+import org.junit.Test;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
+import org.kaazing.gateway.server.context.resolve.StandaloneClusterContext;
+
 public class StandaloneClusterContextTest {
+    
+  
+       CountDownLatch entryAdded, entryRemoved, entryUpdated, entryEvicted;
```
dpwspoon: nit formatting

dpwspoon: nit not aligned with Test

Issue #254, opened 2015-07-29
WSN Transport Bug : shouldEchoBinaryFrameWithPayloadLength0 and shouldEchoTextFrameWithPayloadLength0


chadpowers:
Test shouldEchoBinaryFrameWithPayloadLength0 uses [this K3PO script](https://github.com/k3po/k3po/blob/develop/specification/ws/src/main/scripts/org/kaazing/specification/ws/framing/echo.binary.payload.length.0/handshake.request.and.frame.rpt).

Test shouldEchoTextFrameWithPayloadLength0 uses [this K3PO script](https://github.com/k3po/k3po/blob/develop/specification/ws/src/main/scripts/org/kaazing/specification/ws/framing/echo.text.payload.length.0/handshake.request.and.frame.rpt)

Tests are added in gateway pull #236 

BaseFramingIT.java & `@Ignore` has this issue's title.

Error is that script is expecting 'read [0x82 0x00]', actually getting 'Disconnected'. The GATEWAY is not echoing 0 length payload.

chadpowers:
This issue will address many ignores in pull #236, besides just these two. 4 tests in FragmentationIT.java also fail because the gateway throws an exception when echoing an empty payload.

cmebarrow:
Changed bug title from the original "WSN Transport Bug : shouldEchoBinaryFrameWithPayloadLength0 andshouldEchoTextFrameWithPayloadLength0" because we get the same issue with the WSE spec tests (for example: BinaryIT.shouldEchoFrameWithPayloadLength0).

Debugging shows the underlying cause is that the AbstractIoSession.write method so as of IllegalArgumentException if the input data buffer has zero length, i.e. !hasRemaining().

danibusu:
I reproduced the issue. I'm trying to figure out the impact of removing the throw statement from the AbstractIoSession.write method.

danibusu:
@cmebarrow Looking at the gateway project, I can't seem to understand why there's a mina.core module and a mina.build  one. Could you please shed some light?
Also, if I were to change the write() method to no longer throw IllegalArgumentException, I would need to do it in both mina.core and mina,netty, correct?
Thanks!

danibusu:
@cmebarrow I opened this [PR](https://github.com/kaazing/gateway/pull/640), so you can see my changes and discuss whether this is the way to go. The mina modules are used all over the gateway, so I'm not sure of the impact that this change might have. Thanks!

cmebarrow:
[...]

PR #640, opened 2016-07-19
[Work in progress] add support for sending empty payload frame (gateway issue #254)

danibusu:
Changed just the write() method of the AbstractIoSession class in the mina.netty module to not throw IllegalArgumentException, as debugging revelead that this class is the one being used, and not the one in the mina.core module.
Should I also perform the change in the AbstractIoSession from mina.core? When is one class used over the other? Also, what's the purpose of the mina.build module?
I removed the @Ignore annotations from impacted ITs and they all pass.

### DIFF
--- a/mina.netty/src/main/java/org/kaazing/mina/core/session/AbstractIoSession.java
+++ b/mina.netty/src/main/java/org/kaazing/mina/core/session/AbstractIoSession.java
@@ -59,6 +59,8 @@ import org.kaazing.mina.core.future.DefaultWriteFutureEx;
 import org.kaazing.mina.core.future.WriteFutureEx;
 import org.kaazing.mina.core.write.DefaultWriteRequestEx;
 import org.kaazing.mina.core.write.WriteRequestEx;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Base implementation of {@link IoSession}.
@@ -98,6 +100,8 @@ public abstract class AbstractIoSession implements IoSession, IoAlignment {
                 session.writtenBytesThroughput = 0;
             }
     };
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractIoSession.class);
 
     /**
      * An internal write request object that triggers session close.
@@ -426,9 +430,9 @@ public abstract class AbstractIoSession implements IoSession, IoAlignment {
         try {
             if (message instanceof IoBuffer
                     && !((IoBuffer) message).hasRemaining()) {
-                // Nothing to write : probably an error in the user code
-                throw new IllegalArgumentException(
-                "message is empty. Forgot to call flip()?");
+                if (LOG.isInfoEnabled()) {
+                    LOG.info("Message is empty. Forgot to call flip()?");
+                }
             } else if (message instanceof FileChannel) {
                 FileChannel fileChannel = (FileChannel) message;
                 message = new DefaultFileRegion(fileChannel, 0, fileChannel.size());
--- a/mina.netty/src/main/java/org/kaazing/mina/core/session/AbstractIoSession.java
+++ b/mina.netty/src/main/java/org/kaazing/mina/core/session/AbstractIoSession.java
@@ -17,11 +17,7 @@ package org.kaazing.mina.core.session;
 
 import static java.lang.String.format;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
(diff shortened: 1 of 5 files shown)


cmebarrow:
@danibusu I think we only use the AbstractIoSession from mina-netty.

mina.netty/src/main/java/org/kaazing/mina/core/session/AbstractIoSession.java
```diff
@@ -99,6 +101,8 @@ public void operationComplete(CloseFuture future) {
             }
     };
 
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractIoSession.class);
```
jitsni: May be s/LOG/LOGGER

mina.netty/src/main/java/org/kaazing/mina/core/session/AbstractIoSession.java
```diff
@@ -426,9 +430,9 @@ public WriteFuture write(Object message, SocketAddress remoteAddress) {
         try {
```
jitsni: It is there to make sure to catch application errors. I guess we can remove the while try { } block.

But if possible we should limit this change to websocket transport only. Can you paste the complete stack trace  ?

danibusu:
@jitsni I pasted the stack trace below:

<pre><code>
Thread [main] (Suspended (breakpoint at line 430 in AbstractIoSession))	
	WsnSession(AbstractIoSession).write(Object, SocketAddress) line: 430	
	WsnSession(AbstractIoSession).write(Object) line: 391	
	BaseFramingIT.shouldEchoBinaryFrameWithPayloadLength0() line: 97	
	NativeMethodAccessorImpl.invoke0(Method, Object, Object[]) line: not available [native method]	
	NativeMethodAccessorImpl.invoke(Object, Object[]) line: 62	
	DelegatingMethodAccessorImpl.invoke(Object, Object[]) line: 43	
	Method.invoke(Object, Object...) line: 498	
	FrameworkMethod$1.runReflectiveCall() line: 50	
	FrameworkMethod$1(ReflectiveCallable).run() line: 12	
	FrameworkMethod.invokeExplosively(Object, Object...) line: 47	
	InvokeMethod.evaluate() line: 17	
	JUnitRuleMockery$1.evaluate() line: 49	
	SpecificationStatement.evaluate() line: 63	
	Verifier$1.evaluate() line: 35	
	WsnConnectorRule$ConnectorStatement.evaluate() line: 134	
	TestWatcher$1.evaluate() line: 55	
	RunRules.evaluate() line: 20	
	BlockJUnit4ClassRunner(ParentRunner<T>).runLeaf(Statement, Description, RunNotifier) line: 325	
	BlockJUnit4ClassRunner.runChild(FrameworkMethod, RunNotifier) line: 78	
	BlockJUnit4ClassRunner.runChild(Object, RunNotifier) line: 57	
	ParentRunner$3.run() line: 290	
	ParentRunner$1.schedule(Runnable) line: 71	
	BlockJUnit4ClassRunner(ParentRunner<T>).runChildren(RunNotifier) line: 288	
	ParentRunner<T>.access$000(ParentRunner, RunNotifier) line: 58	
	ParentRunner$2.evaluate() line: 268	
	BlockJUnit4ClassRunner(ParentRunner<T>).run(RunNotifier) line: 363	
	JUnit4TestReference.run(TestExecution) line: 86	
	TestExecution.run(ITestReference[]) line: 38	
	RemoteTestRunner.runTests(String[], String, TestExecution) line: 459	
	RemoteTestRunner.runTests(TestExecution) line: 675	
	RemoteTestRunner.run() line: 382	
	RemoteTestRunner.main(String[]) line: 192	
</pre></code>

I am thinking of updating the WsnConnector and WsnAcceptor classes and whenever a new WsnSession is created, also set a session attribute like "wsFlag" to true. And then, in the AbstractIoSession.write() method, throw the IllegalArgumentException only if the flag is not set. Do you think this approach is the right one?

I'm not sure about removing the try-catch block. For example, the FileInputStream constructor called in one of the if  blocks may throw FileNotFoundException.

jitsni:
One option is to create a method void validateWriteMessage((Object message) with default behaviour.  WsnSession overrides this method to allow empty message. But seems like heavy weight.

I think the current change is ok. Just remove the whole try block. I don't we use File or FileChannel as message.

danibusu:
@jitsni Updated pull request to incorporate review comments. Thanks!

PR #644, opened 2016-07-19
Http acceptor tests pulled from K3po

a-zuckut:


a-zuckut:
For https://github.com/kaazing/roadmap/issues/483

transport/http/pom.xml
```diff
@@ -102,6 +102,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.kaazing</groupId>
+            <artifactId>gateway.service.http.directory</artifactId>
```
DoruM: Hi Alex,

This change here is wrong. By introducing it you created a cyclic dependency. The artifact gateway.service.http.directory depends on gateway.transport.http as well. So maven will not know which to build first.

Best regards,
Doru

transport/http/src/test/java/org/kaazing/gateway/transport/http/acceptor/specification/rfc7230/MessageFormatIT.java
```diff
@@ -0,0 +1,423 @@
...
+    @Test
+    @Specification({"server.must.reject.header.with.space.between.header.name.and.colon/request"})
+    public void serverMustRejectHeaderWithSpaceBetweenHeaderNameAndColon() throws Exception {
+        final CountDownLatch latch = new CountDownLatch(1);
+
+        final IoHandler acceptHandler = new IoHandlerAdapter<HttpAcceptSession>() {
+            @Override
+            protected void doSessionOpened(HttpAcceptSession session) throws Exception {
+                latch.countDown();
+                session.setStatus(HttpStatus.CLIENT_BAD_REQUEST);
```
DoruM: This line is incorrect. It should return HttpStatus.SUCCESS_OK. 
The acceptor itself should return HttpStatus.CLIENT_BAD_REQUEST, but not because it calls doSessionOpened, but as a validation before actually calling doSessionOpen. Otherwise, we aren't actually checking the gateway but this newly created acceptHandler, and this is not what we want.

This comment applies to other tests in this commit as well.

@a-zuckut, @dpwspoon

dpwspoon:
@sbadugu let's back log this for now...?

PR #651, opened 2016-07-25
turn.rest service

vinaysatish:
includes modifications to how properties work


dpwspoon:
This adds a "turn.rest" service

with a service `<property> of `<generate.credentials>class:SomeClass</generate.credentials>`

And options that get passed to that class in form of 

```
.nestedProperty("options")
 +                                .property("secret", "secret")
 +                                .property("symbol", ":")
 +                                .nestedProperty("uris")
 +                                    .property("uri", "uri1")
 +                                    .property("uri", "uri2")
 +                                    .property("uri", "uri3")
 +                                .done()
 +                            .done()
```

With class
```java
public interface TurnRestCredentialGenerator {
    
    /**
      * 
      * @param options nested property containing a shared secret key, timestamp/username
      * separator to be used in the generated username, etc.  
      */
     public void init(ServiceProperties options);
 
    /**
      * 
      * @param parameterUsername username passed in as HTTP parameter if present, null otherwise 
      * @param subject Subject representing the currently logged in user
      * @return TurnRestCredentials object 
      */
     public TurnRestCredentials generateCredentials(String parameterUsername, Subject subject);
```

And rest credentials of 
```java
public interface TurnRestCredentials {

    public String getUsername();
    
    public String getPassword();
    
    /**
     * 
     * @return long data type representing the duration for which the credentials are valid
     * (time to live)
     */
    public long getTTL();
    
    public String getURIs();
    
    public String getResponseString();
    
}
```

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/TurnRestCredentialGenerator.java
```diff
@@ -0,0 +1,38 @@
...
+     */
+    public void init(ServiceProperties options);
+
+    /**
+     * 
+     * @param parameterUsername username passed in as HTTP parameter if present, null otherwise 
+     * @param subject Subject representing the currently logged in user
+     * @return TurnRestCredentials object 
+     */
+    public TurnRestCredentials generateCredentials(String parameterUsername, Subject subject);
```
dpwspoon: Need to think about the username and api key future compatibility on the parameters

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/TurnRestCredentials.java
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * 
+     * @return long data type representing the duration for which the credentials are valid
+     * (time to live)
+     */
+    public long getTTL();
+    
+    public String getURIs();
+    
+    public String getResponseString();
```
dpwspoon: Have a utility method in a utility class that constructs that string.  This utility method/class will not go into the public API

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/TurnRestCredentials.java
```diff
@@ -0,0 +1,35 @@
...
+    public String getPassword();
+    
+    /**
+     * 
+     * @return long data type representing the duration for which the credentials are valid
+     * (time to live)
+     */
+    public long getTTL();
+    
+    public String getURIs();
```
dpwspoon: pull (URIs) this out into the service configuration directly

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/TurnRestCredentials.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kaazing.gateway.service.turn.rest;
+
+public interface TurnRestCredentials {
```
dpwspoon: Create an actual class implementation of this that end users can use

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/TurnRestCredentialGenerator.java
dpwspoon: only need Subject

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/TurnRestCredentialGenerator.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kaazing.gateway.service.turn.rest;
+
+import javax.security.auth.Subject;
+
+import org.kaazing.gateway.service.ServiceProperties;
+
+public interface TurnRestCredentialGenerator {
```
dpwspoon: TurnRestCredentialsGenerator

dpwspoon:
When the service does not have a CredentialGenerator, have it return the endpoint json with out credentials

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/TurnRestCredentialGenerator.java
```diff
@@ -0,0 +1,38 @@
...
+    
+    /**
+     * 
+     * @param options nested property containing a shared secret key, timestamp/username
+     * separator to be used in the generated username, etc.  
+     */
+    public void init(ServiceProperties options);
```
dpwspoon: Remove this init.  Add setters for TTL, Key, and Separator. 

Add to the turn.rest service config the `<ttl>`, `<key.alias>` `<separator>`.  If key.alias is called we pass in Keystore key from that alias

dpwspoon:
Set Max Age header to the TTL.

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/TurnRestCredentials.java
```diff
@@ -0,0 +1,35 @@
...
+    public String getUsername();
+    
+    public String getPassword();
+    
+    /**
+     * 
+     * @return long data type representing the duration for which the credentials are valid
+     * (time to live)
+     */
+    public long getTTL();
```
dpwspoon: consider removing TTL.

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/TurnRestCredentialGenerator.java
dpwspoon: rename method to generate

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/TurnRestCredentials.java
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kaazing.gateway.service.turn.rest;
+
+public interface TurnRestCredentials {
+
+    public String getUsername();
+    
+    public String getPassword();
```
dpwspoon: change password to char[], and null out values on use

pom.xml
```diff
@@ -73,6 +73,7 @@
         <module>service/http.directory</module>
         <module>service/http.proxy</module>
         <module>service/proxy</module>
+        <module>service/turn.rest</module>
```
DoruM: This should be placed after <module>test.util</module> . Otherwise, the compilation won't work. 

The module service/turn.rest depends upone the EarlyAccessFeature which is declared in ateway.util/src/main/java/org/kaazing/gateway/util/feature/EarlyAccessFeatures.java .

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/TurnRestService.java
```diff
@@ -0,0 +1,127 @@
...
+public class TurnRestService implements Service {
+    
+    private static final String CLASS_PREFIX = "class:";
+    
+    private TurnRestServiceHandler handler;
+    private ServiceContext serviceContext;
+    private SecurityContext securityContext;
+    private Properties configuration;
+
+    public TurnRestService() {
```
DoruM: No reason for this constructor here.

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/TurnRestService.java
```diff
@@ -0,0 +1,127 @@
...
+        StringBuilder u = new StringBuilder();
+        for (String uri: properties.getNested("uris").get(0).get("uri").split(LIST_SEPARATOR)) {
+            u.append("\"").append(uri).append("\",");
+        }
+        u.setLength(u.length() - 1);
+        String uris = u.toString();
+        
+        String credentialGeneratorClassName = properties.get("generate.credentials");
+        TurnRestCredentialsGenerator credentialGeneratorInstance = null;
+        if (credentialGeneratorClassName != null) {
```
DoruM: This code contains way too many nested ifs. This check should be 
`if (credentialGeneratorClassName != null) {`
`    return;`
`}`

The if/else below (on line 81) should be similarly treated:

` if (!credentialGeneratorClassName.startsWith(CLASS_PREFIX)) {`
`     throw new IllegalArgumentException("Class name must have \"class:\" prefix.");`
`}`

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/TurnRestServiceFactorySpi.java
```diff
@@ -0,0 +1,36 @@
...
+public class TurnRestServiceFactorySpi extends ServiceFactorySpi {
+
+    @Override
+    public Collection<String> getServiceTypes() {
+        return Collections.singletonList("turn.rest");
+    }
+
+    @Override
+    public Service newService(String serviceType) {
+        assert "turn.rest".equals(serviceType);
```
DoruM: The evaluation here should not be placed in an assert. Asserts are turned on or of by the JVM. Assuming the asserts are not turned on, the code here will not be evaluated. Better throw a RuntimeException.

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/TurnRestServiceHandler.java
```diff
@@ -0,0 +1,110 @@
...
+        Subject subject = session.getSubject();
+        
+        if (ttl == null) {
+            ttl = session.getParameter("Max-Age");
+        }
+       
+        TurnRestCredentials credentials = null;
+        String username = null;
+        char[] password = null;
+        if (credentialGenerator != null) {
```
DoruM: What happens if credentialGenerator is null?

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/TurnRestService.java
```diff
@@ -0,0 +1,127 @@
...
+        
+        StringBuilder u = new StringBuilder();
+        for (String uri: properties.getNested("uris").get(0).get("uri").split(LIST_SEPARATOR)) {
+            u.append("\"").append(uri).append("\",");
+        }
+        u.setLength(u.length() - 1);
+        String uris = u.toString();
+        
+        String credentialGeneratorClassName = properties.get("generate.credentials");
+        TurnRestCredentialsGenerator credentialGeneratorInstance = null;
```
DoruM: Remove this null assignment. This is not needed in current version of Java.

Even assigning null to a variable is a dead store if the variable is not subsequently used. Assigning null as a hint to the garbage collector used to be common practice, but is no longer needed and such code should be eliminated.

service/spi/src/main/java/org/kaazing/gateway/service/ServiceProperties.java
```diff
@@ -44,6 +44,8 @@
  */
 public interface ServiceProperties {
 
+    public static final String LIST_SEPARATOR = ",";
```
DoruM: This LIST_SEPARATOR constant is an implementation detail of the implementation of TurnRestService. Another class might use another LIST_SEPARATOR. This should be moved in TurnRestService

DoruM:
There is no change done to the gateway config xsd. So these changes can't be used by the user, by accessing the gateway config.

DoruM:
There is no change done in the distribution/pom.xml. Because of this, the new service will not be included.

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/TurnRestServiceHandler.java
```diff
@@ -0,0 +1,110 @@
...
+        
+        if (password != null) {
+            Arrays.fill(password, '0');
+        }
+        
+        // get io buffer for file
+        IoBufferAllocatorEx<?> allocator = session.getBufferAllocator();
+        IoBufferEx out = allocator.wrap(allocator.allocate(response.length())).setAutoExpander(allocator);
+        out.put(response.getBytes());
+        out.putString(responseChars, StandardCharsets.UTF_8.newEncoder());
```
DoruM: Why is there a need to do a `putString` after the initial `put`? This will duplicate the response.

Additionally a `out.flip()` should be added here, otherwise it won't work

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/TurnRestServiceHandler.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    protected void doSessionOpened(HttpAcceptSession session) throws Exception {        
+        HttpMethod method = session.getMethod();
+        String service = session.getParameter("service");
+      
+        String ttl = options.get("credentials.ttl");
+        Certificate alias = keystore.getCertificate(options.get("key.alias"));
+        char separator = options.get("username.separator").charAt(0);
+         
+        if (method != HttpMethod.GET) {
```
DoruM: According to https://tools.ietf.org/html/draft-uberti-rtcweb-turn-rest-00 , the method used should be POST and not GET. 

> To retrieve a new set of credentials, the client makes a HTTP POST request, specifying TURN as the service to allocate credentials for, and optionally specifying a user id parameter.

DoruM:
Created kaazing/gateway#674. The changes are included in that pull request. Closing this one.

PR #652, opened 2016-07-25
Added validation for empty space after HTTP header field name

DoruM:
Also, gateway was made to work with k3po develop-SNAPSHOT

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
         <maven.compiler.testSource>1.8</maven.compiler.testSource>
         <maven.compiler.testTarget>1.8</maven.compiler.testTarget>
         <hazelcast.version>1.9.4.8</hazelcast.version>
-        <k3po.version>3.0.0-alpha-34</k3po.version>
+        <k3po.version>3.0.0-alpha-37</k3po.version>
         <jdom.version>1.1</jdom.version>
         <jmock.version>2.6.0</jmock.version>
         <slf4j.log4j.version>1.7.21</slf4j.log4j.version>
--- a/transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpHeaderDecodingState.java
+++ b/transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpHeaderDecodingState.java
@@ -15,6 +15,9 @@
  */
 package org.kaazing.gateway.transport.http.bridge.filter;
 
+
+import static org.kaazing.gateway.transport.http.HttpStatus.CLIENT_BAD_REQUEST;
+
 import java.nio.charset.Charset;
 import java.nio.charset.CharsetDecoder;
 import java.util.ArrayList;
@@ -113,6 +116,9 @@ public abstract class HttpHeaderDecodingState extends DecodingStateMachine {
                 throw new ProtocolDecoderException("Invalid header name in the request");
             }
             lastHeaderName = buffer.getString(asciiDecoder);
+            if (!lastHeaderName.trim().equals(lastHeaderName)) {
+                throw new HttpProtocolDecoderException(CLIENT_BAD_REQUEST);
+            }
             return AFTER_READ_HEADER_NAME;
         }
     };


pom.xml
```diff
@@ -28,7 +28,7 @@
         <maven.compiler.testSource>1.8</maven.compiler.testSource>
         <maven.compiler.testTarget>1.8</maven.compiler.testTarget>
         <hazelcast.version>1.9.4.8</hazelcast.version>
-        <k3po.version>3.0.0-alpha-34</k3po.version>
+        <k3po.version>develop-SNAPSHOT</k3po.version>
```
dpwspoon: I've released 3.0.0-alpha-37

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpAcceptor.java
```diff
@@ -467,6 +470,37 @@ public DefaultHttpSession call() {
                         SESSION_KEY.set(session, newHttpSession);
                         return newHttpSession;
                     }
+
+
+
+                    private IoHandler getIoHandler(DefaultHttpSession newHttpSession, boolean invalidHeaders) {
```
dpwspoon: I would think it would be better to implement this logic in the decoder, as I would in general expect the decoder to detect badly formatted requests responses.  I'm not overly familiar with where in the Decoder this would go.  Perhaps somewhere near: https://github.com/kaazing/gateway/blob/develop/transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpHeaderDecodingState.java

DoruM: Hi David,

I would agree if the code would have been written differently.

However, the reason why I chose this particular place is because you need access to the session in order to be able to send back the proper HTTP 400 response. Placing it in the HttpAcceptor is the easier way of doing it. 

The issue is, as ever, the Pokemon exceptions. I would have to fix them everywhere in this code and this bug would take quite a while to fix. 

Let me explain it more clearly:

Within HttpHeaderDecoding we would have to throw a ProtocolDecoderException exception. Now the code, is done in such a way that the function that actually interprets this exception closes the connection. This is not what we want (we want to send back HTTP 400).

The way to do it would be to interpret the exception, and if it is a protocol decoding error we would then send back HTTP 400. This is not doable because the function calling the decoding state machine wraps all the exceptions as "Exception". Anyway, the same old anti-pattern...

Anyway, if you want it done in that particular way, I can do it (after all, that is the proper way of doing it), but I suggest a thorough refactoring if we are going to do it in that particular way.

Kindest regards,
Doru

pom.xml
DoruM: Will alpha-37 include the changes in [PR#348](https://github.com/k3po/k3po/pull/348)? 

If not, it will not work.

dpwspoon: yes, I believe so

dpwspoon:
The transport stack is a pipeline pattern implementation.  In the framework implementation of the pipeline by netty/mina exceptions are treated as events that flow up the pipeline.  Throwables are then caught and then presented in the filters/handlers in the pipeline that can catch them.  

It looks like the HttpProtocolDecoderException is thrown in the decoder, and then caught in the HttpAcceptor (a handler).  The HttpProtocolDecoderException looks like you can set the exception type on this.  See example of this in HttpRequestLineDecodingState.

The model looks like this
```

 Lower level transport                       pipeline                       AcceptHandler
                                     |---------------------------------------------|

                                            |                   |              |            
                                         filter             filter         filter 
```

I think the design approach is taken to allow the AcceptHandler to decide what to do with the sessions when they throw exceptions, and cleaning up external resources that could be associated with it.  For example the AcceptHandler could be a Proxy Handler that needs to close the frontend.  As a framework design I think this is ok and you are mistaking the catch all on the throwables as a Pokemon design pattern, but they are really presented strongly typed at the AcceptHandler in the exceptionCaught method.  Implementing strongly typed exception handling for individual filters (or sets of filters) would then just be introducing another subsystem, that may or may not be necessary depending on the use case for that filter (or set of filters), but that would add more complexity.

Let me know if that works for you think...

DoruM:
Hi David,

Indeed, it is much cleaner this way. And I have missed it, indeed. I changed the implementation accordingly.

Best regards,
Doru

PR #656, opened 2016-07-26
Help written from file

DoruM:
This allows for additional help messages to be written from a csv file. This is in conjuction with the paramter "--broker" which is in Enterprise gateway. This parameter is parsed and executed in the launching script of the gateway, and not in the gateway itself

### DIFF
--- a/distribution/src/main/assembly/generic-bin.xml
+++ b/distribution/src/main/assembly/generic-bin.xml
@@ -103,6 +103,7 @@
         <include>commons-cli:commons-cli</include>
         <include>org.json:json</include>
         <include>org.apache.maven.shared:maven-shared-utils</include>
+        <include>net.sf.opencsv:opencsv</include>
       </includes>
     </dependencySet>
 
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -253,7 +253,7 @@
         <dependency>
             <groupId>net.sf.opencsv</groupId>
             <artifactId>opencsv</artifactId>
-            <version>2.3</version>
+            <version>${opencsv.version}</version>
         </dependency>
     </dependencies>
 
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -250,6 +250,11 @@
             <classifier>keystore</classifier>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>net.sf.opencsv</groupId>
+            <artifactId>opencsv</artifactId>
+            <version>2.3</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/server/src/main/java/org/kaazing/gateway/server/GatewayCommandLineProcessor.java
+++ b/server/src/main/java/org/kaazing/gateway/server/GatewayCommandLineProcessor.java
@@ -15,15 +15,23 @@
  */
 package org.kaazing.gateway.server;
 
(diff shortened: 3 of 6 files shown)


distribution/src/main/assembly/generic-bin.xml
```diff
@@ -103,6 +103,7 @@
         <include>commons-cli:commons-cli</include>
         <include>org.json:json</include>
         <include>org.apache.maven.shared:maven-shared-utils</include>
+        <include>net.sf.opencsv:opencsv</include>
```
dpwspoon: We need to add this to the notice.txt of the gateway and ee gateway now that we are using this

server/src/main/java/org/kaazing/gateway/server/GatewayCommandLineProcessor.java
```diff
@@ -107,19 +121,70 @@ public void run() {
...
+
+                while ((nextLine = script_options.readNext()) != null) {
+                    Option o = new Option(null, nextLine[0], Boolean.parseBoolean(nextLine[1]), nextLine[2]);
+                    all_opt.addOption(o);
+                }
+                script_options.close();
+            }
+
+        } catch (IOException e) {
+            System.out.println("hex");
```
dpwspoon: Why the system.out with "hex"

server/src/main/java/org/kaazing/gateway/server/GatewayCommandLineProcessor.java
```diff
@@ -107,19 +121,70 @@ public void run() {
...
+            
+            if (!o1.getLongOpt().equals(SCRIPT_ARG)) {
+                all_opt.addOption(o1);
+            } 
+        }
+
+        // also, get the file that has documentation for the values parsed in the launching script
+        File helpScript = getScriptedOptsFile(cmd);
+        appendScriptedOptions(helpScript, all_opt);
+        helpFormatter.printHelp("gateway.start", all_opt, true);
```
dpwspoon: I don't think we should tie ourselves down to gateway.start

dpwspoon:
Please add a test, it will make it easier to review.  This looks ok to me.  Consider replacing the `SCRIPT_ARG` with an env variable that populates it instead.  It might be a little cleaner usage of it...

server/src/main/java/org/kaazing/gateway/server/GatewayCommandLineProcessor.java
DoruM: Removed.

server/pom.xml
```diff
@@ -250,6 +250,11 @@
             <classifier>keystore</classifier>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>net.sf.opencsv</groupId>
+            <artifactId>opencsv</artifactId>
+            <version>2.3</version>
```
DoruM: Done.

distribution/src/main/assembly/generic-bin.xml
DoruM: Done. Added to gateway/NOTICE.txt, gateway.server/NOTICE.txt, enterprise.gateway.NOTICE.txt

DoruM:
Changes added. Also unit test was added.

PR #661, opened 2016-07-28
Tickets 665 validation issues

DoruM:
This pull request contains also the merges to 5.1.3


DoruM:
Agreed with @AdrianCozma to use another commit.

PR #664, opened 2016-07-28
Corrected namespace for management.xsd

DoruM:
New pull request after discussions with @AdrianCozma .

This change is needed for QA to be able to continue


AdrianCozma:
merging to unblock itests and soak tests

PR #679, opened 2016-08-16
Turn.proxy implementation

Anisotrop:
Implement https://tools.ietf.org/html/rfc5389 encoding/decoding and changes of Mapped-Address and Xor-Mapped-Address attributes. Also Message-Integrity attribute is regenerated when adresses are modified.


server/src/main/xsd/gateway-config-201616.xsd
```diff
@@ -1109,6 +1106,75 @@
...
+                    <element maxOccurs="1" minOccurs="0"
+                             name="connect-options"
+                             type="gateway:ServiceConnectOptionsType" />
+
+                    <!-- TODO is this supported ? -->
+                    <element maxOccurs="1" minOccurs="0"
+                             name="realm-name"
+                             type="gateway:CollapsedString" />
+
+                    <!-- TODO is this supported ? -->
```
dpwspoon: realm, balance, and authorization-constraint might come into play in the future.  cross-site and notify are not, but I think they might have to be in the current form of the xsd otherwise it won't compile...?  (perhaps, put a todo to remove them, see https://github.com/kaazing/roadmap/issues/927 for details)

service/http.proxy/pom.xml
```diff
@@ -77,6 +77,11 @@
             <artifactId>specification.http</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
dpwspoon: I'm not sure if your code requires this change, or if this is a side affect of catchup merge, or if this is not needed at all...?

service/turn.proxy/src/main/java/org/kaazing/gateway/service/turn/proxy/TurnProxyAcceptHandler.java
```diff
@@ -0,0 +1,170 @@
...
+
+public class TurnProxyAcceptHandler extends AbstractProxyAcceptHandler {
+
+    static final Logger LOGGER = LoggerFactory.getLogger(TurnProxyAcceptHandler.class);
+    public static final String TURN_STATE_KEY = "turn-state";
+
+    // Configuration properties
+    public static final String PROPERTY_MAPPED_ADDRESS = "mapped.address";
+    public static final String PROPERTY_KEY_ALIAS = "key.alias";
+    public static final String PROPERTY_KEY_PASS_FILE = "key.password-file";
```
dpwspoon: This file can be removed... see https://github.com/kaazing/gateway/pull/674#discussion_r75177451

service/turn.proxy/src/main/java/org/kaazing/gateway/service/turn/proxy/TurnProxyAcceptHandler.java
```diff
@@ -0,0 +1,170 @@
...
+    public static final String PROPERTY_KEY_ALGORITHM = "key.algorithm";
+
+    private String connectURI;
+    private Key sharedSecret;
+    private String keyAlgorithm;
+
+    private ConcurrentHashMap<String, String> currentTransactions = new ConcurrentHashMap<>();
```
dpwspoon: I think this is a bug... 

1) I think hypothetically two stun sessions might create the same transaction ID at the same time.  One would override the other.

2) Concurrent data structures have a large overhead in terms of performance, especially when done per every message multiple times through decoders and encoders.   Perhaps, a better approach would to create the transaction map per a session and store it on the session object (via Attribute?) and then accesses via the session.  In such a case, access to it should be thread safe (if thread alignment is working) and thus it should all be good.    Note, the RFC says 

> 
   At any time, a client MAY have multiple outstanding STUN requests
   with the same STUN server (that is, multiple transactions in
   progress, with different transaction IDs).  Absent other limits to
   the rate of new transactions (such as those specified by ICE for
   connectivity checks or when STUN is run over TCP), a client SHOULD
   space new transactions to a server by RTO and SHOULD limit itself to
   ten outstanding transactions to the same server.

so perhaps a good initial capacity for the hash map may be 10...?

service/turn.proxy/src/main/java/org/kaazing/gateway/service/turn/proxy/TurnProxyAcceptHandler.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /*
+     * Initializer for connect session. It adds the processed accept session headers on the connect session
+     */
+    class ConnectSessionInitializer implements IoSessionInitializer<ConnectFuture> {
+
+        @Override
+        public void initializeSession(IoSession session, ConnectFuture future) {
+            session.getFilterChain().addLast("STUN_CODEC", new StunCodecFilter(currentTransactions, sharedSecret, keyAlgorithm));
```
dpwspoon: if we do stash the currentTransactions to the session then all the constructors will become cleaner

service/turn.proxy/src/main/java/org/kaazing/gateway/service/turn/proxy/TurnProxyAcceptHandler.java
dpwspoon: I fixed this via another iteration in this PR: https://github.com/Anisotrop/gateway/pull/2

service/turn.proxy/src/main/java/org/kaazing/gateway/service/turn/proxy/stun/StunFrameEncoder.java
```diff
@@ -0,0 +1,151 @@
...
+            LOGGER.trace("Encoding STUN message: " + message);
+        }
+        StunProxyMessage stunMessage = (StunProxyMessage) message;
+        String username = null;
+        if (stunMessage.getMessageClass().equals(StunMessageClass.RESPONSE) ||
+            stunMessage.getMessageClass().equals(StunMessageClass.ERROR)) {
+            username = currentTransactions.remove(Base64.getEncoder().encodeToString(stunMessage.getTransactionId()));
+            LOGGER.trace(String.format("Removed username %s from transactions map", username));
+            if (stunMessage.isModified() && (username == null || sharedSecret ==null)) {
+                LOGGER.warn("Stun message is modified but MESSAGE-INTEGRITY attribute will not be recalculated.");
```
dpwspoon: %s/Stun message is modified but MESSAGE-INTEGRITY attribute will not be recalculated/STUN message is modified but MESSAGE-INTEGRITY attribute can not be recalculated because username and/or shared secret is not available/

service/turn.proxy/src/main/java/org/kaazing/gateway/service/turn/proxy/stun/StunMessageClass.java
```diff
@@ -0,0 +1,51 @@
...
+        case 0x000:
+            return REQUEST;
+        case 0x010:
+            return INDICATION;
+        case 0x100:
+            return RESPONSE;
+        case 0x110:
+            return ERROR;
+        default:
+            throw new IllegalStateException("No such STUN class from: " + leadingBitesAndMessageType);
```
dpwspoon: is this formatted correctly?

service/turn.proxy/src/main/java/org/kaazing/gateway/service/turn/proxy/stun/StunMessageMethod.java
```diff
@@ -0,0 +1,67 @@
...
+        case 0x006:
+            return SEND;
+        case 0x007:
+            return DATA;
+        case 0x008:
+            return CREATE_PERMISSION;
+        case 0x009:
+            return CHANNEL_BIND;
+        default:
+            throw new IllegalStateException("No such STUN method from: " + method);
```
dpwspoon: same question

service/turn.proxy/src/test/java/org/kaazing/gateway/service/turn/proxy/TurnProxyIT.java
```diff
@@ -0,0 +1,107 @@
...
+import java.io.FileInputStream;
+import java.io.FileWriter;
+import java.security.KeyStore;
+
+public class TurnProxyIT {
+
+    private final K3poRule k3po = new K3poRule()
+            .setScriptRoot("org/kaazing/gateway/service/turn/proxy")
+            .scriptProperty("acceptURI 'tcp://localhost:3479'");
+    ;
```
dpwspoon: extra semi colon

dpwspoon:
Generally looks really good to me.  See PR with some updates, and comments above.  Lets try to get this in before start of next sprint.

service/http.proxy/pom.xml
Anisotrop: I removed the dependency and the build of http.proxy service passes.

PR #683, opened 2016-08-24
Made WSE not negotiate x-kaazing-ping-pong extension 

dpwspoon:


cmebarrow:
@dpwspoon Need to add a wse spec test for connection established no ws extensions requested and consume it in org.kaazing.gateway.transport.wseb.specification.wse.connector.OpeningIT (and org.kaazing.gateway.transport.wseb.specification.wse.acceptor.OpeningIT)

cmebarrow:
@dpwspoon David, I think is a much better and cleaner way to fix this. It should be the ping pong extension that knows it does not apply to wse. Making wse resource address deal with it is breaking encapsulation of extensions: it should not be the responsibility of transports to know details about individual extensions. We can allow ping pong extension to deal with it simply by changing method org.kaazing.gateway.transport.ws.bridge.extensions.pingpong.PingPongExtensionFactory.negotiate(ExtensionHeader, ExtensionHelper, WsResourceAddress) to return null if the WsResourceAddress is a wse one. If we do that then probably you can back out the changes to the wse spec tests to pass in a wse instead of a ws resource address.

dpwspoon:
@cmebarrow I generally agree that the fix here brakes the encapsulation of the extension.  However, the extensions are currently set in the DefaultOptionsContext, which already breaks that.  Additionally, your proposal will still have the WseConnector requesting an extension it can not support as the negotiate method is only called from Acceptors (I believe/haven't double checked that).

Perhaps the best fix would be to remove the DefaultOptionsContext setting of the extensions.  Have each Ws*FactorySpi call out to the supported extensions with its scheme to see which extension should be enabled.  In that call it should also send the other resolved options (ex ws.inactivity.timeout) as that seems to be the one special logic case in DefaultOptionsContext (if ws.inactivty.timeout > 0 addOption(IDLE_TIMEOUT).

This would have the behavior that, 

1. All Extensions on the class path are enabled by default to their choosing for the supported schemes and setup scenarios that they support.  
2. The special edge case logic in DefaultAcceptOptionsContext of needing to pass them on to future child layers would no longer be needed, specifically

```java
        result.put("ws.extensions", wsExtensions);
        result.put("ws[ws/rfc6455].ws[ws/rfc6455].extensions", wsExtensions);
        result.put("ws[ws/draft-7x].ws[ws/draft-7x].extensions", wsExtensions);
```
3. Encapsulation is supported

dpwspoon:
@cmebarrow @jitsni 

From discussion offline, best solution is to remove references to extensions in `DefaultOptionsContext` and remove them from the resource address.  Instead have the transports call new method `offer(ResourceAddress address)` from the connector to iterate over which extensions should be offered.  This corresponds to the `negotiate(...)` method that is used for the similar purpose in the Acceptors

dpwspoon:
Closing to do the rework instead

PR #684, opened 2016-08-24
Refactor turn.rest API to match feedback

mgherghe:


service/turn.rest/src/test/java/org/kaazing/gateway/service/turn/rest/TurnRestIT.java
```diff
@@ -67,18 +67,14 @@
...
-                                .property("uri", "uri3")
-                            .done()
-                            .nestedProperty("options")
-                                .property("credentials.ttl", "43200")
-                                .property("username.separator", ":")
-                            .done()
+                            .property("credentials.generator", "class:" + DefaultCredentialsGenerator.class.getName())
+                            .property("credentials.ttl", "43200")
+                            .property("username.separator", ":")
+                            .property("url", "url1")
```
dpwspoon: please change the url1, url2, url3 to be real examples:

for example
"turn:1.2.3.4:9991?transport=udp",
       "turn:1.2.3.4:9992?transport=tcp",
       "turns:1.2.3.4:443?transport=tcp"

as from Spec https://tools.ietf.org/html/draft-uberti-rtcweb-turn-rest-00

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/internal/TurnRestService.java
```diff
@@ -67,8 +69,7 @@ public void init(ServiceContext serviceContext) throws Exception {
         String uris = getTurnURIs(properties);
```
dpwspoon: consider changing %s/uri/url/g

dpwspoon:
LGTM, 2 minor comments

server/src/main/xsd/gateway-config-201616.xsd
```diff
@@ -1622,44 +1622,32 @@
...
-                        <complexType>
-                            <sequence>
-                                <element maxOccurs="unbounded" minOccurs="1" name="uri"
-                                    type="gateway:CollapsedString">
-                                    <annotation />
-                                </element>
-                            </sequence>
-                        </complexType>
+                    <element maxOccurs="1" minOccurs="0" name="credentials.ttl"
+                        type="gateway:CollapsedString">
```
dpwspoon: This should not be of typed collapsed gateway:TimeIntervalString, I believe

mgherghe:
Addressed feedback comments.

PR #690, opened 2016-08-29
Turn.proxy tests

NicoletaOita:
Turn.proxy tests for checking service configuration


dpwspoon:
Please try to combine the misconfiguration test classes into a single file.  You can do this by not using a GW rule, and instead invoking the test.Gateway class directly and catching the Exception.   Please use the ExpectedExceptions rule for this (if not already doing so)

NicoletaOita:
@dpwspoon Ok, I'll integrate the misconfiguration test classes into a single file.

NicoletaOita:
Why has this PR been closed? @dpwspoon Would you rather prefer to make a new PR?

PR #696, opened 2016-09-01
Http Connector authenticating to 401 and Login Module expiring state

dpwspoon:
2 new features:

1) The login modules will have an entry in their Login Map with the key "ExpiringState". The value of this entry will be the new public API of ExpiringState class. (Note: Java Doc needs to be public). This ExpiringState class is shared across all Login Modules and GW members in a cluster and can be used to share information across the login modules. (Like one time nonces, or multifactor credentials)

2) The http connector in the gateway can now respond to http challenges on 401 status codes. It will attempt to provide the credentials via a call to the Java Authenticator. The Authenticator is set on the JVM so we don't provide a GW specific way to configuring it and pointing it to related oracle docs on this should be sufficient. (For example Authenticator.setDefault). The http connector will only attempt to authenticate on the HTTP if a new connect-options property is set, Specifically this is "http.max.authentication.attempts" with an integer value. This specifies how many times an http connector will attempt to authenticate when it is challenged. By default this value is 0.


transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpConnector.java
```diff
@@ -426,15 +446,105 @@ private boolean shouldFollowRedirects(DefaultHttpSession httpSession) {
...
+            return retryConnect(httpSession, session, newConnectAddress);
+        }
+
+        /**
+         * Retries a connect based on new ConnectAddress
+         * @param httpSession
+         * @param session
+         * @param newConnectAddress
+         * @return
+         */
```
jitsni: May be, just delete the javadoc

server.api/src/main/java/org/kaazing/gateway/server/ExpiringState.java
```diff
@@ -0,0 +1,60 @@
...
+package org.kaazing.gateway.server;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * ExpiringState is accessible from the @javax.security.auth.spi.LoginModule
+ * via the Map<String,?> options, under the name "ExpiringState".
+ *
+ */
+public interface ExpiringState {
```
jitsni: If it is public API, shouldn't it be in server.spi module ?

server.api/src/main/java/org/kaazing/gateway/server/ExpiringState.java
```diff
@@ -0,0 +1,60 @@
...
+
+    // TODO remove internally
```
jitsni: should this be removed as per TODO comment ?

server.api/src/main/java/org/kaazing/gateway/server/ExpiringState.java
```diff
@@ -0,0 +1,60 @@
...
+    public static final String NAME = "ExpiringState";
```
jitsni: You can remove "public static final" for this constant.

server.api/src/main/java/org/kaazing/gateway/server/ExpiringState.java
```diff
@@ -0,0 +1,60 @@
...
+     * if the specified key is not already associated with a value.
+     * Entry will expire and get evicted after the ttl.
+     *
+     * @param key      key of the entry
+     * @param value    value of the entry
+     * @param ttl      maximum time for this entry to stay in the map
+     * @param timeunit time unit for the ttl
+     * @return null if absent, or returns the old value of the entry
+     */
+    public abstract Object putIfAbsent(String key, Object value, long ttl, TimeUnit timeunit);
```
jitsni: You can remove "public abstract" for interface methods

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultExpiringState.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.kaazing.gateway.server.ExpiringState;
+import org.kaazing.gateway.service.messaging.collections.CollectionsFactory;
+
+import com.hazelcast.core.IMap;
+
+final class DefaultExpiringState implements ExpiringState {
+    private final IMap<Object, Object> delegate;
+
+    protected DefaultExpiringState(CollectionsFactory collectionsFactory) {
```
jitsni: editor wans about "protected" constructor, suggests package-private scope

server/src/main/xsd/gateway-config-201616.xsd
```diff
@@ -1501,6 +1501,16 @@
                     <documentation>The maximum redirects http connector follows</documentation>
                 </annotation>
             </element>
+            <element name="http.max.authentication.attempts" maxOccurs="1" minOccurs="0" type="nonNegativeInteger">
+                <annotation>
+                    <documentation>The maximum responses to a 401 a http connector will make</documentation>
```
jitsni: Seems partial or want to change wording.

server/src/main/xsd/gateway-config-201616.xsd
```diff
@@ -1501,6 +1501,16 @@
...
+                </annotation>
+            </element>
+            <element name="http.authenticator" maxOccurs="1" minOccurs="0" type="gateway:CollapsedString">
+                <annotation>
+                    <documentation>The authenticator the http connector should authenticate with</documentation>
```
jitsni: What is this ? Is this a class name ?

transport/wsn/src/main/java/org/kaazing/gateway/transport/wsn/WsnAcceptor.java
```diff
@@ -1073,7 +1073,7 @@ private void doUpgrade(final HttpAcceptSession session) throws Exception {
                         return;
                     }
 
-                    // If configured with "Application xxx" challenge scheme, reject and close the connection
+//                     If configured with "Application xxx" challenge scheme, reject and close the connection
```
jitsni: formatting

transport/wsn/src/main/java/org/kaazing/gateway/transport/wsn/WsnConnector.java
```diff
@@ -174,8 +174,8 @@ public void addBridgeFilters(IoFilterChain filterChain) {
 
         // (KG-7391) Use ping and pong to detect and close dead connections, if ws inactivity timeout is active
         final ResourceAddress connectAddress = (ResourceAddress) session.removeAttribute(WSN_CONNECT_ADDRESS_KEY);
-        WsCheckAliveFilter.addIfFeatureEnabled(filterChain, WsnAcceptor.CHECK_ALIVE_FILTER,
-                connectAddress.getOption(WsResourceAddress.INACTIVITY_TIMEOUT), logger);
+//        WsCheckAliveFilter.addIfFeatureEnabled(filterChain, WsnAcceptor.CHECK_ALIVE_FILTER,
+//                connectAddress.getOption(WsResourceAddress.INACTIVITY_TIMEOUT), logger);
```
jitsni: This filter is not needed anymore ?? 
If yes, assigning to connectAddress is not required

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/proxy/ResetAuthenticatorRule.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package org.kaazing.gateway.transport.wsn.proxy;
+
+import static java.net.Authenticator.setDefault;
+
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+public class ResetAuthenticatorRule implements TestRule {
```
jitsni: IDE shows that it is not used.

server.api/src/main/java/org/kaazing/gateway/server/ExpiringState.java
dpwspoon: It actually is, but the packages are the same so you wouldn't know unless you expanded the path.  Due to this confusion I've filed this story to combine the modules: https://github.com/kaazing/gateway/issues/700

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpConnector.java
dpwspoon: yes

server.api/src/main/java/org/kaazing/gateway/server/ExpiringState.java
dpwspoon: yes, thanks

server.api/src/main/java/org/kaazing/gateway/server/ExpiringState.java
dpwspoon: Thanks, it was an abstract class before and I missed this in the refactor

server/src/main/xsd/gateway-config-201616.xsd
dpwspoon: see update

server/src/main/xsd/gateway-config-201616.xsd
dpwspoon: This should of been removed in refactoring. Now we use default Authenticator on JVM.  In the future, we might have our own Authenticator class you can implement.  This will allow us to scope the Authenticator to a service.

transport/wsn/src/main/java/org/kaazing/gateway/transport/wsn/WsnConnector.java
dpwspoon: I think this changed snuck in from another ticket I was working on.  I've reverted the change.

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/proxy/ResetAuthenticatorRule.java
dpwspoon: Thanks, this should of been removed in a refactor that put it in the http transport.

dpwspoon:
Thank you @jitsni, I've applied all the feedback except https://github.com/kaazing/gateway/pull/696#discussion_r77426354.  This is ready for another round.

PR #701, opened 2016-09-06
Changed turn.rest json response to include stun URLs

mgherghe:


service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/internal/TurnRestService.java
```diff
@@ -67,12 +67,12 @@ public void init(ServiceContext serviceContext) throws Exception {
         EarlyAccessFeatures.TURN_REST_SERVICE.assertEnabled(getConfiguration(), serviceContext.getLogger());
         ServiceProperties properties = serviceContext.getProperties();
 
-        String urls = getTurnURLs(properties);
+        String[] urls = getUrls(properties);
```
dpwspoon: The URLs variable here is hiding the contents, that is urls[0] has the turn URLs and urls[1] has the stun URLs.  This makes reading the code harder.  I would suggest you assign the values to named variables.  Either with: 
```
turnUrls = urls[0];
stunUrls = urls[1];
```

or IMHO eliminating the `urls` variable altogether by either passing in lists that get populated to the getUrls call, or just separating the logic into two separate methods, or by calling the method `getTurnUrls` and then removing the entries in the result from the array.

dpwspoon:
Conceptually looks good to me.  One minor feedback.  @DoruM please feel free to merge after the code change update.

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/internal/TurnRestService.java
dpwspoon: One other approach would be to do it in one line with java 8 lambda/filter

PR #705, opened 2016-09-14
WebSocket extensions decide whether to be part of handshake or not

jitsni:


transport/ws/src/main/java/org/kaazing/gateway/transport/ws/bridge/extensions/idletimeout/IdleTimeoutExtensionFactory.java
```diff
@@ -31,6 +31,12 @@ public String getExtensionName() {
     }
 
     @Override
+    public WebSocketExtension offer(ExtensionHelper extensionHelper, WsResourceAddress address) {
+        // connectors are not sending this extension in handshake request
```
dpwspoon: Please check with @jfallows that this is ok.  Also, I'm curious if this is how 4.x behaved.  I don't think this was the default behavior of 5.x out of the gateway when it was a transport for socks.

transport/ws/src/main/java/org/kaazing/gateway/transport/ws/bridge/extensions/pingpong/PingPongExtensionFactory.java
```diff
@@ -36,6 +36,12 @@ public ExtensionOrderCategory getOrderCategory() {
     }
 
     @Override
+    public WebSocketExtension offer(ExtensionHelper extensionHelper, WsResourceAddress address) {
+        // This extension is used for client <--> gateway, but not for gateway <--> gateway
```
dpwspoon: I believe this is ok, but probably want to double check with John

transport/wseb/src/main/java/org/kaazing/gateway/transport/wseb/WsebConnector.java
```diff
@@ -294,12 +323,11 @@ public void initializeSession(IoSession session, T future) {
...
-                List<String> wsExtensions = connectAddressNext.getOption(WsResourceAddress.EXTENSIONS);
-                if (wsExtensions!= null) {
-                    for (String extension : wsExtensions) {
-                        if (extension != null) {
-                            httpSession.addWriteHeader(HEADER_X_WEBSOCKET_EXTENSIONS, extension);
-                        }
+                List<WebSocketExtension> extensions = webSocketExtensionFactory.offerWebSocketExtensions(
+                        (WsResourceAddress) connectAddressNext, extensionHelper);
+                for (WebSocketExtension extension : extensions) {
+                    if (extension != null) {
```
dpwspoon: is `if (extension != null)` better than `assert extension != null`?.  My reading of extension factory is that it can never be null

transport/wseb/src/test/java/org/kaazing/gateway/transport/wseb/specification/wse/connector/OpeningIT.java
```diff
@@ -125,31 +125,6 @@ public void shouldEstablishConnectionWithRequestHeaderXWebSocketProtocol() throw
         k3po.finish();
     }
```
dpwspoon: We should really still test this as it is part of the spec, but to do this you would need to mock the extension factory or something...

transport/wseb/src/test/java/org/kaazing/gateway/transport/wseb/specification/wse/connector/OpeningIT.java
```diff
@@ -335,33 +310,6 @@ public void shouldFailConnectionWhenXWebSocketProtocolNotNegotiated() throws Exc
...
 
     @Test
```
dpwspoon: We should really still test this as it is part of the spec, but to do this you would need to mock the extension factory or something...

transport/wseb/src/test/java/org/kaazing/gateway/transport/wseb/specification/wse/connector/OpeningIT.java
dpwspoon: Maybe rewrite the spec scripts to parameterize what extensions are requested

transport/wseb/src/test/java/org/kaazing/gateway/transport/wseb/specification/wse/connector/OpeningIT.java
dpwspoon: Maybe you could parameterize the spec tests so they work for us

transport/wseb/src/test/java/org/kaazing/gateway/transport/wseb/specification/wse/connector/OpeningIT.java
dpwspoon: nvm, I see you moved these tests below

transport/wseb/src/test/java/org/kaazing/gateway/transport/wseb/specification/wse/connector/OpeningIT.java
dpwspoon: nvm, I see you moved these tests below

transport/ws/src/main/java/org/kaazing/gateway/transport/ws/bridge/extensions/idletimeout/IdleTimeoutExtensionFactory.java
jitsni: 4.x (and 5.x) is not sending either. Hence I didn't change in the change set. They use WsCheckAliveFilter which gets added when inactivity.timeout is configured.

transport/wseb/src/main/java/org/kaazing/gateway/transport/wseb/WsebConnector.java
jitsni: Yes, that is true. I will fix that.

transport/wseb/src/test/java/org/kaazing/gateway/transport/wseb/specification/wse/connector/OpeningIT.java
jitsni: The problem is to set up a classloader (so that tests are isolated) and hence I moved them to separate tests.

transport/ws/src/main/java/org/kaazing/gateway/transport/ws/bridge/extensions/pingpong/PingPongExtensionFactory.java
dpwspoon: I checked with John and this is ok

transport/ws/src/main/java/org/kaazing/gateway/transport/ws/bridge/extensions/idletimeout/IdleTimeoutExtensionFactory.java
dpwspoon: I see 

In WSN Connector
```
        // (KG-7391) Use ping and pong to detect and close dead connections, if ws inactivity timeout is active
        final ResourceAddress connectAddress = (ResourceAddress) session.removeAttribute(WSN_CONNECT_ADDRESS_KEY);
        WsCheckAliveFilter.addIfFeatureEnabled(filterChain, WsnAcceptor.CHECK_ALIVE_FILTER,
                connectAddress.getOption(WsResourceAddress.INACTIVITY_TIMEOUT), logger);
```

I would expect it would be better to use the extension for this purpose instead of doing it this way.  The difference in the implementations would be that the extension would hit the wire in the extension header, and then the value could be negotiated.  Since it apparently has never worked that way, its probably ok to check it in as it is now and then open an issue to fix it.

jitsni:
@dpwspoon Ready for one more round of review/merge

PR #658, opened 2016-07-26
Connector tests for HTTP

a-zuckut:
For https://github.com/kaazing/roadmap/issues/483

PR #706, opened 2016-09-14
Connector tests for HTTP

justinma246:
Following a-zuckut's changes from https://github.com/kaazing/gateway/pull/658#issuecomment-244206152 for ticket https://github.com/kaazing/roadmap/issues/483


resource.address/http/src/main/java/org/kaazing/gateway/resource/address/http/HttpResourceAddress.java
```diff
@@ -426,6 +440,18 @@ private HttpServerHeaderOption() {
...
 
+    private static final class HttpUserAgentHeaderOption extends HttpResourceOption<Boolean> {
+        private HttpUserAgentHeaderOption() {
+            super(Kind.USER_AGENT_HEADER, "userAgentHeaderEnabled", Boolean.TRUE);
+        }
+    }
+
+    private static final class HttpHostHeaderOption extends HttpResourceOption<Boolean> {
+        private HttpHostHeaderOption() {
+            super(Kind.HOST_HEADER, "hostHeaderEnabled", Boolean.TRUE);
```
dpwspoon: According to spec http should always send a host header.  I believe

dpwspoon: Thus this should not be needed. `   A client MUST send a Host header field in all HTTP/1.1 request
   messages.` f from :https://tools.ietf.org/html/rfc7230#section-5.4

justinma246: It seems like calling the constructor is necessary as there is no valid constructor with no arguments.

Should I remove all instances of HttpHostHeaderOption instead?

sbadugu:
This PR was incorporated in #807 so closing this one.

PR #711, opened 2016-09-16
Worker threads are going to PARK state unnecessarily

jitsni:
Worker threads were going to PARK state since processRead() doesn't have any work done in ring buffer.  Now the work done from the selector (plus work
done from reading ring buffer) is fed into idle strategy.

Also, selector selects immediately. If there is no work done idle strategy
takes care of parking etc

### DIFF
--- a/mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioSelector.java
+++ b/mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioSelector.java
@@ -59,6 +59,8 @@ import org.jboss.netty.util.ThreadRenamingRunnable;
 import org.jboss.netty.util.internal.DeadLockProofWorker;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioSelector.java
```diff
@@ -351,8 +355,9 @@ public void run() {
                     shutdownLatch.countDown();
                     break;
                 } else {
-                    process(selector);
-                    processRead();
+                    workCount += process(selector);
+                    workCount += processRead();
+                    idleStrategy.idle(workCount);
```
jfallows: The intent might be clearer if the `idleStrategy.idle(workCount)` is moved immediately outside this `else` clause, since `workCount` is also modified before the `else` clause. The main reason this all works is due to the `break` at line 356 above, but that's a bit too subtle and moving `idleStrategy.idle(workCount)` is probably more clear.

jitsni: done

PR #713, opened 2016-09-20
Turn Proxy - Unify Logging Output

mgherghe:


service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/internal/TurnRestServiceHandler.java
```diff
@@ -75,6 +78,7 @@ protected void doSessionOpened(HttpAcceptSession session) throws Exception {
             TurnRestCredentials credentials = credentialGenerator.generate(subject);
             username = credentials.getUsername();
             password = credentials.getPassword();
+            LOGGER.info(String.format("Generated username: %s, generated password: %s", username, new String(password)));
```
mgherghe: Should anything else be logged besides the username and password?

transport/spi/src/main/java/org/kaazing/gateway/transport/LoggingFilter.java
```diff
@@ -134,14 +135,8 @@ public static boolean addIfNeeded(Logger logger, IoSession session, String trans
         if (!logger.isInfoEnabled()) {
             return false;
         }
-        String user = getUserIdentifier(session);
         String loggingFilterName = transportName + "#logging";
```
dpwspoon: why did you replace this user?

service/turn.rest/src/main/java/org/kaazing/gateway/service/turn/rest/internal/TurnRestServiceHandler.java
dpwspoon: Don't log the password, log the username with the session

transport/spi/src/main/java/org/kaazing/gateway/transport/LoggingFilter.java
mgherghe: I extracted the logic for getting the user into the `getLoggingFormat` method in order to be able to also use this in derived classes of the LoggingFilter class.

PR #727, opened 2016-10-03
Tcp.ip.whitelist

danibusu:
Hi,

This is the first part of the tcp realm accept option feature. The second part will be raised on the enterprise gateway repo.


server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultServiceContext.java
```diff
@@ -729,7 +729,28 @@ private void injectServiceOptions(String transportURI, Map<String, Object> optio
...
+                                    serviceName, tcpRealmOptionName);
+                            throw new IllegalArgumentException(
+                                    tcpRealmOptionName + " needs to be the same as the configured realm");
+                        }
+                        options.put(format(optionPattern, LOGIN_CONTEXT_FACTORY), serviceRealmContext.getLoginContextFactory());
+//                    initial support will just be authenticate or not
+//                    options.put(format(optionPattern, REQUIRED_ROLES),
+//                            getRequireRoles());
+//                    options.put(format(optionPattern, REALM_USER_PRINCIPAL_CLASSES.name()),
+//                            getUserPrincipalClasses(serviceRealmContext.getUserPrincipalClasses()));
```
jfallows: What is the rationale for requiring the `tcp.realm` to match the `<service><realm-name>`?

For example, could the above code (lines 735-753) be replaced by the following:

```
UnaryOperator<String> tcpFormat = o -> String.format("tcp.%s", o);
String tcpRealm = (String) options.get(tcpFormat.apply("realm"));
if (tcpRealm != null) {
    options.put(tcpFormat.apply(LOGIN_CONTEXT_FACTORY), serviceRealmContext.getLoginContextFactory());
}
```

transport/nio/src/test/java/org/kaazing/gateway/transport/tcp/specification/TcpAcceptorRule.java
```diff
@@ -127,4 +127,8 @@ public TcpAcceptorRule addConfigurationProperty(String propertyName, String prop
         return this;
     }
 
+    public ResourceAddressFactory getAddressFactory() {
+        return addressFactory;
+    }
+
```
jfallows: Is this change still needed?

danibusu: It was done by David when we initiated the tcp ip whitelist support, so I'm not sure why it was done, but I will change it back. Let me know if I should also add the test he removed.

server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultServiceContext.java
danibusu: A service context holds just one realm context and that is resolved to the realm value set in the '<service><realm-name>' tag. That's why I mentioned earlier that both `tcp.realm` and `realm-name` need to be specified in order for the realm context to be added to the service. I can resolve the service realm context using the `tcp.realm` accept option value when the `realm-name` is not specified, but that would require adding `tcp.realm` to the community xsd also.

danibusu:
@jfallows I updated also the tcp resource address name to `REALM`, as you requested. I don't completely agree, because other resource address classes use `REALM_NAME` as an option name and it might get confusing.

jfallows:
@danibusu: as mentioned, we have some inconsistency in the `gateway-config.xml` syntax that refers to `<realm>`, `<realm-name>` and was going to be `tcp.realmName`.

After your update, we now have `<realm>`, `<realm-name>` and `tcp.realm`.

It would make sense to modify both `<realm-name>` to `<realm>` (maintaining backwards compatibility for existing configs) and also change any internal references to `REALM_NAME` to `REALM` too.

If the inconsistency is bothering you, then please feel free to make the internal `REALM_NAME` changes to `REALM` as part of this PR.

danibusu:
@jfallows I can do the internal changes, but I don't know the impact it will have. I'm betting it won't be just a simple rename. Is it OK with you if I address this issue in another pull request, after I finish more pressing work? Thanks!

PR #746, opened 2016-10-18
Altered NioWorker.select and AbstractNioSelector.select to use 10ms select timeout

cmebarrow:
… instead of 0, and commented out use of idleStrategy in AbstractNioSelector.run.

Note: this is a preliminary fix for a performance regression. A better fix (using selector.wakeUp) will be done for the next release.


mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioSelector.java
```diff
@@ -503,11 +503,15 @@ protected int processRead() throws IOException {
     }
 
     protected int select(Selector selector, boolean quickSelect) throws IOException {
-        return select(selector);
+        if (quickSelect) {
+            return SelectorUtil.select(selector, 0L);
```
jitsni: Should we pass QUICK_SELECT_TIMEOUT instead of 0L

PR #753, opened 2016-10-20
Http acceptor multi factor acceptor tests

dpwspoon:
- Added acceptor tests
- Refactored HttpRealmInfo to remove authorization mode

### DIFF
--- a/resource.address/http/src/main/java/org/kaazing/gateway/resource/address/http/HttpRealmInfo.java
+++ b/resource.address/http/src/main/java/org/kaazing/gateway/resource/address/http/HttpRealmInfo.java
@@ -24,7 +24,6 @@ import org.kaazing.gateway.security.LoginContextFactory;
 public class HttpRealmInfo {
 
     private final String name;
-    private final String authorizationMode;
     private final String challengeScheme;
     private final String description;
     private final String[] headerNames;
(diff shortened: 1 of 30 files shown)


transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpLoginSecurityFilter.java
```diff
@@ -373,6 +374,18 @@ protected boolean login(NextFilter nextFilter,
                     }
                     // Login was successful, but the subject identified does not have required roles
                     // to create the web socket.  We re-issue a challenge here.
```
jfallows: This comment needs to move below codeblock, right?

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpLoginSecurityFilter.java
```diff
@@ -373,6 +374,18 @@ protected boolean login(NextFilter nextFilter,
...
+                        loginContexts[realmIndex] = loginContext;
+
+                        // remember login context
+                        httpRequest.setLoginContext(loginContext);
+
+                        // remember subject
+                        httpRequest.setSubject((loginContext == null || loginContext == LOGIN_CONTEXT_OK) ? subject : loginContext.getSubject());
+//                        realmIndex++;
+                        return true;
+                    }
```
jfallows: Does `resultType` `SUCCESS` mean that all factors have succeeded or just the current factor?

If it means only the current factor has succeeded, then this seems to be returning `true` eagerly, even before all factors have been processed, right?

resource.address/http/src/main/java/org/kaazing/gateway/resource/address/http/HttpRealmInfo.java
```diff
@@ -24,7 +24,6 @@
 public class HttpRealmInfo {
```
jfallows: Recommend splitting this into a `HttpRealmInfo` interface with getters only and a `DefaultHttpRealmInfo` implementation with setters and getters in an internal package, then mock interface from tests.

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpLoginSecurityFilter.java
dpwspoon: fixed

PR #755, opened 2016-10-20
Sync with netty 3.5.10

cmebarrow:
I went through  all of our classes in mina.netty which are copies of classes from Netty, checking if all changes that were made in Netty between releases 3.6.3 (from which some of our classes were originally copied) and 3.10.5 (which we are now using as a dependency) have been applied to our copies. Some changes have not been applied, so this PR applies them. It also adds a mina.netty/README.txt file giving a summary of differences between our copies and Netty 3.10.5.

Note that I did not fully apply the changes to AbstractNioSelector and AbstractNioChannel. Comments in the code explain why (there was a risk of degrading performance).

If you diff our classes with Netty 3.10.5, the differences you see now consist only of changes we actually need to adapt them for Kaazing usage.

### DIFF
--- a/mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioSelector.java
+++ b/mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioSelector.java
@@ -177,7 +177,7 @@ abstract class AbstractNioSelector implements NioSelector {
         }
 
         try {
-            newSelector = Selector.open();
+            newSelector = SelectorUtil.open();
         } catch (Exception e) {
             logger.warn("Failed to create a new Selector.", e);
             return;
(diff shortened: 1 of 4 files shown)


mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioSelector.java
```diff
@@ -247,6 +247,11 @@ public void run() {
                 int workCount = 0;
                 long beforeSelect = System.nanoTime();
                 int selected = select(selector, quickSelect);
+                // The SelectorUtil.EPOLL_BUG_WORKAROUND condition was removed in Netty 3.10.5 and instead
+                // added to the if (selectReturnsImmediately == 1024) condition later on. This seems inefficient
+                // for the (common) case where the workaround is not enabled since in that case there's no point
+                // in looping through the selector keys, so we are keeping the condition here. There's no risk
+                // of a busy loop (https://github.com/netty/netty/issues/2426) when the workaround is not enabled.
```
cmebarrow: Note that the latest netty code (4.1, class http://netty.io/4.1/xref/io/netty/channel/nio/NioEventLoop.html) does now always rebuild the selector if select returns too "quickly" 512 times. But the current Netty 3.10.5 code only ever rebuilds the selector if EPOLL_BUG_WORKAROUND is enabled (see https://github.com/netty/netty/blob/netty-3.10.5.Final/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioSelector.java#L262).

mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioSelector.java
```diff
@@ -247,6 +247,11 @@ public void run() {
...
                 if (SelectorUtil.EPOLL_BUG_WORKAROUND && selected == 0 && !wakenupFromLoop && !wakenUp.get()) {
```
jitsni: I don't mind removing this whole if condition. I don't think we will enable EPOLL_BUG_WORKAROUND. My understanding is that this for very old JDKs

jitsni:
Looks good to me.

cmebarrow:
Replaced by PR https://github.com/kaazing/gateway/pull/757 since we need these changes (and especially the selector wakeup fix which will be done on top of them) in 5.3.1.

Issue #752, opened 2016-10-19
UDP processing may be slowed by select, we should call selector.wakeup when udp events come in


cmebarrow:
PR https://github.com/kaazing/gateway/pull/746 made a change to have us always do selector.select(10) in the NioWorker run method (actually, AbstractNioSelector.run). This will cause us to block for 10ms if there are no TCP events coming in, even if UDP events are coming in. Before the change we were using selectorNow (no delay) but introducing delay using idleStrategy.idle(workCount); if no events were processed. That had the unfortunate effect of adding a fixed 10ms delay after a certain number of run() loop iterations did no work, which caused a performance regression for TCP (QA-Roadmap#191). However, the selector.select(10) approach  will now  immediately block for 10ms for UDP case where there are UDP events coming in but there are no TCP events. In circumstances where UDP packets are coming in very closely spaced in time but not touching this may result in adding an extra 10 ms of latency compared with the prior use of idleStrategy.

The best way to fix this is probably to arrange for incoming UDP events (from the boss thread) to wake up the selector of the target NioWorker thread. That way we will manage to immediately process both TCP and UDP events. Calling selector.wakeup for every UDP packet may slow things up, so we will need to use the same AtomicBoolean guard as used in AbstractNioSelector.registerTask to reduce the cost to wakeup, and benchmark this approach to make sure it actually does result in an improvement, or at least no regressions. But it's worth pointing out we already do call selector.wakeup very frequently since it is potentially called whenever a task is added to a worker's task queue (see AbstractNioSelector.registerTask).

If it turns out selector.wakeup is adding too much overhead, the alternative is to use an IdleStrategy implementation which calls selector.select in order to idle instead of parkNanos.

cmebarrow:
This was fixed by PR https://github.com/kaazing/gateway/pull/759

PR #759, opened 2016-10-21
mina.netty: use selector.wakeup to speed up UDP event processing

cmebarrow:
This PR is for issue #752.
Added code to wakeup the selector in AbstractNioWorker.messageReceived so we process UDP events immediately. Undid the code changes to return workCount from the various process methods, and restored the logic to use quickSelect if processTasks(deadline) finds there are still tasks waiting to complete.


cmebarrow:
Closing because accidentally created to merge into develop instead of release/5.3.1

PR #760, opened 2016-10-21
5.3.1 nioworker perf fix selector wakeup

cmebarrow:
This PR is for issue #752.
Added code to wakeup the selector in AbstractNioWorker.messageReceived so we process UDP events immediately. Undid the code changes to return workCount from the various process methods, and restored the logic to use quickSelect if processTasks(deadline) finds there are still tasks waiting to complete.

@jitsni Please review and merge this PR after merging #752 since it includes those changes.


mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/NioClientBoss.java
```diff
@@ -90,15 +90,13 @@ protected Runnable createRegisterTask(Channel channel, ChannelFuture future) {
     }
 
     @Override
-    protected int process(Selector selector) {
+    protected void process(Selector selector) {
```
jitsni: May be we can removed this class so netty's class is picked up

PR #761, opened 2016-10-25
Fixing udp server boss thread names

jitsni:


mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/NioServerDatagramBoss.java
```diff
@@ -55,7 +57,7 @@
  * A class responsible for registering channels with {@link Selector}.
  * It also implements the {@link Selector} loop.
  */
-public class NioServerDatagramBoss extends AbstractNioWorker implements Boss {
+public class NioServerDatagramBoss extends AbstractNioWorker /* AbstractNioSelector */ implements Boss {
```
cmebarrow: What's the purpose of the comment /* AbstractNioSelector */ ? Might be better to remove it

jitsni: We want to extend AbstractNioSelector instead of AbstractNioWorker. But that will be done some time in future.

PR #762, opened 2016-10-25
Configuring mina.netty logging factory

jitsni:
Configuring mina.netty logging factory so that it can be controlled by log4j config file


server/src/main/java/org/kaazing/gateway/server/GatewayCommandLineProcessor.java
```diff
@@ -97,6 +99,9 @@ private void launchGateway(String[] args, Properties properties) {
         final Gateway gateway = GatewayFactory.createGateway();
         gateway.setProperties(properties);
 
+        // Configure mina.netty/netty logging so that we can control logging using log4j-config.xml
+        InternalLoggerFactory.setDefaultFactory(new Slf4JLoggerFactory());
```
jitsni: May be just copy InternalLoggerFactory to mina.netty and change the default from JDK one to sl4j. That may be better since we don't have to change in Gateway.

cmebarrow: Yes, I think that would be better

jitsni:
Updated the PR by copying netty class into mina.netty

dpwspoon:
This looks good to me!

jitsni:
Updated README.txt and added a comment in the code

cmebarrow:
@jitsni Looks good. The built hung somewhere in http transport tests, restarted. When it's green I'll merge.

PR #764, opened 2016-10-26
Added support for configuring multiple http realms

dpwspoon:


server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultServiceContext.java
```diff
@@ -680,83 +686,87 @@ private void injectServiceOptions(String transportURI, Map<String, Object> optio
...
-                            throw new IllegalArgumentException(
-                                    tcpRealmOptionName + " needs to be the same as the configured realm");
-                        }
-                        options.put(format(optionPattern, "loginContextFactory"), serviceRealmContext.getLoginContextFactory());
-                    }
-                }
+        final String httpRealmOptions = (String) options.remove("http.realm");
+
+//        TODO DPW
+//        if (serviceRealmContext == null &&
```
dpwspoon: This is a check we had, but there is no matching tests.  I think I will remove it

transport/wsn/src/test/java/org/kaazing/gateway/transport/wsn/auth/EchoServiceAuthorizationIT2.java
```diff
@@ -0,0 +1,117 @@
...
+                    .done();
+            // @formatter:on
+            init(configuration);
+        }
+    };
+
+    @Rule
+    public TestRule chain = createRuleChain(gateway, robot);
+
+    @Specification("basic.2factor.authorized.access.with.valid.credentials")
```
jitsni: Do you want to add a test case for the following scenario :
first realm valid credentials, second realm *invalid* credentials

dpwspoon: That and other cases are already tested [here](https://github.com/kaazing/gateway/tree/develop/transport/http/src/test/java/org/kaazing/gateway/transport/http/multi/auth).  The reason I added this test was for end to end configuration, i.e. showing that it was possible to configure from the gateway configuration and have it show up on the resource address correctly.  I wasn't aware / am not aware of other tests that do this in a more precise manner...

dpwspoon:
Closing in favor of https://github.com/kaazing/gateway/pull/765

PR #771, opened 2016-11-01
(transport/nio) Build NIO transport test jar so other projects can make use of testware…

cmebarrow:
...like TcpConnectorRule.

### DIFF
--- a/transport/nio/pom.xml
+++ b/transport/nio/pom.xml
@@ -117,6 +117,27 @@
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


jitsni:
@cmebarrow where do you plan to use this dependency ? Then changing something (for e.g. TcpConnectorRule) in this artifact may break the component that uses this dependency.

cmebarrow:
@jitsni In enterprise.gateway, in two of the ITs for the jms service.

cmebarrow:
Closing to avoid Jitu's objection (introducing a test dependency on the way and forcing publishing of a test jar from gateway). Will duplicate TcpConnectorRule class instead.

PR #772, opened 2016-11-02
Udp benchmarks

jitsni:


transport/nio/src/test/java/org/kaazing/gateway/transport/udp/UdpConnectorBM.java
```diff
@@ -0,0 +1,344 @@
...
+        IoBufferAllocatorEx<?> allocator = session.getBufferAllocator();
+        CountDownLatch latch = new CountDownLatch(1);
+        state.handler.setLatch(latch);
+        session.write(allocator.wrap(data));
+        latch.await(1, TimeUnit.SECONDS);
+
+        clientSent.incrementAndGet();
+    }
+
+    // Synchronous client using DatagramSocket client
```
cmebarrow: So this is pure udp with no Kaazing? How does it compare to our UDP connector?

transport/nio/pom.xml
```diff
@@ -123,6 +133,34 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
             </plugin>
+<!-- Enable to build benchmarks.jar after there is new version of plugin
```
cmebarrow: What's the plan for getting this to work? Would be worth adding more comment to explain, and say how it can be run now if folks want to run it (maybe QA).

transport/nio/src/test/java/org/kaazing/gateway/transport/udp/UdpConnectorBM.java
jitsni: I added a benchmark by running udp connector synchronously (after every sent message it waits for the echoed message before continuing further). IIRC, no kaazing client works slightly better.

transport/nio/pom.xml
jitsni: @jfallows will create a PR against maven plugins. Until then, just run in the IDE.

PR #774, opened 2016-11-03
Ticket#434 fix frame masking check

danibusu:


danibusu:
Closing in order to fix ignored test.

PR #775, opened 2016-11-03
Ticket#434 fix frame masking check

danibusu:


transport/ws/src/main/java/org/kaazing/gateway/transport/ws/bridge/filter/WsFrameDecoder.java
```diff
@@ -320,13 +321,13 @@ private static void validateWireCloseCode(int statusCode) throws ProtocolDecoder
...
     /**
-     * Checks if the gateway is acting as a client in a websocket connection
-     * with a server that just sent a masked frame (non-compliant with RFC 6455),
-     * in order to be able to terminate the connection.
+     * Checks if masking is allowed/expected for the frame being decoded.
      */
-    private static void validateMaskingAllowed(IoSession session, boolean masked) throws ProtocolDecoderException {
-        if (masked && session.getService() instanceof IoConnector) {
+    private void validateMaskingAllowed(boolean masked) throws ProtocolDecoderException {
+        if (masked && !maskingExpected) {
```
jitsni: How about xor,  masked ^ maskingExpected ?

jitsni:
Just a minor comment, otherwise looks good to me.

transport/ws/src/main/java/org/kaazing/gateway/transport/ws/bridge/filter/WsFrameDecoder.java
danibusu: I changed the `^` with `!=`, seeing how they behave the same in this case and thought it would improve readability. Hope that's ok. Thanks!

danibusu:
@jitsni Implemented your feedback and added a reply to you comment. Let me know if anything else is required. Thanks!

PR #778, opened 2016-11-04
update HTTP request parsing state machine to validate partial method

danibusu:


mina.netty/src/main/java/org/kaazing/mina/filter/codec/statemachine/ConsumeToDynamicTerminatorDecodingState.java
```diff
@@ -110,10 +110,25 @@ public DecodingState decode(IoBuffer in, ProtocolDecoderOutput out)
...
+     * Validate partially received message, from {@code startPos} to the buffer's limit, in
+     * order to fail fast if what was received does not comply with what the current state
+     * is waiting for.
+     * 
+     * @param buffer the buffer containing the partial message 
+     * @param startPos the starting position inside the buffer
+     * @throws Exception if the validation fails or if any problems are encountered in the
+     *         validation process
+     */
+    protected void validate(IoBufferEx buffer, int startPos) throws Exception {
```
jitsni: I think it may not be worth fixing this way. May be we need to pick different DecodingState impl or not do it. It doesn't seem worth to introduce all the complexity in the parsing in HTTP method.

danibusu: @jitsni If you decide not to fix it, I can change the script to send something else other than "Hello", so it passes. Let me know how you want to proceed. Thanks!

jitsni:
@danibusu I would contend that k3po script is wrong. Since it is not a proper HTTP request there is no point in sending HTTP response. Ideally, gateway should timeout during parsing and close the underlying connection. Let us get the k3po script fixed (or you can file a bug on httpx spec)

PR #782, opened 2016-11-09
Using a queue instead of ring buffer to pass messages to udp child channels

jitsni:
Replacing ring buffer with Agrona's OneToOneConcurrentArrayQueue. I think ring buffer is not necessary for this scenario and the code becomes simpler: no conversions among channel buffer/agrona buffer/byte buffer.

Though, I haven't seen much improvement in the benchmarks (it is slightly better or the same).


mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioWorker.java
```diff
@@ -663,15 +645,10 @@ public void run() {
...
-        ByteBuffer byteBuffer = buf.toByteBuffer();
-        atomicBuffer.wrap(byteBuffer);
-        // If there is no space in the ring buffer,  the message would be dropped (ok for UDP)
-        boolean written = ringBuffer.write(channel.getId(), atomicBuffer, 0, atomicBuffer.capacity());
-        if (LOGGER.isDebugEnabled() && !written) {
-            LOGGER.debug(String.format("Message %s for channel %s is not written to ring buffer", message, channel));
+        UpstreamMessageEvent event = new UpstreamMessageEvent(channel, message, null);
+        boolean written = readQueue.offer(event);
+        if (!written && LOGGER.isDebugEnabled()) {
+            LOGGER.debug(String.format("Message %s for channel %s is not written to buffer", message, channel));
```
jfallows: Is there any practical benefit to logging this information?

cmebarrow: In case change it to info level (or even warn)

mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioWorker.java
```diff
@@ -94,20 +85,14 @@
     protected final SocketReceiveBufferAllocator recvBufferPool = new SocketReceiveBufferAllocator();
     protected final SocketSendBufferPool sendBufferPool = new SocketSendBufferPool();
     private final DefaultWriteCompletionEventEx writeCompletionEvent = new DefaultWriteCompletionEventEx();
-
-    private final OneToOneRingBuffer ringBuffer;
-    private final Int2ObjectHashMap<Channel> channels;
-    private final AtomicBuffer atomicBuffer = new UnsafeBuffer(new byte[0]);
+    private final Queue<UpstreamMessageEvent> readQueue = new OneToOneConcurrentArrayQueue<>(UDP_CHANNEL_BUFFER_SIZE_PER_WORKER);
```
cmebarrow: What is UDP_CHANNEL_BUFFER_SIZE_PER_WORKER now intended to mean? Maybe the name should be changed to something like UDP_CHANNEL_READ_QUEUE_SIZE_PER_WORKER

mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioWorker.java
```diff
@@ -682,33 +659,18 @@ public void messageReceived(AbstractNioChannel<?> channel, Object message) {
...
+                channel.getPipeline().sendUpstream(event);
+            }
         }
-        ringBuffer.read(this::handleRead);
     }
 
-    private void handleRead(int msgTypeId, DirectBuffer buffer, int index, int length) {
-        Channel childChannel = channels.get(msgTypeId);
-        if (childChannel == null) {
-            // register task may still be in the task queue, so this gives the register task a chance
```
cmebarrow: Do we still need this logic?

mina.netty/src/main/java/org/kaazing/mina/netty/config/InternalSystemProperty.java
```diff
@@ -35,7 +35,7 @@
 
     // A worker is serving multiple UDP child channels and they share an Agrona buffer.
     // This configures the buffer size in bytes. It must be a positive power of 2
```
cmebarrow: It doesn't have to be a positive power of 2, Agrona will round it up to the next highest power of 2 if it isn't. Change the comment to say "It will be rounded up to the next highest power of 2".

jitsni: Changed it. Also removed register/deregister of child channels

Issue #783, opened 2016-11-10
JMX management service: closeSessions (by principal) service bean method causes NPEs


cmebarrow:
This is reproducible sporadically in JmxSessionPrincipalIT.shouldKillSessionsByRolePrincipal.  Running with debugging you can see the following NPE thrown when the session is attempted to be fetched via a value from a <Principal, SessionId> map via `serviceManagementBean.removeSessionManagementBean(sessionId)`.  I believe the serviceManagementBean has not been populated with the new session yet returning the NPE.  Guarding against the NPE (by not doing anything if it is null) will result in the session not being closed

```
1825 org.kaazing.gateway.management.session.SessionManagementBeanImpl [management-1] WARN  SessionManagementBeanImpl  - Error during doSessionClosed session listener notifications:
java.lang.NullPointerException
	at org.kaazing.gateway.management.jmx.JmxSessionManagementListener.doSessionClosed(JmxSessionManagementListener.java:67)
	at org.kaazing.gateway.management.session.SessionManagementBeanImpl$2.run(SessionManagementBeanImpl.java:287)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
...
```

cmebarrow:
The issue probably is that member variable sessionBeanMap in JmxManagementServiceHandler is not thread safe. It is initialized as follows: `sessionBeanMap = new HashMap<>();`
Debugging in the test confirms that member variable and serviceBeanMap exist from multiple threads.
Solution would be to use ConcurrentHashMap instead for both of these variables.
I have make this change in my [management_fixes](https://github.com/cmebarrow/gateway/tree/management_fixes) branch.

cmebarrow:
[...]

Issue #773, opened 2016-11-02
management.jmx service:  user principals are not visible in jconsole in our mxbeans unless concrete class is named in gateway config


cmebarrow:
To reproduce:
- Install Gateway (community edition)
- Make sure the amqp service requires the user to log on and is operational
- Edit gateway-config.xml to add the following in the demo realm definition:
```
     <user-principal-class>
        org.kaazing.gateway.security.auth.config.UserConfig
     </user-principal-class>
```
- Edit gateway-config.xml to add the management.jmx service:
```
<service>
  <name>JMX Management</name>
  <description>JMX management service</description>

  <type>management.jmx</type>

  <properties>
    <connector.server.address>jmx://${gateway.hostname}:2020/</connector.server.address>
  </properties>

  <realm-name>demo</realm-name>

  <authorization-constraint>
    <require-role>ADMINISTRATOR</require-role>
  </authorization-constraint>
</service>
```
- Edit gateway-config.xml to require login on a service (for example, AMQP or echo)
- Start the Gateway
- Use the Javascript demo to establish a client connection to the AMQP or echo service logging on as joe/welcome
- In jconsole, examine the corresponding session MBean attributes. Observe that the Principles attribute is blank
- Workaround is to specify the concrete class name in the gateway-config.xml  as the user identity class: `org.kaazing.gateway.security.auth.config.parse.DefaultUserConfig`

If you edit log4j-config.xml to set info level for logger like transport.http (or just transport) and repeat the test you can see log messages including user identity. So for logging purposes we do correctly detect the user principal class.

cmebarrow:
This issue  is happening because in method org.kaazing.gateway.management.session.SessionManagementBeanImpl.doSessionCreated() we only finding a match if the principal class name is exactly that specified in the gateway configuration:
```
if (userPrincipalClasses.contains(principalClass))
```

By contrast, for logging (in HttpIdentityResolver) we call Subject.getPrincipals(user-principal Class) which return principals which are instances of (implement or extend or are) the user principal class. This is much more user-friendly. We should something similar in SessionManagementBeanImpl.doSessionCreated(), for example use searchedForClass.isAssignableFrom(candidatePrincipal).

cmebarrow:
Related issue: https://github.com/kaazing/gateway/issues/709

cmebarrow:
Closed, fixed by https://github.com/kaazing/gateway/pull/784

PR #784, opened 2016-11-10
Management fixes

cmebarrow:
This PR addresses the following two issues:
- [user principals are not visible in jconsole in our mxbeans unless the concrete implementation class is named in gateway config #773](https://github.com/kaazing/gateway/issues/773).  Fixed as follows:
  - In ServiceManagementBean, DefaultServiceManagementBean constructor, load the configured user principal classes and sstore them as class instances instead of string values
  - In SessionManagementBeanImpl.doSessionCreated(), store as the user principals for the session all of the Principals from the Subject which are **instances of** any of the configured user principal classes
  - In ServiceMXBeanImpl.closeSessions(String, String), use isAssignableFrom to match the passed in principal class against the actual principles on the session
- [JMX management service: closeSessions (by principal) service bean method causes NPEs #783](https://github.com/kaazing/gateway/issues/783)
  - Fixed JmxManagementServiceHandler to use ConcurrentHashMap instead of HashMap for two members which are accessed from multiple threads. This is a needed code change but further changes will be required to fully fix the bug so I am leaving it open.

I also removed some dead code from a few classes.


cmebarrow:
@sbadugu Please aassign some suitable to review this PR, preferably someone who has worked a bit with our MX beans monitoring stuff (management.jmx service).

management/src/main/java/org/kaazing/gateway/management/session/SessionManagementBeanImpl.java
```diff
@@ -202,7 +187,7 @@ public boolean areNotificationsEnabled() {
 
     @Override
     public void incrementExceptionCount() {
-        exceptionCount++;
```
dpwspoon: This use to be reported in JMX / SNMP if I recall correctly?  (maybe it was never implemented).  Presumably this would break that?

management/src/test/java/org/kaazing/gateway/management/monitoring/MonitoringFileIT.java
```diff
@@ -38,10 +38,11 @@
 import org.kaazing.gateway.server.test.GatewayRule;
 import org.kaazing.gateway.server.test.config.GatewayConfiguration;
 import org.kaazing.gateway.server.test.config.builder.GatewayConfigurationBuilder;
+import org.kaazing.gateway.util.InternalSystemProperty;
 import org.kaazing.k3po.junit.annotation.Specification;
 import org.kaazing.k3po.junit.rules.K3poRule;
 
-public class JmxSessionPrincipalIT {
+public class MonitoringFileIT {
```
dpwspoon: Do we want this file now?

dpwspoon: It seem like the test was moved elsewhere

dpwspoon:
Looks good to me other than question above

management/src/test/java/org/kaazing/gateway/management/monitoring/MonitoringFileIT.java
cmebarrow: Good catch, I added this file accidentally. Now removed.

cmebarrow:
@dpwspoon I have now removed that file (MonitoringFileIT) because it should not have been included. Thanks for catching that. Should be good to merge now.

management/src/test/scripts/org/kaazing/gateway/management/jmx/wse.session.with.user.principal.joe.rpt
```diff
@@ -51,10 +51,3 @@ read header "Content-Type" /(?!text\/plain;charset=windows-1252)/
 read header "Connection" "close"
 read [0x01 0x30 0x30 0xFF] # NO-OP
 
-read notify JOE_WSE_SESSION_ESTABLISHED
-
-# read close frames
```
dpwspoon: Perhaps the intention of these close frames are for when it it is killed that it sends them??

dpwspoon:
@cmebarrow 

Test failure related to theses tests on travis:

```
Failed tests: 
  JmxSessionPrincipalsSupertypeIT.shouldKillSessionsByRolePrincipalSupertype:219 Not all sessions have been closed expected:<0> but was:<1>
```

management/src/test/scripts/org/kaazing/gateway/management/jmx/wse.session.with.user.principal.joe.rpt
cmebarrow: This is covered by our wse spec tests.  Those verify that the close works properly when the session is closed on the accept (server) side. We do not need to reverify that in this test. I took out the close frames so that the script  could be reused in sessionAttributePrincipalsShouldListUserPrincipals without having to make that test duplicate the logic of closing the sessions.

cmebarrow:
The test failure is a test issue. The test is not waiting for the three sessions to be open before calling closeSessions. The last session is actually  opening after closeSessions was called. Test needs to wait for three sessions to be open before  calling closeSessions.

PR #785, opened 2016-11-10
UDP child channel write future is fired as success always

jitsni:
Not propagating updates from parent channel's write future to child channel
write future. Since it is UDP, the child channel's write future is fired
as success.

### DIFF
--- a/mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/NioChildDatagramPipelineSink.java
+++ b/mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/NioChildDatagramPipelineSink.java
@@ -71,25 +71,18 @@ class NioChildDatagramPipelineSink extends AbstractNioChannelSink {
                     break;
             }
         } else if (e instanceof MessageEvent) {
-            // Making sure that child channel WriteFuture is fired on child channel's worker thread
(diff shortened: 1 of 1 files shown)


mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/NioChildDatagramPipelineSink.java
```diff
@@ -71,25 +71,18 @@ public void eventSunk(final ChannelPipeline pipeline, final ChannelEvent e)
...
 
             // Write to parent channel
             NioDatagramChannel parentChannel = (NioDatagramChannel) childChannel.getParent();
             boolean offered = parentChannel.writeBufferQueue.offer(parentMessageEvent);
             assert offered;
             parentChannel.worker.writeFromUserCode(parentChannel);
+
+            // No need to propagate parentMessageEvent.getFuture() to childFuture
+            // as UDP is unreliable.
+            childFuture.setSuccess();
```
cmebarrow: From John's comment I think you may need to schedule this onto the child io thread, something like 
```
childChannel.getWorker().executeInIoThread(() -> {
                        childFuture.setSuccess();
});
```

PR #790, opened 2016-11-15
Issue 783: NullPointerExceptions after calling closeSessions on ServiceMXBean

cmebarrow:
This PR fixes [JMX management service: closeSessions (by principal) service bean method causes NPEs #783](JMX management service: closeSessions (by principal) service bean method causes NPEs #783).

It also contains the changes from PR https://github.com/kaazing/gateway/pull/784 so better to wait for that to be merged before reviewing this one.

The fix is to remove the ManagementContext.removeSessionManagementBean method (which was previously called from FullManagementFilterStrategy.doSessionClosed and called managementServiceHandler.removeSessionManagementBean) and instead call a new method JmxManagementServiceHandler.removeSessionMXBean from JmxSessionManagementListener


dpwspoon:
LGTM, waiting for travis

PR #791, opened 2016-11-15
(management) Change junit and (especially) jmock-junit dependencies …

cmebarrow:
to be of test scope so consumers of management do not find themselves with these coming in as transient dependencies.


dpwspoon:
Looks good to me, waiting for travis

PR #792, opened 2016-11-15
Remove sigar dependency, remove system MBeans

cmebarrow:
This is for issue#781. 

- Removed the dependency on Sigar.
- Removed all management system classes and beans since they no longer provide any value now we have removed sigar. This means the System (MBean) node and child nodes is no longer visible in jconsole. 
- Cleaned up all compilation warnings in the management code.


dpwspoon:
LGTM

dpwspoon:
travis failure... I think it is unrelated to this change: https://travis-ci.org/kaazing/gateway/builds/176209485

cmebarrow:
Test failure was due to an error in the test (not due to this PR).

PR #797, opened 2016-11-16
Case sensative bug

dpwspoon:
Ready to merge


dpwspoon:
See commits, one did white spacing fix

server/src/test/java/org/kaazing/gateway/server/config/parse/ServiceDefaultsTest.java
```diff
@@ -69,8 +69,8 @@ public void testDefaultConnectOptions() throws Exception {
         Assert.assertTrue("TLSv1".equals(sslProtocols[0]));
 
         Assert.assertTrue("en0".equals(connectOptionsMap.get("udp.interface")));
-        System.out.println(connectOptionsMap.get("tcp.transport"));
-        Assert.assertTrue("socks://localhost:8000".equals(connectOptionsMap.get("tcp.transport").toString().trim()));
+        final String trim = connectOptionsMap.get("tcp.transport").toString().trim();
```
jitsni: rename trim to tcpTransport, also remove final (I feel final for local variables is just noise).

resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/Comparators.java
```diff
@@ -134,7 +135,7 @@ public int compare(ResourceAddress addr1, ResourceAddress addr2) {
             String authority1 = location1.getAuthority();
             String authority2 = location2.getAuthority();
             
-            int compareAuthority = STRING_COMPARATOR.compare(authority1, authority2);
+            int compareAuthority = CASE_INSENSITIVE_ORDER.compare(authority1, authority2);
```
jitsni: Do you want to wrap CASE_INSENSIVE_ORDER in compareComparable ? STRING_COMPARATOR is doing that and it further wraps in compareNonNull (so it protects from null authority values)

resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/uri/URIUtils.java
```diff
@@ -622,5 +622,63 @@ private String buildURIFromTokens(String scheme, String authority, String path,
...
+                try {
+                    canonicalURI = buildURIAsString(scheme, getUserInfo(uri), host == null ?
+                            null : host.toLowerCase(), getPort(uri), path, getQuery(uri), getFragment(uri));
+                } catch (URISyntaxException ex) {
+                    throw new IllegalArgumentException("Invalid URI: " + uri + " in Gateway configuration file", ex);
+                }
+            }
+        }
+        return canonicalURI;
+    }
```
jitsni: Looks like this code is just getting moved to this file from GatewayContextResolver.

dpwspoon: I can put it back if you think its better there (or in a different Util class in gateway.server...?  DefaultServiceContext now uses this, and originally I was thinking the ServiceContext / other  stuff in service module needed access to this method, which is why I moved it...

jitsni:
Can you run enterprise.gateway once with this changeset (because comparators change affects things drastically).

PR #804, opened 2016-11-29
Hazelcast upgrade

danibusu:


server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultClusterContext.java
```diff
@@ -992,17 +759,48 @@ private void initializeCluster(Config config) {
...
             instanceKeyMap.put(localNodeId, localInstanceKey);
-            instanceKeyMap.addEntryListener(instanceKeyEntryListener, true);
 
             IMap<String, Collection<String>> balancerMap = collectionsFactory.getMap(BALANCER_MAP_NAME);
-            balancerMap.addEntryListener(balancerMapEntryListener, true);
+            addBalancerMapEntryListeners(balancerMap);
         }
     }
 
+    private void addBalancerMapEntryListeners(IMap<String, Collection<String>> balancerMap) {
```
danibusu: This is what I meant when I said that the logging statements are still there. I removed the firing of the BalancerMapListeners from this Hazelcast map listener, but left the logging of changes.

service/spi/src/main/java/org/kaazing/gateway/service/cluster/ClusterContext.java
```diff
@@ -45,31 +45,13 @@
...
-    void send(Object msg, final SendListener listener, MemberId member);
-    void send(Object msg, final SendListener listener, String name);
-    <T> void setReceiver(Class<T> type, ReceiveListener<T> receiveListener);
-    <T> void removeReceiver(Class<T> type);
-
     // event listener
     void addMembershipEventListener(MembershipEventListener eventListener);
     void removeMembershipEventListener(MembershipEventListener eventListener);
 
-    // instanceKey listener
```
dpwspoon: I think it is better to remove them.  @jitsni ?

service/spi/src/test/java/org/kaazing/gateway/service/messaging/collections/MemoryCollectionsFactoryTest.java
```diff
@@ -251,7 +272,7 @@ public void expiringMap() throws Exception {
         map.putIfAbsent("two", "2", 10000, MILLISECONDS);
         map.putIfAbsent("three", "3", 2, MILLISECONDS);
 
-        sleep(5);
+        sleep(10);
```
dpwspoon: I would not expect this test to require a change.  Moreover, I think the `assertEquals("2", map.putIfAbsent("two", "20", 1000, MILLISECONDS));` would fail with this change

dpwspoon:
Presumably we need to update the notice.txt of what is included (specifically the version of HazelCast)

dpwspoon:
This looks ok to me.  @jitsni do you mind taking a look as well?

distribution/src/main/assembly/generic-bin.xml
```diff
@@ -100,7 +100,7 @@
         <include>org.apache.xmlbeans:xmlbeans</include>
         <include>com.hazelcast:hazelcast</include>
         <include>com.hazelcast:hazelcast-client</include>
-        <include>com.hazelcast:hazelcast-cloud</include>
+        <include>com.hazelcast:hazelcast-aws</include>
```
jitsni: why do we need hazelcast's aws dependency ?

jitsni:
Is MessageBufferFactory#createMessageBuffer used ? If not, can we clean it up.

jitsni:
ClusterContext#removeMembershipEventListener() is not used any where. management adds a listener but never removes the listener.

jitsni:
Also, ClusterContext#getLock() is exposing the internal details like lock object

service/spi/src/main/java/org/kaazing/gateway/service/cluster/ClusterContext.java
```diff
@@ -45,31 +44,12 @@
...
-
-    // cluster messaging
-    void addReceiveTopic(String name);
-    void addReceiveQueue(String name);
-    <T> T send(Object msg, MemberId member) throws Exception;
-    <T> T send(Object msg, String name) throws Exception;
-    void send(Object msg, final SendListener listener, MemberId member);
-    void send(Object msg, final SendListener listener, String name);
-    <T> void setReceiver(Class<T> type, ReceiveListener<T> receiveListener);
-    <T> void removeReceiver(Class<T> type);
```
jitsni: Looks like a lot is cleaned up here.

distribution/src/main/assembly/generic-bin.xml
danibusu: It is used for auto-discovery of cluster members on AWS. The gateway had a dependency for `hazelcast-cloud`, which was renamed to `hazelcast-aws`, so I assumed that this feature was needed. If you know that auto-discovery is not required, then I can remove it.

danibusu:
@jitsni I removed the rest of the support for cluster messaging, including `MessageBufferFactory`.
I left the `ClusterContext.removeMembershipEventListener()` method, even though it is not used, as I thought that it would be weird to have an API that allows for listeners to be added, but not removed. If you think this is not reason enough, I can remove it.

distribution/src/main/assembly/generic-bin.xml
dpwspoon: Auto-discovery is needed via AWS security groups

PR #810, opened 2016-11-30
Cleaned up inner class based on suspicous memory behavior in idle and…

dpwspoon:
… parent sessions

### DIFF
--- a/transport/spi/src/main/java/org/kaazing/gateway/transport/AbstractBridgeSession.java
+++ b/transport/spi/src/main/java/org/kaazing/gateway/transport/AbstractBridgeSession.java
@@ -122,8 +122,7 @@ public abstract class AbstractBridgeSession<S extends IoSessionEx, B extends IoB
         
         this.service = service;
         this.processor = processor;
-        final AtomicReference<IoSessionEx> parentReference = new AtomicReference<>(parent);
-        this.parent = parentReference;
+        this.parent = new AtomicReference<>(parent);
         this.allocator = allocator;
 
         this.metadata = service.getTransportMetadata();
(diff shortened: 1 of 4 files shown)


transport/spi/src/main/java/org/kaazing/gateway/transport/AbstractBridgeSession.java
```diff
@@ -122,7 +122,8 @@ private AbstractBridgeSession(int ioLayer,
         
         this.service = service;
         this.processor = processor;
-        this.parent = new AtomicReference<>(parent);
+        final AtomicReference<IoSessionEx> parentReference = new AtomicReference<>(parent);
```
jitsni: you don't require parentReference. Just pass in parent to BridgeSessionConfigEx

dpwspoon:
@jitsni Fixed, thanks.  And added a few related fixes

transport/ssl/src/main/java/org/kaazing/gateway/transport/ssl/SslConnector.java
```diff
@@ -124,7 +124,7 @@ public void setSecurityContext(SecurityContext securityContext) {
             // avoid caching SSLSession in shared SSLContextFactory instance
             // Note: SSLSessionContext.setSessionCacheSize(0) means unlimited,
             // so we use 1 instead
-            sslContextFactory.setServerSessionCacheSize(1);
+            sslContextFactory.setClientSessionCacheSize(1);
```
jitsni: Do you want to move the ssl change into a separate PR ?

dpwspoon: sure, will do

dpwspoon: It is now here: https://github.com/kaazing/gateway/pull/813.  I'll change this PR request with the other issues to remove the TLS fix

dpwspoon:
Closing in favor of https://github.com/kaazing/gateway/pull/814

PR #811, opened 2016-12-02
Typo in passing parameter to fail safe plugin

jitsni:


AdrianCozma:
@jitsni This is not a typo, I modified it when I added the profile for jacoco(to get code coverage for SonarQube)
From the failsafe plugin doc:
```
Using an alternate syntax for argLine,
@{...}
allows late replacement of properties when the plugin is executed, so properties that have been modified by other plugins will be picked up correctly. 
```

jitsni:
@AdrianCozma Can you paste the failsafe doc link ? intellij is complaining about this while running the tests. I will file a bug if it is standard maven convention.

dpwspoon:
http://maven.apache.org/surefire/maven-surefire-plugin/faq.html#late-property-evaluation

dpwspoon:
@jitsni are you alright with closing this?

PR #812, opened 2016-12-02
Added the `aggregate-add-third-party-mojo` plugin

mgherghe:


mgherghe:
The build passes now (locally) with the template file path set to `${session.executionRootDirectory}/third-party-template.ftl` instead of `${project.basedir}/third-party-template.ftl`, but this only works when building from the root directory, if I try to build only a submodule, the build fails not finding the .ftl file.

I couldn't find a simple way to make this work, without either modifying all the submodules' pom files (to define a `basedir` property in each module that references the root path) or by modifying the plugin's code on how it loads the template file.

dpwspoon:
To me, it would seem like the correct solution would be for the plugin to support loading the ftl from a dependency via the class path.   You could then publish the ftl as an artifact and add it to a dependency to the plugin. 

I've seen several plugins add this support in modern implementations, and I would bet the code change would be fairly straightforward and likely to be accepted as a contribution back to the project.   

One plugin that recently added this was the com.mycilia license plugin for the file that defines the header.  Example of our use case for this is [here](https://github.com/kaazing/community/blob/develop/pom.xml), summarized as

```xml


                <plugin>
                    <!-- To update the license, do: -->
                    <!--     mvn com.mycila:license-maven-plugin:format -->
                    <groupId>com.mycila</groupId>
                    <artifactId>license-maven-plugin</artifactId>
                    <version>2.11</version>
                    <configuration>
                        <!-- this defines the classpath location of the dependency below -->
                        <header>org/kaazing/community/license/COPYRIGHT</header>
                          ...
                    </configuration>
                    <dependencies>
                        <dependency>
                            <!-- 
                               this dependency includes common copyright headers for sources
                               specifically org/kaazing/community/license/COPYRIGHT
                             -->
                            <artifactId>community.license</artifactId>
                            <type>jar</type>
                             <version>develop-SNAPSHOT</version>
                        </dependency>
                    </dependencies>
                </plugin>
```

I would suggest submitting a patch to the open source project with this and explaining that this adds support for multi-module a distributed projects that share the same parent pom.

mgherghe:
I created a [PR](https://github.com/kaazing/community/pull/44) in the community repository, where I added a module with the .ftl file. 
I added the module as a dependecy for the license plugin, with the develop-SNAPSHOT version, but I see that the Travis build fails with the following error:
>[ERROR] Failed to execute goal org.codehaus.mojo:license-maven-plugin:1.10:add-third-party (add-third-party) on project gateway: Execution add-third-party of goal org.codehaus.mojo:license-maven-plugin:1.10:add-third-party failed: Plugin org.codehaus.mojo:license-maven-plugin:1.10 or one of its dependencies could not be resolved: Could not find artifact org.kaazing:community.notice:jar:develop-SNAPSHOT in ossrh-snapshots (http://oss.sonatype.org/content/repositories/snapshots)

pom.xml
```diff
@@ -377,38 +377,30 @@
...
+                    <outputDirectory>target</outputDirectory>
+                    <thirdPartyFilename>NOTICE.txt</thirdPartyFilename>
+                    <fileTemplate>org/kaazing/community/notice/third-party-template.ftl</fileTemplate>
+                </configuration>
+                <dependencies>
+                    <dependency>
+                        <groupId>org.kaazing</groupId>
+                        <artifactId>community.notice</artifactId>
+                        <type>jar</type>
+                         <version>develop-SNAPSHOT</version>
```
mgherghe: The version should be changed after merging [PR](https://github.com/kaazing/community/pull/44) and releasing community.

pom.xml
```diff
@@ -377,38 +377,32 @@
...
             <plugin>
-                <groupId>org.kaazing</groupId>
+                <groupId>org.codehaus.mojo</groupId>
                 <artifactId>license-maven-plugin</artifactId>
-                <version>1.0.0.11</version>
+                <version>1.10</version>
+                <inherited>false</inherited>
+                <configuration>
+                    <outputDirectory>target</outputDirectory>
+                    <thirdPartyFilename>NOTICE.txt</thirdPartyFilename>
```
mgherghe: Should this generated Notice.txt file also be checked with the one from the top level to see if it is up to date?

mgherghe: Should the top level Notice.txt file be checked against this generated one to see if it is up to date?

distribution/pom.xml
```diff
@@ -242,6 +242,58 @@
 
         <plugins>
             <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>license-maven-plugin</artifactId>
+                <configuration>
+                    <outputDirectory>target</outputDirectory>
```
dpwspoon: See maven properties: https://cwiki.apache.org/confluence/display/MAVEN/Maven+Properties+Guide

distribution/pom.xml
```diff
@@ -242,6 +242,58 @@
 
         <plugins>
             <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>license-maven-plugin</artifactId>
```
dpwspoon: Should we specify the version here?

distribution/pom.xml
```diff
@@ -242,6 +242,58 @@
...
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
+                <version>1.4.1</version>
+                <dependencies>
+                    <dependency>
+                        <groupId>org.kaazing</groupId>
+                        <artifactId>community.enforcer</artifactId>
+                        <version>develop-SNAPSHOT</version>
```
dpwspoon: develop-SNAPSHOT here seems like this would eventually break when released.  Maybe you want ${project.version}?   I'm not sure what would happen when it is released if this property would be overridden or it would stay project.version.  So down stream builds of say kaazing/gateway might be looking for their project.verison???

distribution/pom.xml
```diff
@@ -242,6 +242,58 @@
...
+                </dependencies>
+                <executions>
+                    <execution>
+                        <id>default-enforce</id>
+                        <configuration combine.self="override">
+                            <skip>false</skip>
+                            <rules>
+                                <myCustomRule implementation="org.kaazing.community.enforcer.rule.RequireSameFile">
+                                    <originalFilePath>src/main/gateway/NOTICE.txt</originalFilePath>
+                                    <revisedFilePath>target/NOTICE.txt</revisedFilePath>
```
dpwspoon: see mvn properties?

distribution/pom.xml
```diff
@@ -242,6 +242,58 @@
...
+                    </dependency>
+                </dependencies>
+                <executions>
+                    <execution>
+                        <id>default-enforce</id>
+                        <configuration combine.self="override">
+                            <skip>false</skip>
+                            <rules>
+                                <myCustomRule implementation="org.kaazing.community.enforcer.rule.RequireSameFile">
+                                    <originalFilePath>src/main/gateway/NOTICE.txt</originalFilePath>
```
dpwspoon: suggest putting ${build.basedir} in front of source

pom.xml
```diff
@@ -377,38 +377,32 @@
 
         <plugins>
             <plugin>
-                <groupId>org.kaazing</groupId>
+                <groupId>org.codehaus.mojo</groupId>
```
dpwspoon: Same general comments about build properties

distribution/pom.xml
dpwspoon: Oh I see, nvm.  It will be updated on release

pom.xml
dpwspoon: Ideally, I would think it would be. Correct?

distribution/pom.xml
mgherghe: The rule is implemented to only take the relative path to the current project. 
I created [PR](https://github.com/kaazing/community/pull/45) in the community repo with this change. After this is merged and community released, then I can update the paths in the pom to use mvn properties,

mgherghe:
The Travis build is failing because when the maven-license-plugin is trying to create the Notice.txt from the parent module, it forks all the submodules and looks for their dependencies. Since they depend on each other, errors such as the one below appear.
>Could not transfer artifact org.kaazing:gateway.service.turn.rest:jar:develop-SNAPSHOT from/to codehaus-snapshots (https://nexus.codehaus.org/snapshots/): nexus.codehaus.org: Unknown host nexus.codehaus.org

I found the same issue referenced [here](http://stackoverflow.com/questions/23038269/an-error-while-using-the-license-maven-plugin-in-a-multi-module-project).

Locally, the build passed because the .m2 repo already contained the Kaazing modules develop-SNAPSHOT versions, but if I delete all the dependencies, the issue also reproduces locally.

I am not sure how to proceed with this.

mgherghe:
As a workaround for the issue above, I created a different module with all the gateway submodules as dependencies and configured the maven-license-plugin there, instead of the parent pom. 

However, just by adding all the submodules does not generate the same Notice.txt file that would be generated running the `aggregate-add-third-party` goal from the parent pom. The generated file from the newly created submodule doesn't include `optional` or `provided` dependencies. All these should be added manually in the submodule's pom and checked at every build if they are up to date.

This is if the Notice.txt file from the root directory needs to contain all the dependencies that are not included in the distribution. If it is enough to be the same as the one in the distribution, then the license plugin can only be configured in the distribution module.

mgherghe:
@dpwspoon According to the table from this [link](http://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope), for the dependencies with `provided` and `test` scope, their transitive dependencies will be omitted when running either the `dependency:tree` or the `add-third-party` commands from the newly created sub-module.

To workaround this, I manually added these dependencies in the generate.notice pom file and also the dependencies that were declared as optional. This resulted in generating a Notice.txt file similar to the one generated with running `aggregate-add-third-party` from the parent pom. 

The differences between the files are mostly related to duplicate dependencies being either omitted or repeated.

The issue with this approach is that the pom file from this sub-module should always be checked if it is up-to-date. This would probably need to be done by implementing a parser that should go through all the poms from the other sub-modules and check if all the `provided`, `test` and `optional` dependencies are added in this pom. 

I have pushed my changes for this in a [separate branch](https://github.com/mgherghe/gateway/tree/test_submodule_for_notice).

[This](https://github.com/kaazing/gateway/files/685454/NOTICE.txt) is how the generated Notice from the `distribution` looks like and [this](https://github.com/kaazing/gateway/files/685461/NOTICE.txt) is the one generated from the `generate.notice` sub-module.

Do we want to continue with this approach or is it enough to have the file generate from the distribution module in both distribution/src/main/gateway and the root directory ?

NOTICE.txt
```diff
@@ -1,314 +1,765 @@
...
+
+	License:	http://www.apache.org/licenses/LICENSE-2.0.txt (The Apache Software License, Version 2.0)
+	Homepage:	https://github.com/kaazing/community
+
+This product depends on Kaazing Corporation License 2.18
+
+	License:	http://www.apache.org/licenses/LICENSE-2.0.txt (The Apache Software License, Version 2.0)
+	Homepage:	https://github.com/kaazing/community
+
+This product depends on Kaazing Gateway Bridge develop-SNAPSHOT
```
dpwspoon: This develop-SNAPSHOT version seams wrong?

dpwspoon: or we have a problem somewhere with that

NOTICE.txt
```diff
@@ -1,314 +1,765 @@
...
 
-	License:	http://www.apache.org/licenses/LICENSE-2.0 (Apache License, Version 2.0)
+	License:	http://www.apache.org/licenses/LICENSE-2.0.txt (The Apache Software License, Version 2.0)
 	Homepage:	https://github.com/kaazing/gateway.transport.ssl.git
 
-This product depends on WS Transport 1.0
+This product depends on WS Transport develop-SNAPSHOT
 
-	License:	http://www.apache.org/licenses/LICENSE-2.0 (Apache License, Version 2.0)
+	License:	http://www.apache.org/licenses/LICENSE-2.0.txt (The Apache Software License, Version 2.0)
```
dpwspoon: Can we not include kaazing stuff by chance, as this version would change everytime?

distribution/src/main/gateway/NOTICE.txt
```diff
@@ -0,0 +1,425 @@
...
+
+	License:	http://www.json.org/license.html (provided without support or warranty)
+	Homepage:	http://www.json.org/java/index.html
+
+This product depends on Kaazing Corporation License 2.18
+
+	License:	http://www.apache.org/licenses/LICENSE-2.0.txt (The Apache Software License, Version 2.0)
+	Homepage:	https://github.com/kaazing/community
+
+This product depends on Kaazing Gateway Bridge develop-SNAPSHOT
```
dpwspoon: Same question with kaazing stuff

dpwspoon:
@mgherghe I'm ok with this PR, other than the constant changes to kaazing in the notice.txt.  That needs to be fixed

mgherghe:
Created [PR](https://github.com/kaazing/community/pull/46) with changes to exclude Kaazing dependencies. I will update this PR after that one is merged.

mgherghe:
I implemented the feedback changes.
However, the Travis build is failing at the enforcer rules when checking if the Notice.txt file is up-to-date.
This is because the file generated on Travis does not include the shaded dependencies from the mina.netty module. 
**Locally**, this happens when building only the distribution module, but when building the whole project, the Notice.txt file is generated with the shaded dependencies.
I am not clear on why this is happening and I will continue with the investigation.

mgherghe:
The different behaviour between the two builds (the local one and the one from Travis) is because of different versions of Maven.
When runing the `maven-license-plugin`, versions of maven <3.3.x generate the Notice.txt file without the shaded dependencies from the `mina.netty` module. 
However, this only happens from the `package` phase and higher. The lower phases have the same behaviour. 
Also, when running the build with maven 3.2.5, in the `target` folder from the `mina.netty` module there is a `mina.netty-develop-SNAPSHOT-shaded.jar` generated, that isn't generated with newer versions of maven.
I will continue with the investigation tomorrow. 
As a fix for this issue, maybe the enforcer rules that check if the Notice files are the same could be ignored for a maven version < 3.3.1 or change the maven version from the Travis build (currently, it is 3.1.1)

dpwspoon:
We should specify the plugin version here: https://github.com/kaazing/community/blob/e6a890fa8fe322028d4b12e799933a7fefbd1fe3/community.main/pom.xml

mgherghe:
That is the license plugin used to check the license headers. 
I am using `license-maven-plugin` from `org.codehaus.mojo`, and I am specifying the version [here](https://github.com/mgherghe/gateway/blob/update_notice_file/pom.xml#L43). 
However, this seems to behave differently depending on the system's default maven version.

mgherghe:
If the `maven shade` plugin does not generate a `dependency-reduced-pom.xml` file, then building the project with any version of maven has the same behavior in generating the `Notice.txt` file.
Setting the `createDependencyReducedPom` to false fixes the Travis build failure. This is consistent with how the `maven shade` plugin is [configured](https://github.com/search?q=org%3Akaazing+createDependencyReducedPom&type=Code) in other projects

NOTICE.txt
```diff
@@ -1,314 +1,135 @@
...
 
-This product depends on NIO Transport 1.0
+This product depends on Netty 3.10.5.Final
 
 	License:	http://www.apache.org/licenses/LICENSE-2.0 (Apache License, Version 2.0)
-	Homepage:	https://github.com/kaazing/gateway.transport.nio.git
+	Homepage:	http://netty.io/
 
-This product depends on Pipe Transport 1.0
+This product depends on Javassist 3.7.ga
```
dpwspoon: This is a runtime dependency, is it possible to exclude those as we don't depend on them

dpwspoon: nvm, I see it is compile in a few modules

dpwspoon:
Thanks @mgherghe, please remove all                 
```
                <artifactId>license-maven-plugin</artifactId>
```

as this is effectively replacing that logic

dpwspoon:
The notice.txt is missing SNMP4j.  This is perhaps because we had to package it up in maven ourselves so it falls under the kaazing namespace: https://github.com/kaazing/snmp4j.  Please figure out how to include those dependencies in the notice.txt

dpwspoon:
I believe antlr4 and treelayout are only pulled in by k3po.lang.  And k3po.lang should be able to be at scope test for all modules, thereby removing them

Please try to change all instances of 
```
        <dependency>
            <artifactId>k3po.lang</artifactId>
            <version>${k3po.version}</version>
        </dependency>
```

to 

```
        <dependency>
            <artifactId>k3po.lang</artifactId>
            <version>${k3po.version}</version>
            <scope>test</scope>
        </dependency>
```

mgherghe:
Regarding the SNMP4j dependency, I found them on the maven repository, [here](https://mvnrepository.com/artifact/org.snmp4j/snmp4j-agent/1.4.3) is the snmp4j-agent and [here](https://mvnrepository.com/artifact/org.snmp4j/snmp4j/1.11.3) the snmp4j. The build passed locally with the last commit, in which I changed the snmp dependencies from org.kaazing to org.snmp.
However, the Travis build is failing, since it is looking for the dependencies on http://repo.maven.apache.org/maven2, where the versions currently used are not uploaded. (http://repo.maven.apache.org/maven2/org/snmp4j/)

mgherghe:
@dpwspoon I updated the PR after synchronizing with the changes for [#1036](https://github.com/kaazing/roadmap/issues/1036)

PR #816, opened 2016-12-06
Validating Host header when client sends a request in absolute-form

jitsni:
RFC 7230 says "A server MUST respond with a 400 (Bad Request) status code to any
HTTP/1.1 request message that lacks a Host header field and to any
request message that contains more than one Host header field or a
Host header field with an invalid field-value."

From Server's point of view, if there is Host header with invalid field-value,
Http acceptor should send 400 status code.

### DIFF
--- a/service/http.directory/src/test/java/org/kaazing/gateway/service/http/directory/HttpDirectoryServiceIT.java
+++ b/service/http.directory/src/test/java/org/kaazing/gateway/service/http/directory/HttpDirectoryServiceIT.java
@@ -179,11 +179,6 @@ public class HttpDirectoryServiceIT {
     }
 
     // /////////////////// HOST HEADER ///////////////////////
-    @Specification("host/host.empty.header.with.absolute.uri")
-    @Test
-    public void testEmptyHostHeaderWithAbsoluteUri() throws Exception {
-        robot.finish();
-    }
 
     @Specification("host/host.empty.header.with.relative.uri")
     @Test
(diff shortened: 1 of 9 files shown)


service/http.directory/src/test/java/org/kaazing/gateway/service/http/directory/HttpDirectoryServiceIT.java
```diff
@@ -179,11 +179,6 @@ public void testGetForbiddenFileExists() throws Exception {
     }
 
     // /////////////////// HOST HEADER ///////////////////////
-    @Specification("host/host.empty.header.with.absolute.uri")
```
jitsni: Not a valid test. Every request should have Host header.

dpwspoon: Hmm, I think the point of the test comes from this part of the spec:

```
 When a proxy receives a request with an absolute-form of
   request-target, the proxy MUST ignore the received Host header field
   (if any) and instead replace it with the host information of the
   request-target.

```

I think I wrote this test and interpreted the if any to imply that the http acceptor might see requests of this type.  Perhaps this was a mistake and will only be seen from http 1.0?  If so, then maybe we can remove this test from the spec and from here

jitsni: RFC definition of proxy is forwarding proxy. Our http.proxy needs to be interpreted as gateway or server.

PR #819, opened 2016-12-07
.ksn parameter is converted into header only when httpxe is detected.

jitsni:


dpwspoon:
LGTM, assuming the build passes

Issue #676, opened 2016-08-10
test.util dependency should always be test scope


cmebarrow:
The test.util module is only intended for use in tests. However several modules have a compile scope dependency on it. This is incorrect. Dependency scope for test.tutil should always be test. Result is project which are shading gateway jars to form composite jars are incorrectly including the test.util.jar classes. Modules with this fault include:
- transport/spi
- transport/nio
- transport/ws

cmebarrow:
I have a fix available in my issue_676_test_util_scope. Just need to check I didn't miss any modules.

cmebarrow:
Verified all usages in gateway have been corrected, and there are no invalid usages in enterprise.gateway.

PR #826, opened 2016-12-15
Issue 676 test.util scope

cmebarrow:
(#676) test.util module should be used for tests only: make all dependencies on test.util test scope, or (transport.spi) make sure it's not transitive.


cmebarrow:
@dpwspoon This should be merged only after https://github.com/kaazing/enterprise.gateway/issues/569 is merged (since that PR includes changes needed to accommodate this one).

transport/spi/pom.xml
```diff
@@ -45,11 +45,13 @@
         <dependency>
             <groupId>org.jmock</groupId>
             <artifactId>jmock</artifactId>
+            <scope>provided</scope>
```
dpwspoon: It doesn't need to be done now, but we should consider moving these mocks into a test-jar and using that test jar dependency in other builds that need the "matchers".

PR #827, opened 2016-12-15
http.proxy service is changed to accept only with URIs ending with /

jitsni:
This addresses https://github.com/kaazing/tickets/issues/920


service/http.proxy/src/test/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceTest.java
```diff
@@ -36,16 +36,16 @@ public void testCreateService() throws Exception {
         Assert.assertNotNull("Failed to create http.proxy service", service);
     }
 
-    @Test
-    public void shouldFailGatewayStartupAcceptAndConnectDifferentEndingSlashes() throws Exception {
+    @Test(expected = IllegalArgumentException.class)
```
dpwspoon: Not required and maybe not needed here, but consider using [junit expected exception rule](http://junit.org/junit4/javadoc/4.12/org/junit/rules/ExpectedException.html) to assert both the exception type and error message.  I think that would probably eliminate the need for the throw new AssertionError below

dpwspoon:
I believe a "/" path gets injected for an accept that is missing a path, but not for a connect that is missing the path.  This would cause confusion if someone configured the proxy with a 

```
<accept>http://kaazing.com:80</accept>
<connect>http://localhost.com:8080</connect>
```
as the error message would only complain about the connect.  I would suggest confirming with a test, and then changing the behavior of the connect url to also inject the path if it is http.  So effectivly the config above would be valid as the / path is implied.

jitsni:
Updated the PR for second review

dpwspoon:
Thanks @jitsni, LGTM, just waiting for travis to pass

PR #829, opened 2016-12-15
Made http connector follow redirects spec compliant for appending old…

dpwspoon:
… fragment to path, with other minor implementation improvements

PR #833, opened 2016-12-16
Http proxy redirect (not following) occasional test failure fix

dpwspoon:


dpwspoon:
I've pulled in [this related PR](https://github.com/kaazing/gateway/pull/829) into this one (they touch same code and work on very similar issues)

PR #838, opened 2016-12-28
Add logging filter constructor

cmebarrow:
Added a new constructor to LoggerFilter and subclasses to make them easier to use in services. Services can pass in the service type as the transport name in in order to cause service level log messages to be logged, including user identity obtained from the IoSession.

### DIFF
--- a/transport/spi/src/main/java/org/kaazing/gateway/transport/LoggingFilter.java
+++ b/transport/spi/src/main/java/org/kaazing/gateway/transport/LoggingFilter.java
@@ -15,6 +15,8 @@
  */
 package org.kaazing.gateway.transport;
 
+import static java.lang.Thread.currentThread;
+
 import java.io.IOException;
 
 import org.apache.mina.core.filterchain.IoFilterAdapter;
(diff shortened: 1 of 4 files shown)


transport/spi/src/main/java/org/kaazing/gateway/transport/LoggingFilter.java
```diff
@@ -94,6 +97,10 @@ public LoggingFilter(Logger logger) {
         this(logger, "%s");
     }
 
+    public LoggingFilter(Logger logger, IoSession session, String transportName) {
```
jitsni: It seems like this restricts a logging filter to be used per session. Previously, a logging filter could be used across all sessions (though, I didn't find any usages for that). One instance of the filter was *not* shared across the sessions in the current code, so should be fine.

cmebarrow: Yes you are right. This is in line with all current usage, as you pointed out, so I think it is okay. An alternative would be to stash the format on a session attribute. But I think it's more straightforward to have an instance per session.

PR #844, opened 2017-01-05
Added keystore and truststore to LoginModule options

dpwspoon:


jitsni:
Add a testcase with a login module that accesses the new properties.
Also, as a developer there is bit of confusion when developing a login module (two different javadocs, may be two jars in the classpath). We need a clear way for a developer to write a login module and how they look for these LoginModule options.

PR #848, opened 2017-01-06
Make unbind execute asynchronously without using the scheduler.

stanculescu:
The change consists in calling unbindInternal in unbind method without submitting this call to the unbindScheduler and forcing in unbindInternal a call to unbindAsync instead of unbind.

On Windows 10 without these changes, the hang can be reproduced after about 20~30 consecutive runs of the test TcpRealmEarlyAccessFeatureIT.shouldStartWhenTcpRealmAcceptOptionSetAndFeatureEnabled in enterprise gateway.

It also fails when the scheduler is used. When removing it, the test no longer failed.

However there are tests in gateway that fail during the build of which some are sporadic but we are not sure the changes have not introduced regressions.

### DIFF
--- a/mina.netty/src/main/java/org/kaazing/mina/core/service/AbstractIoAcceptorEx.java
+++ b/mina.netty/src/main/java/org/kaazing/mina/core/service/AbstractIoAcceptorEx.java
@@ -65,7 +65,7 @@ public abstract class AbstractIoAcceptorEx extends AbstractIoAcceptor implements
             public void operationComplete(BindFuture future) {
                 if (future.isBound()) {
                     boolean activate = false;
-                    synchronized (bindLock) {
+                    synchronized (boundAddresses) {
                         if (boundAddresses.isEmpty()) {
                             activate = true;
                         }
--- a/transport/nio/src/main/java/org/kaazing/gateway/transport/nio/internal/AbstractNioAcceptor.java
+++ b/transport/nio/src/main/java/org/kaazing/gateway/transport/nio/internal/AbstractNioAcceptor.java
@@ -377,23 +377,19 @@ public abstract class AbstractNioAcceptor implements BridgeAcceptor {
         // a hang when there's a race with Gateway.destroy().
         if (!unbindScheduler.isShutdown()) {
             final UnbindFuture future = new DefaultUnbindFuture();
-            unbindScheduler.submit(new FutureTask<>(new Runnable() {
-                @Override
-                public void run() {
-                    try {
-                        unbindInternal(address);
-                        future.setUnbound();
-                    } catch (ThreadDeath td) {
-                        throw td;
-                    } catch (Throwable t) {
-                        LOG.error("Exception during unbinding:\n"+address, t);
-                        //System.out.println("Exception during unbinding:\n"+address+"\n");
-                        //t.printStackTrace(System.out);
-                        future.setException(t);
-                        throw new RuntimeException(t);
-                    }
-                }
-            }, Boolean.TRUE));
+            try {
+                unbindInternal(address);
+                future.setUnbound();
+            } catch (ThreadDeath td) {
+                throw td;
+            } catch (Throwable t) {
+                LOG.error("Exception during unbinding:\n"+address, t);
+                //System.out.println("Exception during unbinding:\n"+address+"\n");
+                //t.printStackTrace(System.out);
+                future.setException(t);
+                throw new RuntimeException(t);
+            }
+
             return future;
         }
         else {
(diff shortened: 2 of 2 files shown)


jitsni:
can you attach the thread dump (without your changes - so that line numbers match develop branch) ?

stanculescu:
I attached the thread dump without the changes.
[ThreadDump.txt](https://github.com/kaazing/gateway/files/693276/ThreadDump.txt)

jitsni:
I wonder why the CloseFuture doesn't succeed in the following thread:
```
"org.kaazing.gateway.transport.nio.internal.NioSocketAcceptor@52feb982_unbind-1" #20 prio=5 os_prio=31 tid=0x00007fbe2514e800 nid=0x6b03 in Object.wait() [0x000070000f1b2000]
   java.lang.Thread.State: WAITING (on object monitor)
	at java.lang.Object.wait(Native Method)
	- waiting on <0x000000076c45c4e8> (a org.jboss.netty.channel.AbstractChannel$ChannelCloseFuture)
	at java.lang.Object.wait(Object.java:502)
	at org.jboss.netty.channel.DefaultChannelFuture.awaitUninterruptibly(DefaultChannelFuture.java:237)
	- locked <0x000000076c45c4e8> (a org.jboss.netty.channel.AbstractChannel$ChannelCloseFuture)
	at org.kaazing.mina.netty.ChannelIoAcceptor.unbind0(ChannelIoAcceptor.java:194)
	at org.kaazing.mina.core.service.AbstractIoAcceptor.unbind(AbstractIoAcceptor.java:382)
```

jitsni:
https://issues.apache.org/jira/browse/DIRMINA-807 which deals with different deadlock issue in AbstractIoAcceptor. I think we should use similar kind of strategy of using different lock (boundAddresses) where only addresses are manipulated.

Not sure why service listeners are fired within bindLock (this is a change from mina's version).

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/internal/AbstractNioAcceptor.java
```diff
@@ -377,23 +377,19 @@ public UnbindFuture unbind(final ResourceAddress address) {
         // a hang when there's a race with Gateway.destroy().
         if (!unbindScheduler.isShutdown()) {
```
jitsni: If this is not scheduled onto unbindScheduler, no point in checking for shutdown status

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/internal/AbstractNioAcceptor.java
```diff
@@ -377,23 +377,19 @@ public UnbindFuture unbind(final ResourceAddress address) {
...
+                future.setUnbound();
+            } catch (ThreadDeath td) {
+                throw td;
+            } catch (Throwable t) {
+                LOG.error("Exception during unbinding:\n"+address, t);
+                //System.out.println("Exception during unbinding:\n"+address+"\n");
+                //t.printStackTrace(System.out);
+                future.setException(t);
+                throw new RuntimeException(t);
+            }
```
jitsni: I think that this change doesn't fix the issue. 
1. bindAsync releases bindLock and tries acquire it once the bind succeeded.
2. Between those two states, unbind acquires the bindLock and doesn't progress. 

unbind doesn't progress until bind thread NioServerBoss#process(Selector) which fires the channelUnregisteredFutures. But bind thread is blocked in 1, hence there is a deadlock.

stanculescu:
As you suggested I use a similar strategy and change the lock with boundAddresses in bindAsync method. Now the test pass without the previous changes.

mina.netty/src/main/java/org/kaazing/mina/core/service/AbstractIoAcceptorEx.java
```diff
@@ -65,7 +65,7 @@ public BindFuture bindAsync(final SocketAddress localAddress) {
             public void operationComplete(BindFuture future) {
                 if (future.isBound()) {
                     boolean activate = false;
-                    synchronized (bindLock) {
+                    synchronized (boundAddresses) {
```
jitsni: There are other places where boundAddresses is modified across threads. This is not that simple.

jitsni:
I have a *rough* PR https://github.com/kaazing/gateway/pull/850 to address this issue.  Can you check if this solves the current issue. Need to investigate further if there is better solution.

stanculescu:
I checked and it works. The test passed but there are still tests in gateway that fail during the build.

jitsni:
The tests that are failed may not be related this change. What tests are failing in this branch ?

jitsni:
I updated the PR with mina's version of AbstractIoAcceptor which uses boundAdresses lock when updating the boundAddresses.

The main change is fireServiceActivated/fireServiceDeactivated are called outside bindLock, but the listeners use boundAddresses lock in async version.

jitsni:
Closing ths one as PR https://github.com/kaazing/gateway/pull/850 addresses the same

PR #850, opened 2017-01-11
Syncing AbstractIoAcceptor from mina

jitsni:
1. bind async starts
2. unbind thread (some scheduler thread) acquires bindLock and waits on CloseFuture of channels. But doesn't make any progress as these futures are not fired.
3. bind thread (boss thread) finishes binding but is trying to call listeners. It tries to acquire bindLock but cannot since it is owned by unbind thread. The futures in step 2 are fired by bind thread. Since it doesn't make progress the futures are not fired.

Hence we have the deadlock.  I am not sure whether the listeners need to acquire bindLock at all. The following changes address the issue:

- Syncing AbstractIoAcceptor with latest mina version (which uses boundAddresses lock)
- fireServiceActivated/fireServiceDeactivated are not in synchronized block in mina
- Not sure bindLock itself is required
- bindAsync/unbindAsync doesn't use bindLock in the the listener. Instead it uses
boundAddresses lock(From the beginning bindAsync()/unbindAsync() don't hold the
bindLock but tries acquire in listener. This is different is sync versions)

PR for https://github.com/kaazing/tickets/issues/866

### DIFF
--- a/mina.netty/src/main/java/org/kaazing/mina/core/service/AbstractIoAcceptor.java
+++ b/mina.netty/src/main/java/org/kaazing/mina/core/service/AbstractIoAcceptor.java
@@ -25,9 +25,6 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
-import java.util.concurrent.Semaphore;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.mina.core.RuntimeIoException;
 import org.apache.mina.core.service.IoAcceptor;
@@ -39,6 +36,7 @@ import org.apache.mina.core.session.IoSessionConfig;
 /**
  * A base implementation of {@link IoAcceptor}.
  *
+ * @author <a href="http://mina.apache.org">Apache MINA Project</a>
  * @org.apache.xbean.XBean
  */
 /* This class (based on the Mina version) is needed for use in ChannelIoAcceptor in order to use our
@@ -48,9 +46,7 @@ import org.apache.mina.core.session.IoSessionConfig;
  * 2. Add imports of needed classes from the original package (org.apache.mina.core.service)
  * 3. Change checkAddressType, bindLock and boundAddresses from private to package-private so they can be used in
  *    AbstractIoAcceptorEx
- * 4. Fix apparent Mina bugs in bind and unbind: fire service listeners to indicate service activated or deactivated
- *    inside the synchronized bindLock block, to avoid possible wrong outcome in case of race between bind and unbind.
- * 5. Widen scope of bind lock for subclasses.
+ * 4. Widen scope of bind lock for subclasses.
  */
 public abstract class AbstractIoAcceptor
         extends AbstractIoService implements IoAcceptor {
@@ -69,7 +65,7 @@ public abstract class AbstractIoAcceptor
      * Acquire this lock in your property setters which shouldn't be changed while
      * the service is bound.
      */
-    protected final Semaphore bindLock = new Semaphore(1);
+    protected final Object bindLock = new Object();
 
     /**
      * Constructor for {@link AbstractIoAcceptor}. You need to provide a default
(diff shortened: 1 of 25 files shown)


mina.netty/src/main/java/org/kaazing/mina/core/service/AbstractIoAcceptorEx.java
```diff
@@ -59,20 +59,21 @@ public BindFuture bindAsync(final SocketAddress localAddress) {
         if (getHandler() == null) {
             throw new IllegalStateException("handler is not set.");
         }
+
         BindFuture bound = bindAsyncInternal(localAddress);
```
jitsni: Surprisingly, the above was not wrapped in synchronized(bindLock) from the beginning.

mina.netty/src/main/java/org/kaazing/mina/core/service/AbstractIoAcceptor.java
```diff
@@ -106,7 +105,8 @@ public SocketAddress getLocalAddress() {
     @Override
     public final Set<SocketAddress> getLocalAddresses() {
         Set<SocketAddress> localAddresses = new HashSet<>();
-        synchronized (bindLock) {
```
cmebarrow: I don't think bindLock is required since we always synchronize the same code block on boundAddresses as well. Is it still present in the mina version?

mina.netty/src/main/java/org/kaazing/mina/core/service/AbstractIoAcceptorEx.java
```diff
@@ -59,20 +59,21 @@ public BindFuture bindAsync(final SocketAddress localAddress) {
...
+                    synchronized (boundAddresses) {
                         if (boundAddresses.isEmpty()) {
                             activate = true;
                         }
                         boundAddresses.add(localAddress);
-                        if (activate) {
-                            getListeners().fireServiceActivated();
-                        }
+                    }
+                    if (activate) {
```
cmebarrow: The problem about moving the firing of the listeners outside the synchronized block is the outcome (service activated or not activated) may now be wrong if there's a race between bind and unbind done by different threads: if this the first bind and bind sync block completes before unbind sync block but unbind thread first serviceDisactivated before bind thread fires serviceActivated then outcome will be service activated, which is wrong.

mina.netty/src/main/java/org/kaazing/mina/core/service/AbstractIoAcceptor.java
```diff
@@ -45,9 +46,7 @@
  * 2. Add imports of needed classes from the original package (org.apache.mina.core.service)
```
jitsni: Updating the mina version to 2.0.16 in the comment

mina.netty/src/main/java/org/kaazing/mina/core/service/AbstractIoAcceptor.java
jitsni: This file is pretty much same as in mina version. It uses both bondLock and boundAddresses locks

mina.netty/src/main/java/org/kaazing/mina/core/service/AbstractIoAcceptor.java
```diff
@@ -266,11 +267,11 @@ public final void bind(SocketAddress firstLocalAddress, SocketAddress... otherLo
     @Override
     public final void bind(Iterable<? extends SocketAddress> localAddresses) throws IOException {
         if (isDisposing()) {
-            throw new IllegalStateException("Already disposed.");
+            throw new IllegalStateException("The Accpetor disposed is being disposed.");
```
jitsni: It is there in the mina version. So I just kept it

mina.netty/src/main/java/org/kaazing/mina/core/service/AbstractIoAcceptorEx.java
jitsni: I was thinking the same. This must be happening even in the current code. for e.g
1. bind gets scheduled on boss thread
2. unbind completes, acquires bindLock and fires serviceDeactivated, releases bindLock
3. bind completes, acquires bindLock and fires serviceActivated, releases bindLock

1. One solution would be to use semaphore (see my first commit) which holds the bindLock throughout (from bind scheduled to firing listeners) across threads. But semaphore is not reentrant which cause problems (for e.g a listener might call one more bind before the first bind completely finished)
2. Other solution would be to schedule unbind onto the same boss thread. Then we can just remove the syncing on bindLock in async versions.

cmebarrow:
@jitsni Looks good to me but the build failed in NIO transport with 
```
Tests in error: 
  UdpConnectorIT.clientSentData:142 NullPointer
```
Could you take a look>

jitsni:
UdpConnectorIT.clientSentData failure is unrelated. Now travis build passes. I create a separate PR for that.

PR #851, opened 2017-01-13
Added Log warning first time we see Http 1.0.

vmaraloiu:


service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -206,6 +206,12 @@ public void sessionOpened(IoSession session) {
             final DefaultHttpSession acceptSession = (DefaultHttpSession) session;
             // final Subject subject = ((IoSessionEx) acceptSession).getSubject();
 
+            // log warning first time we see Http 1.0 request
+            if (acceptSession.getVersion().toString().equals("HTTP/1.0")){
+                LOGGER.warn(String.format("http.proxy service %s received an HTTP 1.0 request. HTTP 1.0 is not"
```
dpwspoon: @vmaraloiu, this would log it every time you see it correct?  it would be better to log it just the first time

service/http.proxy/src/main/java/org/kaazing/gateway/service/http/proxy/HttpProxyServiceHandler.java
```diff
@@ -206,6 +206,12 @@ public void sessionOpened(IoSession session) {
...
+                         + " explicitly supported.", acceptSession.getService().toString()));
```
dpwspoon: I'm not convinced the abstraction getService().toString() actually is supported in the gateway service code to provide useful information.  What is actually returned in the test?  you could add a field in the init() with the `serviceName = serviceContext.getServiceName()`

service/http.proxy/src/test/java/org/kaazing/gateway/service/http/proxy/HttpProxyNegativeCasesIT.java
```diff
@@ -50,12 +64,16 @@
...
-    public TestRule chain = createRuleChain(gateway, robot);
+    public TestRule chain =
+            RuleChain.outerRule(new MethodExecutionTrace()).around(robot).around(checkLogMessageRule).around(gateway);
 
     @Specification("http.proxy.http.1.0.request")
     @Test
     public void sendHttp_1_0_Request() throws Exception {
         robot.finish();
+        expectedPatterns = new ArrayList<String>(Arrays.asList(new String[] {
+                "http.proxy service .*. received an HTTP 1.0 request. HTTP 1.0 is not explicitly supported."
```
dpwspoon: The change in the comment above would require both explicitly stating what is currently ".*" in this regex.  And adding a .name("name") to the service element

PR #855, opened 2017-01-18
.ksn parameter is converted into X-Sequence-No header for

jitsni:
.ksn parameter is converted into X-Sequence-No header for
wse/1.0 protocol (already does for httpxe/1.0)

### DIFF
--- a/transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpProtocolCompatibilityFilter.java
+++ b/transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpProtocolCompatibilityFilter.java
@@ -301,14 +301,6 @@ public class HttpProtocolCompatibilityFilter extends HttpFilterAdapter<IoSession
                 }
             }
 
-            if (PROTOCOL_HTTPXE_1_1.equals(httpRequest.getHeader(HEADER_X_NEXT_PROTOCOL)) && !httpRequest.hasHeader(HttpHeaders.HEADER_X_SEQUENCE_NO)) {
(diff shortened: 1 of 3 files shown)


apirvu:
JavaScript itests on IE9 pass with this PR, please see comment on [ticket 950](https://github.com/kaazing/tickets/issues/950#issuecomment-273467549).

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpProtocolCompatibilityFilter.java
```diff
@@ -425,6 +417,19 @@ else if (!PROTOCOL_HTTPXE_1_1.equals(nextProtocol)) {
             }
         }
 
+        // Use .ksn for x-sequence-no header for wse/1.0, httpxe/1.1 cases
+        // other cases, don't do anything as some services (for e.g http.proxy) want to pass through
+        String protocol = httpRequest.getHeader(HEADER_X_NEXT_PROTOCOL);
```
dpwspoon: Do you want to get this from the sessions resource address `address.getOption(NEXT_PROTOCOL)`?  I would imagine the http proxy would still treat requests that have that header value with the special rewrite behavior, correct?

jitsni: No, the code above is setting the HEADER_X_NEXT_PROTOCOL to PROTOCOL_HTTPXE_1_1 and to PROTOCOL_WSE_1_0. So, that may not work.

PR #860, opened 2017-01-26
Remove System Properties which should be undocumented from Javadoc

vmaraloiu:


vmaraloiu:
There seems to be some constants unused in [org.kaazing.gateway.util.InternalSystemProperty](https://github.com/kaazing/gateway/blob/develop/util/src/main/java/org/kaazing/gateway/util/InternalSystemProperty.java) as well, like this one [TCP_READ_BUFFER_SIZE](https://github.com/kaazing/gateway/blob/develop/util/src/main/java/org/kaazing/gateway/util/InternalSystemProperty.java#L59). This should be investigated?

dpwspoon:
This seems okay to me.  

> There seems to be some constants unused in org.kaazing.gateway.util.InternalSystemProperty as well, like this one TCP_READ_BUFFER_SIZE. This should be investigated?

Happy to have them cleaned up, either in the PR or another

vmaraloiu:
I have created this [PR](https://github.com/kaazing/gateway/pull/861) for [the clean up](https://github.com/kaazing/gateway/pull/860#issuecomment-275482797).

PR #861, opened 2017-01-27
Remove unused System Properties from util.InternalSystemProperty

vmaraloiu:


dpwspoon:
These all appear to be used in other places (For example: BROADCAST_SERVICE_DISCONNECT_CLIENTS_ON_RECONNECT is [here](https://github.com/kaazing/gateway/search?utf8=%E2%9C%93&q=BROADCAST_SERVICE_DISCONNECT_CLIENTS_ON_RECONNECT&type=Code)).  We should either have them all in the InternalSystemProperty class or all spread out.  I would vote to have them all reference the InternalSystemProperty class. Thus rather than removing them here, we should have all the other places reference them here.

dpwspoon:
Closing based on comment above for now, happy to continue discussing if you disagree

PR #862, opened 2017-01-30
Add.hazelcast.topics

Anisotrop:
First implementation of topics interface defined in Hazelcast. Using 2 executor services: 
- add/remove listeners, which are blocking operations
- publish events, an operation specified as fire and forget


service/spi/src/test/java/org/kaazing/gateway/service/messaging/collections/MemoryTopicTest.java
```diff
@@ -0,0 +1,175 @@
...
+            listenersCalled.countDown();
+        };
+        topic.addMessageListener(m2);
+        topic.publish("msg1");
+        listenersCalled.await();
+        assertTrue("MessageListener m3 must not be called", calledM3.get() == false);
+    }
+
+    @Test
+    public void shouldPubSubFromMessageListeners() {
```
cmebarrow: We won't actually need to do this so we don't need to support it. Would be interesting to know if Hazelast supports it. We'll find out when you do the Hazelcast variant of the test.

service/spi/src/test/java/org/kaazing/gateway/service/messaging/collections/MemoryTopicTest.java
```diff
@@ -0,0 +1,175 @@
...
+        };
+        String name = topic.addMessageListener(m1);
+        MessageListener m2 = message -> {
+            topic.removeMessageListener(name);
+            listenersCalled.countDown();
+        };
+        topic.addMessageListener(m2);
+        topic.publish("msg1");
+        listenersCalled.await();
+        assertTrue("MessageListener m3 must not be called", calledM3.get() == false);
```
cmebarrow: Can we make assumptions about what order the listeners are called in? I would think not.

service/spi/src/test/java/org/kaazing/gateway/service/messaging/collections/MemoryTopicTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+
+    @Test
+    public void shouldNotOverflow() throws InterruptedException {
+        ITopic<String> topic = factory.getTopic("topic");
+        CountDownLatch latch = new CountDownLatch(10000);
+        topic.addMessageListener(message -> {
+            latch.countDown();
+            if (latch.getCount() > 0) {
+                topic.publish("Resend: " + message.getMessageObject());
```
cmebarrow: Again I'm not sure we really need to support this. If we can, it does no harm. When I would be equally happy if we just throw an exception from the publish they attempt to this.

service/spi/src/main/java/org/kaazing/gateway/service/collections/MemoryTopic.java
```diff
@@ -0,0 +1,133 @@
...
+import com.hazelcast.core.Message;
+import com.hazelcast.core.MessageListener;
+import com.hazelcast.monitor.LocalTopicStats;
+import com.hazelcast.monitor.impl.LocalTopicStatsImpl;
+
+public class MemoryTopic<E> extends MemoryDistributedObject implements ITopic<E> {
+
+    private static final ExecutorService AR_EXECUTOR =
+        Executors.newSingleThreadExecutor(runnable -> new Thread(runnable, "memory-topics-add/remove"));
+    private static final ExecutorService P_EXECUTOR =
```
cmebarrow: Using two threads, and especially using ExecutorService and submitting tasks will create garbage, and is using an unbounded queue. I would prefer if you can move to an architecture closer to that of Nuklei and the MQ driver, where each work request (add listener, remove listener and publish) is submitted to a fixed length Agrona queue (instance of ManyToOneConcurrentArrayQueue), and executed by a single worker thread. The registration id returned by addMessageListener would be included in the enqueued setMessageListener command so the method does not need to get an answer from the worker thread. You could have a single worker thread to service all the ITopic instances if you want, though I think it's fine to have a separate thread per topic if it's easier, since we will not be needing to use a large number of topics.

Anisotrop: I will look into it. The reason for the 2 threads is to avoid deadlocks when add/remove is called from a message listener. These methods are specified as blocking while publish is specified as fire-and-forget. It we make add/remove unblocking it cannot guarantee a happens-before relationship between adding a listener and publishing on a topic.

service/spi/src/main/java/org/kaazing/gateway/service/collections/MemoryDistributedObject.java
```diff
@@ -0,0 +1,48 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kaazing.gateway.service.collections;
+
+import com.hazelcast.core.DistributedObject;
+
+
+public class MemoryDistributedObject implements DistributedObject {
```
Anisotrop: The code of the topic implementation was less polluted. But I can move these methods in MemoryTopic.

service/spi/src/test/java/org/kaazing/gateway/service/messaging/collections/MemoryTopicTest.java
```diff
@@ -0,0 +1,175 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.RuleChain;
+import org.kaazing.gateway.service.collections.MemoryCollectionsFactory;
+import org.kaazing.test.util.ITUtil;
+
+import com.hazelcast.core.ITopic;
+import com.hazelcast.core.MessageListener;
+
+public class MemoryTopicTest {
```
Anisotrop: Will be working on this test.

service/spi/src/main/java/org/kaazing/gateway/service/collections/MemoryTopic.java
cmebarrow: If publishing is done same way (by posting a command to the same queue) then ordering is guaranteed.

service/spi/src/test/java/org/kaazing/gateway/service/messaging/collections/MemoryTopicTest.java
```diff
@@ -16,22 +16,28 @@
...
 import org.kaazing.test.util.ITUtil;
 
 import com.hazelcast.core.ITopic;
 import com.hazelcast.core.MessageListener;
 
+
+// TODO move this test to org.kaazing.gateway.server.context.resolve (standalone/default)
+//
+// Tests not covered:
+//    - topic publish will happen-before calling add/remove inside a message listener
```
cmebarrow: I am not sure it's worth supporting calling publish, add and remove listener inside a message listener. We will not need this. Does Hazelcast even support it? If you want you can just throw UnsupportedOperationException and log an error.

cmebarrow:
@Anisotrop I am uneasy about the use of concurrent hashmap and its entryset method in MemoryTopic. It will create garbage and slow up the system. publishing is going to be a frequent event so hot code. I think there are two ways we can go here:
1. Call the listeners directly from the publish implementation, i.e. do not use a separate thread at all. Use a concurrent data structure to hold the list of listeners (probably CopyOnWriteArrayList). I think this might be the best way to go because it's easy, garbage free, and when we use the feature in the MQ driver the listeners will be extremely lightweight (they were just post a command onto an Agrona queue which is processed by the queue reader thread).  Obviously you  would have to forbid calls to publish, add or remove listener from within a llistener, but as I said already in another comment I think that would be just fine.
2. Post add listener, remove listener and publish as commands onto an Agrona queue which is processed by a worker thread (could have just one thread per MemoryCollectionsFactory instance, it can handle all topics). With this approach everything is single threaded and ordering is guaranteed correct. The thread lifecycle would be managed by the MemoryCollectionsFactory instance (so not using SchedulerProvider).

We can talk tomorrow if you want, after the dev meeting.

service/spi/src/test/java/org/kaazing/gateway/service/messaging/collections/MemoryCollectionsFactoryTest.java
```diff
@@ -32,10 +32,12 @@
 import org.junit.Test;
 import org.kaazing.gateway.service.collections.MemoryCollectionsFactory;
 import org.kaazing.gateway.util.AtomicCounter;
+import org.kaazing.gateway.util.scheduler.SchedulerProvider;
```
Anisotrop: Will be removing this import.

service/spi/src/main/java/org/kaazing/gateway/service/collections/MemoryCollectionsException.java
```diff
@@ -0,0 +1,38 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kaazing.gateway.service.collections;
+
+
+public class MemoryCollectionsException extends RuntimeException {
```
cmebarrow: This class is not needed. In MemoryTopic.addMessageListener(MessageListener<E>) you can just throw a NullPointerException when the message listener is null, because that is the behavior stipulated in the javadoc of the ITopic interface.

server/src/test/java/org/kaazing/gateway/server/context/resolve/DefaultClusterContextTest.java
```diff
@@ -234,6 +245,143 @@ public void shouldNotListenOnMulticastInterfaceImplicitly() throws Exception {
...
+            topicMessageListenerMember1.addMessageListener(message -> {listenersCalledNullPointer.countDown(); throw new NullPointerException();});
+            topicMessageListenerMember1.addMessageListener(message -> listenersCalledNullPointer.countDown());
+            ITopic<String> topicMessageListenerMember2 = clusterContext2.getTopic("topic_message_listener_null_pointer");
+            topicMessageListenerMember2.publish("msg1");
+            listenersCalledNullPointer.await();
+            assertEquals(1, topicMessageListenerMember2.getLocalTopicStats().getPublishOperationCount());
+            assertEquals(3, topicMessageListenerMember1.getLocalTopicStats().getReceiveOperationCount());
+
+
+            ITopic<String> topicSameListenerMember1 = clusterContext1.getTopic("topic_same_listener");
```
cmebarrow: This case does not matter to us, we will never do this. Generally I would expect that the tests in this method would be covered by separate test methods in DefaultClusterTopicTest and StandaloneClusterTopicTest, so maybe they are not needed here?

server/src/test/java/org/kaazing/gateway/server/topic/DefaultClusterTopicTest.java
```diff
@@ -0,0 +1,205 @@
...
+import org.junit.Test;
+import org.junit.rules.RuleChain;
+import org.kaazing.gateway.server.context.resolve.DefaultClusterContext;
+import org.kaazing.gateway.service.cluster.MemberId;
+import org.kaazing.test.util.ITUtil;
+
+import com.hazelcast.core.ITopic;
+import com.hazelcast.core.MessageListener;
+
+public class DefaultClusterTopicTest {
```
cmebarrow: I would expect that this test and StandaloneClusterTopicTest would be sharing the same code, probably by both extending the same abstract class,  because the behavior for topics should be the same for the two cases. The superclass can have a test rule or before and after methods to create and dispose the cluster contexts by calling abstract methods which would be implemented differently in the subclasses.

server/src/test/java/org/kaazing/gateway/server/topic/DefaultClusterTopicTest.java
```diff
@@ -0,0 +1,205 @@
...
+        t.join();
+        topicMember1.publish("msg1");
+        listenersCalled.await();
+        assertEquals(1, topicMember1.getLocalTopicStats().getPublishOperationCount());
+        assertEquals(2, topicMember1.getLocalTopicStats().getReceiveOperationCount());
+        topicMember1.destroy();
+    }
+
+    @Test
+    public void shouldSendAndReceiveMessagesBetweenMembers() throws InterruptedException {
```
cmebarrow: Obviously this method does not apply to the standalone case so would not appear in the abstract superclass.

server/src/test/java/org/kaazing/gateway/server/context/resolve/DefaultClusterContextTest.java
```diff
@@ -33,15 +34,22 @@
...
 import org.kaazing.gateway.service.cluster.ClusterContext;
 import org.kaazing.gateway.service.cluster.MemberId;
 import org.kaazing.gateway.service.cluster.MembershipEventListener;
 import org.kaazing.gateway.service.collections.CollectionsFactory;
+import org.kaazing.test.util.ITUtil;
+
+import com.hazelcast.core.ITopic;
+import com.hazelcast.core.MessageListener;
 
 public class DefaultClusterContextTest {
```
cmebarrow: Please fix the compilation warnings in this test and in your two new tests.

server/src/test/java/org/kaazing/gateway/server/topic/DefaultClusterTopicTest.java
```diff
@@ -34,22 +34,22 @@
 import com.hazelcast.core.ITopic;
 import com.hazelcast.core.MessageListener;
 
-public class DefaultClusterTopicTest {
+// TODO Add a parent abstract class that defines test cases for both cluster and single node.
```
cmebarrow: Isn't this already done? If so remove this comment

server/src/test/java/org/kaazing/gateway/server/topic/StandaloneClusterTopicTest.java
```diff
@@ -29,13 +29,13 @@
...
 import org.kaazing.gateway.server.context.resolve.StandaloneClusterContext;
 import org.kaazing.gateway.service.collections.CollectionsFactory;
-import org.kaazing.gateway.service.collections.MemoryCollectionsException;
 import org.kaazing.test.util.ITUtil;
 
 import com.hazelcast.core.ITopic;
 import com.hazelcast.core.MessageListener;
 
-public class StandaloneClusterTopicTest {
+// TODO Add a parent abstract class that defines test cases for both cluster and single node.
```
cmebarrow: Again I think this comment line should now be removed

server/src/test/java/org/kaazing/gateway/server/topic/StandaloneClusterTopicTest.java
```diff
@@ -143,53 +72,6 @@ public void shouldRejectAddAndRemoveFromMessageListenerSameThread() throws Inter
...
-        assertEquals(1, topic1.getLocalTopicStats().getReceiveOperationCount());
-        assertEquals(1, topic2.getLocalTopicStats().getPublishOperationCount());
-        assertEquals(1, topic2.getLocalTopicStats().getReceiveOperationCount());
-        topic1.destroy();
-        topic2.destroy();
-    }
-
-
-    @Test
     public void shouldNotAllowNestedPublish() throws InterruptedException {
```
cmebarrow: You should catch the expected UnsupportedOperationException from the call to publish within the listener and set a variable to indicate the expected exception was thrown.  That would make the test more robust. As it is it could accidentally pass due to timing even if the exception was not thrown.

service/spi/src/main/java/org/kaazing/gateway/service/collections/MemoryTopic.java
```diff
@@ -64,14 +69,31 @@ protected Boolean initialValue() {
...
+    }
+
+    @Override
+    public String getServiceName() {
+        return null;
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
     public void publish(Object o) {
```
cmebarrow: This should take the parameterized type: 
```
   public void publish(E o) {
```
and you should use a parameterized message:
```
Message<E> m = new Message<>(this.getName(), o, System.nanoTime(), null);
```
That may improve the handling of the wrong type case, but at least it will get rid of a compilation warning  and is more correct

server/src/test/java/org/kaazing/gateway/server/topic/AbstractClusterTopicTest.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void shouldNotAddNullMessageListener() {
+        ITopic<String> topic = getTopicMember1ForShouldNotAddNullMessageListener();
+        topic.addMessageListener(null);
+    }
+
+    @Test
+    // TODO take into account the following as limitation or study how the type can be enforced
```
cmebarrow: Very clever of you to think of this test case! What did you have in mind  for the TODO item? Do you think we really can or should try to do something about it? I think no, since your DefaultClusterTopicTest shows Hazelcast as the same behavior. So I would recommend just removing this TODO comment.

PR #866, opened 2017-02-03
ExceptionMonitor can receive a null session.

Anisotrop:
Fallback to org.apache.mina.util.ExceptionMonitor when session is null.

### DIFF
--- a/mina.netty/src/main/java/org/kaazing/mina/core/service/AbstractIoService.java
+++ b/mina.netty/src/main/java/org/kaazing/mina/core/service/AbstractIoService.java
@@ -306,7 +306,11 @@ public abstract class AbstractIoService implements IoService {
                 try {
                     this.disposalFuture = disposalFuture = dispose0();
                 } catch (Exception e) {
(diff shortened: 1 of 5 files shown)


mina.netty/src/main/java/org/kaazing/mina/core/service/AbstractIoService.java
```diff
@@ -306,7 +306,11 @@ public final void dispose() {
                 try {
                     this.disposalFuture = disposalFuture = dispose0();
```
dpwspoon: Can we identify which connector is returning null for dispose0, and fix that instead of doing the NPE guard?  We should then be able to test it

Anisotrop: dispose0 returns null in every case as per @stanculescu's investigation. It might have been easier in this case to call the mina ExceptionMonitor directly and not pass through the mina.netty adapter which does that on null session.

Anisotrop:
@dpwspoon Can you please review again ?

dpwspoon:
Yup! Thanks. LGTM

PR #868, opened 2017-02-13
NPEs in DEBUG and ERROR log if IP address is incorrect 

StCostea:
Fix for https://github.com/kaazing/tickets/issues/930


dpwspoon:
@StCostea see travis failure.

PR #870, opened 2017-02-21
 Remove WSN_302_REDIRECT and WSX_302_REDIRECT features

mgherghe:


service/http.balancer/src/test/java/org/kaazing/gateway/service/http/balancer/WsxBalancerIT.java
```diff
@@ -54,7 +53,6 @@
                             .allowOrigin("*")
```
dpwspoon: We need at least 4 tests.  WSX with kl=y (sends balancer bytes), WSX with out kl=y ( sends wrapped 302),  WS with kl=y (sends balancer bytes), WS with out kl=y ( sends 302).  I believe we only have 2 cases covered in the original tests

transport/wsn/src/main/java/org/kaazing/gateway/transport/wsn/WsnAcceptor.java
```diff
@@ -1114,22 +1112,16 @@ private void doUpgrade(final HttpAcceptSession session) throws Exception {
...
-                            List<String> availableBalanceeURIs = (List<String>) balancerKeys;
-                            String balanceURI = availableBalanceeURIs.get((int) (Math.random() * availableBalanceeURIs.size()));
-                            balanceURI = URLUtils.modifyURIScheme(URI.create(balanceURI), "http").toString();
-                            session.setStatus(HttpStatus.REDIRECT_FOUND);
-                            session.setWriteHeader("location", balanceURI);
-                            session.close(false);
-                            break;
-                        }
+                    if (!"x-kaazing-handshake".equals(wsProtocol0) && wsVersion == HYBI_13 && balancerKeys != null &&
+                            !"Y".equals(session.getParameter(".kl"))) {
```
dpwspoon: This looks good to me, thoughI'm not sure if the .kl is going to be presented properly in the wsx cases (I think the layering might remove it :-().  If we add the tests above, and they prove me wrong, then this works.  Otherwise, we will need some more logic

service/http.balancer/src/test/java/org/kaazing/gateway/service/http/balancer/WsxBalancerIT.java
mgherghe: I believe 3 of those 4 tests already exist. 
- `WSX with kl=y (sends balancer bytes)` - [legacy.wsx.balancer.request](https://github.com/kaazing/gateway/blob/develop/service/http.balancer/src/test/scripts/org/kaazing/gateway/service/http/balancer/legacy.wsx.balancer.request.rpt)
- `WSX without kl=y ( sends wrapped 302)` - [wsx.balancer.request](https://github.com/kaazing/gateway/blob/develop/service/http.balancer/src/test/scripts/org/kaazing/gateway/service/http/balancer/wsx.balancer.request.rpt) which is modified in this PR, [here](https://github.com/kaazing/gateway/pull/870/files#diff-a56af8bf0eb96b06a70a193c8596adf6)
- `WS without kl=y (sends 302)` - [wsn.balancer.request](https://github.com/kaazing/gateway/blob/develop/service/http.balancer/src/test/scripts/org/kaazing/gateway/service/http/balancer/wsn.balancer.request.rpt)

So, the only test case missing is `WS with kl=y (sends balancer bytes)`

dpwspoon: +1 awesome, looking forward to the missing test.  Thanks

mgherghe:
@dpwspoon I updated the PR and added the missing test for the case: `WS with kl=y (sends balancer bytes)`

PR #871, opened 2017-02-21
update gateway login challenge behavior

danibusu:
- update login modules to throw exception instead of returning false
- add integration tests for login behavior
- update xsd and code so challenge scheme is not mandatory
- update unit tests


security/pom.xml
```diff
@@ -49,6 +49,10 @@
             <artifactId>jmock-legacy</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.kaazing</groupId>
```
dpwspoon: `<scope>test</scope>` ? based on usage I see in this diff

management/src/test/java/org/kaazing/gateway/management/jmx/JmxRoundTripLatencyIT.java
```diff
@@ -93,7 +93,6 @@
                             .realm()
                                 .name("jmxrealm")
                                 .description("realm for jmx")
```
dpwspoon: I believe these were removed because the tests don't execute any actual challenging/token.  Please confirm.  (Also I'm unclear why login modules are needed at all for this test, maybe round trip latency requires them?)

dpwspoon: related script: https://github.com/kaazing/gateway/blob/develop/management/src/test/scripts/org/kaazing/gateway/management/jmx/echoServiceToGetRoundTripLatencyAttributesFromJMX.rpt

security/src/main/java/org/kaazing/gateway/security/auth/BaseStateDrivenLoginModule.java
```diff
@@ -141,5 +153,18 @@ public boolean logout() throws LoginException {
 
     protected abstract boolean doLogout() throws LoginException;
 
+    private LoginResult getLoginResultFromCallback() {
+        LoginResultCallback loginResultCallback = new LoginResultCallback();
+
+        try {
+            this.handler.handle(new Callback[]{loginResultCallback});
+        } catch (IOException|UnsupportedCallbackException e) {
+            LOGGER.debug("Encountered exception handling loginResultCallback", e);
```
dpwspoon: loginResultCallback should be capitalized "LoginResultCallback", so user knows you are referring to the class. http://grepcode.com/file/repo1.maven.org/maven2/org.kaazing/gateway.server.spi/5.0.1.23/org/kaazing/gateway/server/spi/security/LoginResultCallback.java

security/src/main/java/org/kaazing/gateway/security/auth/BasicLoginModule.java
```diff
@@ -61,16 +61,25 @@ public void initialize(Subject subject, CallbackHandler callbackHandler,
     protected boolean doLogin() throws LoginException {
 
         if (!authenticationSchemeIsBasic()) {
-            return true;
+            throw new LoginException("Only able to handle Basic authentication scheme");
```
dpwspoon: It's unclear to me on whether we want this behavior or not.  

I believe, that this would imply that for multiple realms this would require the same challenge scheme?  What would be the implications of ignoring (return false) if the scheme is not set?  Maybe further login modules fail

security/src/main/java/org/kaazing/gateway/security/auth/BaseStateDrivenLoginModule.java
dpwspoon: This is a nit, I would suggest the same capitalization for other exceptions in this PR.

security/src/main/java/org/kaazing/gateway/security/auth/BasicLoginModule.java
dpwspoon: This question to a bunch of other login modules below.  I will only comment here, but if we decide a change is needed we will need to change several other spots

security/src/main/java/org/kaazing/gateway/security/auth/FileLoginModule.java
```diff
@@ -310,11 +301,14 @@ private void getUsernamePassword(boolean useSharedState) throws LoginException {
 
         } catch (IOException e) {
 //            throw (LoginException) (new LoginException(e.getMessage()).initCause(e));
```
dpwspoon: please remove pointless comments and code when you see them and they are where you are working.

security/src/main/java/org/kaazing/gateway/security/auth/FileLoginModule.java
```diff
@@ -234,19 +226,19 @@ private void cleanState() {
...
+    private void attemptAuthenticate(boolean useSharedState)
         throws LoginException {
 
         getUsernamePassword(useSharedState);
 
         if (username == null) {
-            LOG.debug("[FileLoginModule] " + "username not found, returning false");
-            return false;
+            LOG.debug("[FileLoginModule] username not found");
+            throw new LoginException("Username not found");
```
dpwspoon: It seems to me that in the Http Security Filter, we should log (at debug) the details of the Login Exception.  If that is the case, I don't know if the Log's here are necessary... (thoughts?, if you agree there are probably several spots we could change this).    But we would ideally also include the exception information (if any) from the call getUsernamePassword in this LoginException (I would think).

security/src/test/java/org/kaazing/gateway/security/auth/BasicLoginModuleTest.java
```diff
@@ -135,17 +142,111 @@ public void testBasicLoginModuleIgnoresUnknownSchemes() throws Exception {
...
+        Subject subject = new Subject();
+        DispatchCallbackHandler handler = new DispatchCallbackHandler();
+        DefaultLoginResult loginResult = new DefaultLoginResult();
+        handler.register(AuthenticationTokenCallback.class, new AuthenticationTokenCallbackHandler(s));
+        LoginContext loginContext = new ResultAwareLoginContext("demo", subject, handler, configuration, loginResult);
+
+        context.assertIsSatisfied();
+        assertNotNull(loginContext);
+
+        thrown.expect(LoginException.class);
```
dpwspoon: There are several cases where we are doing exception handling on the callbacks, where the Gateway should always provide them.  I believe UnsupportedCallbackException would be thrown if the gateway failed to provide them and it was an error.  It seems to me, that the implementation of catching this exception and then throwing a LoginResultException with no reference to the missing callback handler is hiding the system/gateway error that it didn't provide that callback in the first place.

What happens if the original exception is allowed to propagate?(maybe internally on the JDK it get wrapped in a LoginResultException.  Would that be a better thing to have happen thereby showing it is a system error and not something implemented wrong in the login module?  Or perhaps there is some precedent for doing what we are doing, or a different solution somewhere in the JDK (maybe you can see what JDK classes extend Callback and where they are used).

This, combined with logging at the debug level in the HttpSecurityFilter the login exception (and other filters where we call this) I think may greatly decrease the amount of exception handling code we had to write.

transport/http/src/test/java/org/kaazing/gateway/transport/http/security/auth/HttpAcceptorAuthenticationIT.java
```diff
@@ -0,0 +1,305 @@
...
+            setThreadingPolicy(new Synchroniser());
+        }
+    };
+
+    @Rule
+    public TestRule chain = RuleChain.outerRule(trace).around(acceptor).around(k3po).around(timeoutRule);
+
+    @Test
+    @Specification("client.no.credentials.unauthorized.empty.additional.challenges")
+    public void testGatewayReturns401WithEmptyAdditionalChallengeDataWhenNoAuthenticationCredentialsProvided() throws Exception {
```
dpwspoon: Excellent test, but the title of this test, specifically the "WhenNoAuthenticationCredentialsProvided" makes it seem like you are testing both the HttpSecurityFilter behavior and the LoginModule behavior.  When in fact you really just want to test the HttpSecurityFilter behavior here for all events it might see from the login modules.  I would suggest renaming the methods to something of the form, return401WhenLoginModuleThrowsLoginExceptionAndSetChallenge (or related per test).

transport/http/src/test/scripts/org/kaazing/gateway/transport/http/security/auth/client.forbidden.rpt
```diff
@@ -0,0 +1,28 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+property connectAddress "http://localhost:8000/resource"
+property someCredentials ${http:loginBase64Encoder("user:password")}
+property authHeader ${http:append("Basic ", someCredentials)}
+
+connect ${connectAddress}
```
dpwspoon: It would be better if all of these scripts came from the HTTP k3po spec or another spec.

security/src/main/java/org/kaazing/gateway/security/auth/BasicLoginModule.java
dpwspoon: I see, after reading down below, I believe the multiple realm case, the first realm/chain for login modules is bypassed on the way back in (correct?), so then it won't get invoked.  So yes I believe throwing the LoginModule Exception here is the correct behavior

dpwspoon:
@danibusu Thanks, looks very good to me with minor feedback.  @NicoletaOita, does this cover all the tests cases you were thinking of (CC @sbadugu)?

security/pom.xml
danibusu: It is present in the parent gateway pom.xml in the dependencyManagement section with scope test.  I can add test scope for readability.

management/src/test/java/org/kaazing/gateway/management/jmx/JmxRoundTripLatencyIT.java
danibusu: Yes, I removed them because the there's no token exchange. The JmxManagementService is built to require credentials when a client is connecting to the MBeanServer. The realm is actually needed to authenticate the clients of the management service.

security/src/main/java/org/kaazing/gateway/security/auth/BasicLoginModule.java
danibusu: Yes, you are correct. The authentication process ignores the realms in the chain that already passed authentication. So, for a realm/chain of login modules, we can only have one challenge scheme. Returning false would also be correct.

security/src/test/java/org/kaazing/gateway/security/auth/BasicLoginModuleTest.java
danibusu: From what I saw, the JDK will wrap most exception types into a LoginException. The handling of IoException and UnsupportedCallbackException only to throw a LoginException is consistent with what happens in the LoginModule implementations provided with the JDK and is driven by the fact that both are checked exceptions.
As a note, if there are multiple login modules in the chain and more than one fails, the LoginContext will only throw one exception, depending on the type of the login modules configured, a required login module exception taking precedence over an optional one. So, it might be helpful to log exceptions in the login module and not rely only on the single one that is thrown by the LoginContext.login() method.
You're right that some of the login module code exists that just swallows the two exceptions without providing the correct information. I'll try to update that.

security/src/main/java/org/kaazing/gateway/security/auth/FileLoginModule.java
danibusu: I tried to answer below, throwing a LoginException from a login doesn't guarantee that it will show up in the  Http Security Filter. I can create the new LoginException and log that before throwing it, so that I don't duplicate the message. What do you think? I saw some login module implementations have a debug option to enable detailed logging, but I thought that using the debug level on the logger would simulate that.

NicoletaOita:
@dpwspoon Yes, the tests that @danibusu added cover my test cases. It would be useful however to have some gateway requirements related to these login modules. I could check again if there are any corner cases after we have documentation.

danibusu:
@dpwspoon I updated the code of the login modules based on your comments. I'm still working on the integration test observations.

security/src/test/java/org/kaazing/gateway/security/auth/BasicLoginModuleTest.java
dpwspoon: + 1 sounds good to me.  Thanks

dpwspoon:
@DoruM Are you going to review this?

PR #872, opened 2017-02-23
Add udp.align option

Anisotrop:


transport/nio/src/main/java/org/kaazing/gateway/transport/nio/internal/UdpAlignFilter.java
```diff
@@ -0,0 +1,71 @@
...
+    private final Integer align;
+
+    public UdpAlignFilter(Logger logger, Integer align, IoSession session) {
+        this.logger = logger;
+        this.align = align;
+    }
+
+    @Override
+    public void onPostAdd(IoFilterChain parent, String name, NextFilter nextFilter) throws Exception {
+        IoSession session = parent.getSession();
```
dpwspoon: probably don't need to override/implement this method

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/internal/UdpAlignFilter.java
```diff
@@ -0,0 +1,71 @@
...
+        if (logger.isDebugEnabled()) {
+            logger.debug(String.format("Setting align %d on session %s ", align, session));
+        }
+        // TODO should do anything ?
+    }
+
+    @Override
+    protected void doSessionIdle(NextFilter nextFilter, IoSessionEx session, IdleStatus status) throws Exception {
+
+        if (status == BOTH_IDLE) {
```
dpwspoon: ditto

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/internal/UdpAlignFilter.java
Anisotrop: thanks @dpwspoon

Anisotrop:
Do you have any idea on where should the resource address option be checked and added on the connector side ?

dpwspoon:
I don't think we need to have any **behavior** for the connector, or at least I'm not sure what the behavior would be for outbound connections, so we can document that it has no effect on that.  Do you agree, or do you think we should also align outbound bytes?

We do need it as a <connect-option> though, specifically for reverse connectivity, where it gets put on the connect-options, but is really used on udp ACCEPTOR.

Anisotrop:
From here: https://tools.ietf.org/html/rfc5766#section-2.5 it would look like a peer will send turn channel data messages to a client using the channel number. Since the proxy is in the middle it would mean that the data comes on the outbound connection.
> 
>  In the reverse direction, when Peer A sends a UDP datagram to the
>    relayed transport address, this UDP datagram arrives at the server on
>    the relayed transport address assigned to the allocation.  Since the
>    UDP datagram was received from Peer A, which has a channel number
>    assigned to it, the server encapsulates the data into a ChannelData
>    message when sending the data to the client.

Anisotrop:
After giving more thought, the messages coming from the TURN Server without padding could be considered to always come on udp. I don't think we have anything between the server and the proxy that might alter the data or that we use an underlying protocol. 

This would mean that the require.padding option would only apply to the TurnFrameDecoder initiated by the acceptor while the connector side should have the following logic:
- buffer has exact length as declared we are most certainly on UDP so decode
- buffer has more than declared, should assume padding and decode accordingly

We might even try to detect the underlying transport since the assumption is the connection is direct. The lmitation is that we cannot add on the connector side an intermediary gateway that connects to the turn server on udp but has the accept on tcp.

I think the problem comes because of the difference in the transports, where multiple tcp segments inside a buffer can trigger a read, meaning we can only see a stream of bytes, while only a single udp datagram will trigger a read. There is a protocol over IP that has the reliability of tcp but is message oriented: https://en.wikipedia.org/wiki/Stream_Control_Transmission_Protocol

Waiting for your thoughts on this, I will be focusing on implementing the UdpAlignFilter.

dpwspoon:
I think you are correct, that we need UDP align for inbound packets on the outbound/connector.  And then the logic can be added probably as a filter put on by the connector

transport/nio/src/test/java/org/kaazing/gateway/transport/udp/specification/UdpConnectorIT.java
```diff
@@ -166,15 +177,52 @@ protected void doMessageReceived(IoSessionEx session, Object message) {
...
+    @Test
+    @Specification("additions/align.content/server")
+    public void alignData() throws Exception {
+        ResourceAddressFactory addressFactory = ResourceAddressFactory.newResourceAddressFactory();
+        ResourceOptions resourceOptions = ResourceOptions.FACTORY.newResourceOptions();
+        resourceOptions.setOption(UdpResourceAddress.ALIGN, 4);
+        ResourceAddress connectAddress = addressFactory.newResourceAddress("udp://127.0.0.1:8080", resourceOptions);
+        AtomicBoolean bytesAligned = new AtomicBoolean(true);
+        CountDownLatch messagesReceived = new CountDownLatch(2);
+        ConnectFuture connectFuture = connector.connect(connectAddress, new IoHandlerAdapter<IoSessionEx>(){
```
dpwspoon: Are you asserting somewhere that the actual message was aligned?

transport/nio/src/test/scripts/org/kaazing/specification/udp/rfc768/additions/align.content/client.rpt
```diff
@@ -0,0 +1,22 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+connect await BOUND
+        udp://localhost:8080
+connected
+
+write [0x00 0x01 0x02]
```
dpwspoon: I'm not sure if the behavior for multiple writes is strongly specified on K3po.  I.e. does each line go on a new packet or may they be combined?   Because I believe this is not defined (though maybe it is @StCostea), I could see this potentially changing out from you in the future, either by API/k3po changes or by OS or timing if it truly meant to be undefined.  To fully lock it down, I would suggest putting writes on a single line and then doing a write flush inbetween writes if you need multiple writes.
```
write flush
```

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/internal/datagram/UdpAlignFilter.java
```diff
@@ -0,0 +1,63 @@
...
+            if (buffer.remaining() % this.align > 0) {
+                if (logger.isTraceEnabled()) {
+                    logger.trace(String.format("Received message with length: %d, aligning on %d bytes", buffer.remaining(), align));
+                }
+                final int padLength = this.align - (buffer.remaining() % this.align);
+                final byte[] padding = new byte[padLength];
+                Arrays.fill(padding, (byte) 0x00);
+                final IoBufferAllocatorEx<?> allocator = session.getBufferAllocator();
+                IoBufferEx alignedBuffers = allocator.wrap(allocator.allocate(buffer.remaining() + padLength));
+                alignedBuffers.setAutoExpander(allocator);
```
dpwspoon: This seems ok to me, but there might be a better API or way to add bytes (remove the need for `put(buffer)`).

@jitsni do you think this could be improved?

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/internal/datagram/UdpAlignFilter.java
```diff
@@ -0,0 +1,63 @@
...
+
+import org.apache.mina.core.buffer.IoBuffer;
+import org.apache.mina.core.session.IoSession;
+import org.kaazing.gateway.transport.IoFilterAdapter;
+import org.kaazing.mina.core.buffer.IoBufferAllocatorEx;
+import org.kaazing.mina.core.buffer.IoBufferEx;
+import org.kaazing.mina.core.session.IoSessionEx;
+import org.slf4j.Logger;
+
+/**
```
dpwspoon: nit: wrong javadoc comment

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/internal/datagram/NioDatagramConnector.java
```diff
@@ -110,6 +115,14 @@ protected String getTransportName() {
     }
 
     @Override
+    protected void registerConnectFilters(ResourceAddress address, IoSession session) {
+        int align = address.getOption(UdpResourceAddress.ALIGN);
+        if (align > 0) {
```
dpwspoon: This seems correct to me!

transport/nio/src/main/java/org/kaazing/gateway/transport/nio/internal/AbstractNioConnector.java
```diff
@@ -143,79 +147,47 @@ public void connectDestroy(ResourceAddress address) {
...
-        future = connector.connect(inetAddress, new IoSessionInitializer<T>() {
+        ConnectFuture future = connector.connect(inetAddress, new IoSessionInitializer<T>() {
             @Override
             public void initializeSession(IoSession session, T future) {
-
                 if (logger.isTraceEnabled()) {
                     logger.trace(format("AbstractNioConnector.connectInternal()$initializeSession(), session: %s, resource: %s", session, resource));
                 }
-
+                registerConnectFilters(address, session);
```
dpwspoon: I'm ok with adding the register*Filters, But I'm not overly familiar with this code.  So I would want to defer to @jitsni judgment.  

Another alternative would be to make the "align" feature not UDP specific.  Then only expose public configuration to set it on udp.  That would eliminate the need for this new API.  I probably would have done it that way initially but there are tradeoffs with both the approaches mentioned here.

transport/nio/src/test/java/org/kaazing/gateway/transport/udp/specification/UdpConnectorIT.java
Anisotrop: Yes, I'm checking the message length is a multiple of 4 ` bytesAligned.set(((IoBuffer) message).remaining() % 4 == 0);`

transport/nio/src/test/scripts/org/kaazing/specification/udp/rfc768/additions/align.content/client.rpt
Anisotrop: I thought each write on udp would generate a datagram on the wire. Tried with flush but it seems to hang.

dpwspoon:
This looks good to me, is it ready to merge?

PR #874, opened 2017-02-27
Implement update.check as a property

stanculescu:
I changed the update check service to be set like a property (UPDATE_CHECK - default true) in gateway configuration, added unit and integration tests and deprecated the classes used for the service. I also added a new property SERVICE_URL that helps us to test easily the gateway behaviour. If the property is null(default value) it will be set to version.kaazing.com or version.kaazing.org

Issue https://github.com/kaazing/roadmap/issues/1076


server/src/main/java/org/kaazing/gateway/server/GatewayObserver.java
```diff
@@ -151,6 +157,18 @@ public void startingGateway(GatewayContext gatewayContext) {
             injectResources(gatewayListenerSpi, gatewayContext.getInjectables());
         }
 
+        Properties configuation = (Properties) gatewayContext.getInjectables().get("configuration");
```
DoruM: This whole section is not implemented correctly. I believe, according to [this comment](https://github.com/kaazing/roadmap/issues/1076#issuecomment-280720268) that a new `GatewayObserveFactorySpiPrototype` should be extended. Within that new class, the function `initingGateway` should be implemented and the code that you add here should be placed there. 

The `GatewayObserver` class should not be modified.

server/src/test/java/org/kaazing/gateway/server/update/check/UpdateCheckTest.java
```diff
@@ -0,0 +1,234 @@
...
+import org.jmock.Expectations;
+import org.jmock.Mockery;
+import org.jmock.api.Invocation;
+import org.jmock.lib.action.CustomAction;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.kaazing.gateway.util.scheduler.SchedulerProvider;
+
+public class UpdateCheckTest {
```
DoruM: Please add unit tests for `setUpdateCheck` as well

server/src/main/java/org/kaazing/gateway/server/update/check/UpdateCheck.java
```diff
@@ -0,0 +1,129 @@
...
+
+import org.kaazing.gateway.util.InternalSystemProperty;
+import org.kaazing.gateway.util.scheduler.SchedulerProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Creates and manages periodic checks to see if the gateway has updates
+ */
+public class UpdateCheck {
```
DoruM: This class should be the one extending from `GatewayObserverFactorySpiPrototype`, and override the `initingGateway` method.

dpwspoon:
Additionally, I would highly recommend putting this in a new module.  This may not be necessary but is for sure necessary if you are going to include a gateway.server jar in the surefire test, as done here: 

```
+                <plugin>
 +                    <groupId>org.apache.maven.plugins</groupId>
 +                    <artifactId>maven-surefire-plugin</artifactId>
 +                    <configuration>
 +                        <additionalClasspathElements>
 +                            <additionalClasspathElement>${basedir}/src/test/resources/gateway.server-5.0.0.8.jar</additionalClasspathElement>
 +                        </additionalClasspathElements>
 +                        <argLine>${surefireArgLine}</argLine>
 +                        <skipTests>${skipTests}</skipTests>
 +                    </configuration>
 +                </plugin>
```

server.api/src/main/java/org/kaazing/gateway/server/impl/VersionUtils.java
```diff
@@ -148,6 +148,8 @@ public static String getGatewayProductDependencies() {
      *
      */
     private static void getGatewayProductInfo() {
+        // FIXME does the following todo still hold ?
+        //
         // TODO: Now that we've switched the products to include
```
dpwspoon: I think this is false, but I'm not 100% sure

update.check/src/main/java/org/kaazing/gateway/update/check/UpdateCheckTask.java
```diff
@@ -0,0 +1,105 @@
...
+        String updateVersionUrl = getVersionServiceUrl();
+        try {
+            URL url = new URL(updateVersionUrl);
+            HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+            connection.setRequestMethod("GET");
+            connection.setRequestProperty("User-Agent", "Kaazing Update Property");
+            connection.setRequestProperty("Connection", "close");
+            connection.setRequestProperty("Accept", "text/plain");
+            int responseCode = connection.getResponseCode();
+            if (responseCode >= 200 && responseCode <= 300) {
```
DoruM: This if should be reformated:
```
  if (responseCode >= 200 && responseCode <= 300) {
   logger.warn .... // add various log messages here
  return null;
}
```
Then, remove the the lines 86-88 below.

update.check/src/main/java/org/kaazing/gateway/update/check/UpdateCheckTask.java
```diff
@@ -0,0 +1,105 @@
...
+                try (BufferedReader br = new BufferedReader(new InputStreamReader(connection.getInputStream(), "UTF-8"))) {
+                    JSONTokener tokener = new JSONTokener(br);
+                    JSONObject root = new JSONObject(tokener);
+                    String version = root.getString("version");
+                    latestVersion = parseGatewayVersion(version);
+                }
+            } else {
+                throw new Exception(format("Unexpected %d response code from versioning property", responseCode));
+            }
+        } catch (Exception e) {
```
DoruM: Qualify the exception.

update.check/src/main/java/org/kaazing/gateway/update/check/UpdateCheckTask.java
```diff
@@ -0,0 +1,105 @@
...
+            int responseCode = connection.getResponseCode();
+            if (responseCode >= 200 && responseCode <= 300) {
+                try (BufferedReader br = new BufferedReader(new InputStreamReader(connection.getInputStream(), "UTF-8"))) {
+                    JSONTokener tokener = new JSONTokener(br);
+                    JSONObject root = new JSONObject(tokener);
+                    String version = root.getString("version");
+                    latestVersion = parseGatewayVersion(version);
+                }
+            } else {
+                throw new Exception(format("Unexpected %d response code from versioning property", responseCode));
```
DoruM: Never throw Exception (this should be removed in the end as a result of the reformatting described on line 79

update.check/src/main/java/org/kaazing/gateway/update/check/UpdateCheckListener.java
```diff
@@ -0,0 +1,29 @@
...
+
+    /**
+     * Notifies the listener that a new @GatewayVersion is available
+     * Listeners may be notified more than once about the same update
+     *
+     * @param currentVersion
+     * @param latestGatewayVersion
+     */
+    void newVersionAvailable(GatewayVersion currentVersion, GatewayVersion latestGatewayVersion);
+
```
DoruM: Have we  removed `setUpdateCheckService` ? Is this function not needed anymore?

stanculescu: I removed setUpdateCheckService. For the service the method was overridden in the HttpManagementUpdateCheck class, but for the property the management is not required so the method has no implementation. Sorry for didn't notice that from the first time.

stanculescu:
@DoruM Thanks for the review.

server.api/src/main/java/org/kaazing/gateway/server/impl/VersionUtils.java
```diff
@@ -158,40 +163,51 @@ private static void getGatewayProductInfo() {
...
+                foundJar = true;
+            }
+        }
+
+        if (artifact == null) {
+            return;
+        }
+
+        try {
+            JarFile jar = new JarFile(artifact);
```
DoruM: Close this jar file after line 184.

server.api/src/main/java/org/kaazing/gateway/server/impl/VersionUtils.java
```diff
@@ -23,12 +23,16 @@
 import java.util.jar.JarFile;
 import java.util.jar.Manifest;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public final class VersionUtils {
 
     public static String PRODUCT_TITLE;
```
DoruM: Make this field and the following 3 final.

update.check/src/main/java/org/kaazing/gateway/update/check/UpdateCheckGatewayObserver.java
```diff
@@ -0,0 +1,111 @@
...
+
+        productName = getGatewayProductTitle().replaceAll("\\s+", "");
+        currentVersion = parseGatewayVersion(getGatewayProductVersionPatch());
+
+        String serviceUrl = InternalSystemProperty.SERVICE_URL.getProperty(properties);
+        if (serviceUrl != null) {
+            versionServiceUrl = serviceUrl;
+        } else {
+            versionServiceUrl =
+                    (getGatewayProductEdition().toLowerCase().contains("enterprise")) ? ENTERPRISE_URL : COMMUNITY_URL;
```
DoruM: nit.: The outer parentheses are extraneous. ?: operator has the lowest precedence, so it will execute last.

update.check/src/main/java/org/kaazing/gateway/update/check/UpdateCheckListener.java
DoruM: If HQ agrees that `setUpdateCheckService` is not needed anymore here, it would be best to annotate this interface with the @FunctionalInterface annotation.

update.check/src/main/java/org/kaazing/gateway/update/check/UpdateCheckTask.java
```diff
@@ -0,0 +1,104 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * An update check task that updates the latest version
+ */
+public class UpdateCheckTask implements Runnable {
+
+    private final Logger logger = LoggerFactory.getLogger(UpdateCheckTask.class);
+    private final UpdateCheckGatewayObserver updateCheckGatewayObserver;
+    private final String protocolVersion = "1.0";
```
DoruM: The `protocolVersion` field should be static.

util/src/main/java/org/kaazing/gateway/util/InternalSystemProperty.java
```diff
@@ -129,6 +128,11 @@
     GATEWAY_IDENTIFIER
             ("org.kaazing.gateway.server.GATEWAY_IDENTIFIER", ""),
 
+    // Internal system property checking if a newer version of the Gateway is available
+    UPDATE_CHECK("org.kaazing.gateway.server.UPDATE_CHECK", "true"),
+
+    SERVICE_URL("org.kaazing.gateway.server.SERVICE_URL"),
```
dpwspoon: This name doesn't state what it is used for at all.  Please rename to say something about UPDATE_CHECK, maybe UPDATE_CHECK_URL or UPDATE_CHECK_SERVICE_URL.

It's ok to have this property, but is there any real use case other than internal testing?  If yes there is a use case then it iss very good we have it here.  

If not and it is only used for testing, then if it is possible it would probably be better to not have it here and not tied to a system property.

stanculescu: I used service url as a property because it is difficult to test the gateway behaviour. The value is hard-coded in UpdateCheckGatewayObserver class. So in the integration test for the update check property, when I start the gateway I can't set the url and the real one will be access. This is the reason I used a new property. 

Also I think it's useful to have the property, if for any reasons the url will be modify the client can set the value in configuration and don't have to wait for a new release.

dpwspoon: ok, please rename to `UPDATE_CHECK_SERVICE_URL` or `UPDATE_CHECK_URL`, service is an ambiguous term due to several different contexts it can be used in

PR #876, opened 2017-03-01
Fix for NPE generated when the scheme would be ommited in <accept>

zahariaca:
If the scheme would be omited in the gateway.config.xml this would result in a NPE message when trying to start the gateway

### DIFF
--- a/resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/uri/URIUtils.java
+++ b/resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/uri/URIUtils.java
@@ -27,7 +27,7 @@ import org.kaazing.gateway.resource.address.URLUtils;
  *
  */
 public final class URIUtils {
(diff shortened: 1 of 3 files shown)


resource.address/spi/src/main/java/org/kaazing/gateway/resource/address/uri/URIUtils.java
```diff
@@ -57,6 +57,9 @@ public static String uriToString(NetworkInterfaceURI uri) {
     public static String getHost(String uriString) {
         try {
             URI uri = new URI(uriString);
+            if(uri.getHost()==null) {
+                throw new IllegalArgumentException("URI syntax invalid. Protocol, host and port are all required: " + uriString);
```
robinzimmermann: The port is optional, so the message isn't quite correct. And I think the correct term is "scheme", not "protocol". How about something like this?:

> Invalid URI syntax. Scheme and host must be provided (port number is optional):

zahariaca:
Thank you, I have changed the error message.

Issue #822, opened 2016-12-13
Build failure in wseb transport: WsebConnectorLoggingIT.shouldLogOpenAndCleanServerClose unexpected invocation: ioHandler.exceptionCaught(


cmebarrow:
Happened in Travis build https://travis-ci.org/kaazing/gateway/builds/183751819. Console output for the failing test:
[WsebConnectorLoggingIT.fail.txt](https://github.com/kaazing/gateway/files/650470/WsebConnectorLoggingIT.fail.txt)

Extract:
```
WsebConnectorLoggingIT.shouldLogOpenAndCleanServerClose FAILED with exception java.lang.AssertionError: unexpected invocation: ioHandler.exceptionCaught(<(#00000104: kzg wseb, client, ws://localhost:8080/path?query => ws://localhost:8080/path?query)>, <java.io.IOException: Network connectivity has been lost or transport was closed at other end>)
```

cmebarrow:
Happened again just now in a gateway build on Travis

Anisotrop:
Happened again: https://travis-ci.org/kaazing/gateway/builds/219315787

PR #877, opened 2017-03-02
Changed GatewayObserverApi interface to have default methods 

zahariaca:
Methods in `GatewayObserverApi` now are default.
- `GatewayObserverFactorySpi` & `GatewayObserverFactorySpiPrototype` are not needed anymore and have been removed.
- `ManagementGatewayObserver` has been changed to implement the interface.
- `GatewayObserver` has been changed to use the interface instead of  `GatewayObserverFactorySpi` to better adhere to the directive "program to an interface".


server/src/main/java/org/kaazing/gateway/server/GatewayObserver.java
```diff
@@ -150,7 +150,7 @@ public void destroyedService(ServiceContext serviceContext) {
 
     @Override
     public void startingGateway(GatewayContext gatewayContext) {
-        for (GatewayObserverFactorySpi gatewayListenerSpi : gatewayListenerSpi) {
+        for (GatewayObserverApi gatewayListenerSpi : gatewayListenerSpi) {
```
Anisotrop: Should rename the first `gatewayListenerSpi` into `gatewayListener`.

zahariaca: Changed.

Anisotrop:
The failed build is cause by: https://github.com/kaazing/gateway/issues/822

PR #882, opened 2017-03-10
make injected login modules return false on success

danibusu:
- login method now returns false on success, in order to not interfere
with the outcome of the whole login process
- update timeout login module to not clean up the session timeout from
the login result because, as a consequence of the previous change, the
commit() method will call the logout0() method and not commit0(),
disabling the session timeout everytime
- updated tests


dpwspoon:
LGTM, waiting for travis to pass

PR #884, opened 2017-03-14
Fix regression on update.check 

stanculescu:
I added missing dependencies in generic-bin and bom/pom and I added a test for VersionUtils.


server.api/src/main/java/org/kaazing/gateway/server/impl/VersionUtils.java
```diff
@@ -163,52 +163,48 @@ private static void getGatewayProductInfo() {
         }
 
         boolean foundJar = false;
-        String artifact = null;
         String[] pathEntries = System.getProperty("java.class.path").split(System.getProperty("path.separator"));
         HashMap<String, Attributes> products = new HashMap<>(7);
         HashSet<String> removals = new HashSet<>(7);
-        for (String pathEntry : pathEntries) {
-            if (pathEntry.contains("gateway.server.test") || (pathEntry.contains("gateway.server") && artifact == null)) {
```
dpwspoon: It would be best not to have test code if statements in production code

bom/pom.xml
```diff
@@ -66,6 +66,12 @@
                 <version>${project.version}</version>
             </dependency>
 
+            <dependency>
```
dpwspoon: This will need matching changes on enterprise.gateway

dpwspoon:
Can/did we remove the update.check service from the shipping gateway-config.xml?

server.api/src/main/java/org/kaazing/gateway/server/impl/VersionUtils.java
```diff
@@ -224,16 +193,40 @@ private static void getGatewayProductInfo() {
             productVersion = null;
             productEdition = null;
             productDependencies = null;
+            if (LOG.isTraceEnabled()) {
+                LOG.trace(format("Elected default"));
```
DoruM: Remove "format" call here.

server.api/src/main/java/org/kaazing/gateway/server/impl/VersionUtils.java
```diff
@@ -151,71 +132,59 @@ public static String getGatewayProductDependencies() {
...
+                if (mf == null) {
+                    continue;
+                }
+                Attributes attrs = mf.getMainAttributes();
+                if (attrs == null) {
+                    continue;
+                }
+                String title = attrs.getValue("Implementation-Title");
+                String version = attrs.getValue("Implementation-Version");
+                String product = attrs.getValue("Kaazing-Product");
```
DoruM: The String "Kaazing-Product" is used 3 times. Define a constant for it

server.api/src/main/java/org/kaazing/gateway/server/impl/VersionUtils.java
```diff
@@ -151,71 +132,59 @@ public static String getGatewayProductDependencies() {
...
+                String dependencies = attrs.getValue("Kaazing-Dependencies");
```
DoruM: This should be done if only if any of product, title, version is non-null. As such, it should be done somewhere before line 172.

server.api/src/main/java/org/kaazing/gateway/server/impl/VersionUtils.java
```diff
@@ -151,71 +132,59 @@ public static String getGatewayProductDependencies() {
...
-
-                // Store the list of products found, but remove any products
-                // marked as dependencies (i.e. products on which the current
-                // product depends.  We want to find the product that nothing
-                // else depends on.
-                products.put(product != null ? product : title, attrs);
-                if (dependencies != null) {
-                    String[] deps = dependencies.split(",");
-                    Collections.addAll(removals, deps);
+            try (JarFile jar = new JarFile(pathEntry)) {
```
DoruM: Extract this entire try block as a method. The method should probably return a boolean, and we should do `continue` on a `false` return value

DoruM:
VersionUtils still has some issues. However, tickets were opened for those.

Issue #818, opened 2016-12-07
PipeOverSocksOverUdpReverseIT.establishConnection run has ClassCastException


jitsni:
The test is passing but I see some exception in travis log:
```
PipeOverSocksOverUdpReverseIT.establishConnection success (1.25 secs)
clientSentData(com.kaazing.gateway.transport.socks.PipeOverSocksOverUdpReverseIT) starting
java.lang.ClassCastException: org.jboss.netty.channel.socket.nio.NioChildDatagramChannel cannot be cast to org.jboss.netty.channel.socket.nio.NioDatagramChannel
	at org.jboss.netty.channel.socket.nio.AbstractNioWorker$3.run(AbstractNioWorker.java:628)
	at org.jboss.netty.channel.socket.nio.AbstractNioSelector.processTaskQueue(AbstractNioSelector.java:441)
	at org.jboss.netty.channel.socket.nio.AbstractNioSelector.run(AbstractNioSelector.java:358)
	at org.jboss.netty.channel.socket.nio.AbstractNioWorker.run(AbstractNioWorker.java:135)
	at org.jboss.netty.channel.socket.nio.NioWorker.run(NioWorker.java:253)
	at org.jboss.netty.util.ThreadRenamingRunnable.run(ThreadRenamingRunnable.java:108)
	at org.jboss.netty.util.internal.DeadLockProofWorker$1.run(DeadLockProofWorker.java:42)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

PR #885, opened 2017-03-15
Avoid a ClassCastException when channel is NioChildDatagramChannel.

StCostea:
Adding a dummy ReadDispatcher specific for NioChildDatagramChannel. Its dispatch() method is never called.


dpwspoon:
@jitsni will you review this please

StCostea:
@jitsni Did you have a chance to check this one ? Thanks.

mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioWorker.java
```diff
@@ -629,9 +626,13 @@ public void run() {
                     int interestOps = channel.getInternalInterestOps();
                     interestOps |= SelectionKey.OP_WRITE;
                     channel.setInternalInterestOps(interestOps);
-                    ReadDispatcher readDispatcher = channel instanceof NioSocketChannel
-                            ? new TcpReadDispatcher((NioSocketChannel) channel)
-                            : new UdpReadDispatcher((NioDatagramChannel) channel);
+                    ReadDispatcher readDispatcher = null;
```
dpwspoon: please format and include brackets on the ifs, this is especially ugly because you indent on the if to where the bracket would be as if the else if is suppose to be under the general if statement

mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/NioWorker.java
```diff
@@ -578,4 +578,23 @@ public boolean dispatch(NioWorker worker, SelectionKey key) {
...
+        }
+
+        @Override
+        public AbstractNioChannel<DatagramChannel> channel() {
+            return channel;
+        }
+
+        @Override
+        public boolean dispatch(NioWorker worker, SelectionKey key) {
+//            return worker.readUdp(key, channel.);
```
dpwspoon: why have this comment

dpwspoon:
@jitsni please review once @StCostea has updated the code

mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioWorker.java
StCostea: The indentation was ok. But there were few tabs in there that messed it up in github. I don't know from where I have those tabs from time to time...

jitsni:
Also, it is related to https://github.com/kaazing/gateway/issues/818

jitsni:
The fix avoid the ClassCastException, but I wonder if it masks any actual problem

mina.netty/src/main/java/org/jboss/netty/channel/socket/nio/AbstractNioWorker.java
```diff
@@ -629,9 +626,13 @@ public void run() {
                     int interestOps = channel.getInternalInterestOps();
```
jitsni: I think you need the following:
                    if (channel instanceof NioChildDatagramChannel) {
                        return;
                    }
It was there in the previous commits but got removed incorrectly. See the commit that removed it:
432b55c49e29ed9fed27bcd6a9e09ee5dd9696c7

StCostea: The commit you gave did not affect that file. I found the change in commit https://github.com/kaazing/gateway/commit/105d1c6c39419c95840b956522c035c2e0c10652

If I'm adding this piece of code in register and deregister, it means that I can remove all the changes that I did ? I'll try to check it.

StCostea: I ran some tests, it seems that your suggested change is fixing the issue without requiring the changes I did. I suggest to close this PR without merging and opening a new one only with those changes.

jitsni:
Closing this one as a new PR will address the issue.

PR #888, opened 2017-03-23
Adds http.redirect service

DoruM:
This service allows HTTP redirects to HTTPS url when they access a secure service on port 80.  This is a common case for HTTP Proxy configurations which we are now supporting.

### DIFF
--- a/server/src/main/xsd/gateway-config-201606.xsd
+++ b/server/src/main/xsd/gateway-config-201606.xsd
@@ -1191,9 +1191,6 @@
 
                     <!-- connect is not supported for redirect -->
 
-                    <element maxOccurs="unbounded" minOccurs="0"
-                             name="balance" type="anyURI"/>
-
                     <element fixed="http.redirect" maxOccurs="1" minOccurs="1"
                              name="type"
                              type="gateway:CollapsedString"/>
@@ -1214,9 +1211,6 @@
                              name="authorization-constraint"
                              type="gateway:AuthorizationConstraintType"/>
 
-                    <element maxOccurs="unbounded" minOccurs="0"
-                             name="cross-site-constraint"
-                             type="gateway:CrossSiteConstraintType"/>
                 </sequence>
             </restriction>
         </complexContent>
--- a/server/src/main/xsd/gateway-config-201606.xsd
+++ b/server/src/main/xsd/gateway-config-201606.xsd
@@ -1168,6 +1168,60 @@
         </complexContent>
     </complexType>
 
+    <complexType name="redirectServiceType">
+        <annotation>
+            <documentation>This type of service is for redirecting gateway.
+            </documentation>
+        </annotation>
+
+        <complexContent>
+            <restriction base="gateway:ServiceType">
+                <sequence>
+                    <element maxOccurs="1" minOccurs="1"
+                             name="name"
+                             type="gateway:CollapsedString"/>
+
+                    <element maxOccurs="1" minOccurs="0"
+                             name="description"
+                             type="gateway:CollapsedString"/>
+
+                    <element maxOccurs="unbounded" minOccurs="1"
+                             name="accept"
+                             type="anyURI"/>
+
+                    <!-- connect is not supported for redirect -->
+
+                    <element maxOccurs="unbounded" minOccurs="0"
+                             name="balance" type="anyURI"/>
+
+                    <element fixed="http.redirect" maxOccurs="1" minOccurs="1"
+                             name="type"
+                             type="gateway:CollapsedString"/>
+
+                    <element maxOccurs="1" minOccurs="1"
+                             name="properties"
+                             type="gateway:redirectServicePropertiesType"/>
+
+                    <element maxOccurs="1" minOccurs="0"
+                             name="accept-options"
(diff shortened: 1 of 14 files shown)


server/src/main/xsd/gateway-config-201606.xsd
```diff
@@ -1168,6 +1168,60 @@
...
+                             name="description"
+                             type="gateway:CollapsedString"/>
+
+                    <element maxOccurs="unbounded" minOccurs="1"
+                             name="accept"
+                             type="anyURI"/>
+
+                    <!-- connect is not supported for redirect -->
+
+                    <element maxOccurs="unbounded" minOccurs="0"
```
dpwspoon: Is there any real use case for balance??

server/src/main/xsd/gateway-config-201606.xsd
```diff
@@ -1168,6 +1168,60 @@
...
+
+                    <element maxOccurs="1" minOccurs="0"
+                             name="realm-name"
+                             type="gateway:CollapsedString"/>
+
+                    <element maxOccurs="unbounded" minOccurs="0"
+                             name="authorization-constraint"
+                             type="gateway:AuthorizationConstraintType"/>
+
+                    <element maxOccurs="unbounded" minOccurs="0"
```
dpwspoon: Does cross-site-constraints mean anything for 302

service/http.redirect/src/main/java/org/kaazing/gateway/service/http/redirect/HttpRedirectServiceHandler.java
```diff
@@ -0,0 +1,121 @@
...
+        RFC822_FORMAT_PATTERN.setTimeZone(TimeZone.getTimeZone("GMT"));
+    }
+
+    HttpRedirectServiceHandler(Logger logger) {
+        this.logger = logger;
+    }
+
+    @Override
+    public void doSessionCreated(HttpAcceptSession session) throws Exception {
+        // NOOP no license check needed
```
dpwspoon: I don't think  `no license check needed` are needed, is this from copy paste maybe.

Nit: Also, I don't think there is great value in having the `@Override NOOP` method, though I'm ok with them.  I don't want to dig into the API, but perhaps there is a more complete Adapter to use that doesn't require overriding them

service/http.redirect/src/main/java/org/kaazing/gateway/service/http/redirect/HttpRedirectServiceHandler.java
```diff
@@ -0,0 +1,121 @@
...
+        session.close(true);
+    }
+
+    @Override
+    protected void doSessionOpened(HttpAcceptSession session) throws Exception {
+        // Get the method used for this request; if not a GET, refuse the
+        // request (KG-1233).
+        // (KG-11211) Enabling HEAD method too.
+        HttpMethod method = session.getMethod();
+        if (method != HttpMethod.GET && method != HttpMethod.HEAD) {
```
dpwspoon: Why just these methods: copy paste from directory service or actual reason.  IIRC correctly, user agents (browsers) might have different behavior on 3xx based on methods, but I don't think that changes what the redirect service should do

dpwspoon:
Don't forget to add the jar to the distribution

server/src/main/xsd/gateway-config-201606.xsd
DoruM: No.

dpwspoon: let's remove it then

service/http.redirect/src/main/java/org/kaazing/gateway/service/http/redirect/HttpRedirectServiceHandler.java
DoruM: Copy paste from directory service. I haven't found anything related to methods in this particular case. I shall remove them.

server/src/main/xsd/gateway-config-201606.xsd
DoruM: Removed.

service/http.redirect/src/main/java/org/kaazing/gateway/service/http/redirect/HttpRedirectServiceHandler.java
```diff
@@ -64,12 +63,6 @@ protected void doSessionOpened(HttpAcceptSession session) throws Exception {
         // Get the method used for this request; if not a GET, refuse the
```
dpwspoon: Nit, please remove comment

dpwspoon:
LGTM, other than comment that should be removed and has no relevance:
```
// Get the method used for this request; if not a GET, refuse the
// request (KG-1233).
// (KG-11211) Enabling HEAD method too.
```

sbadugu:
@DoruM 
The following test is failing on travis:
```
Tests in error: 
  HttpRedirectServiceIT.shouldAcceptSecureLocation Â» Specification Internal erro...

Tests run: 1, Failures: 0, Errors: 1, Skipped: 0
```

DoruM:
Yes, because K3PO was not merged. I shall remove it from K3PO scripts and add it to this module.

PR #891, opened 2017-04-05
Fix for ClassCastException 

StCostea:
This is a follow-up for https://github.com/kaazing/gateway/pull/885. The new fix is more simple and clearer.


jitsni:
Looks good to me.

PR #892, opened 2017-04-05
Log rule bug

StCostea:


service/http.proxy/src/test/java/org/kaazing/gateway/service/http/proxy/HttpProxyLoggingWarnIT.java
```diff
@@ -55,16 +51,10 @@
     public HttpProxyLoggingWarnIT(String serviceName, String expectedMessage) {
         this.serviceName = serviceName;
         this.expectedMessage = expectedMessage;
-        this.chain = RuleChain.outerRule(new MethodExecutionTrace()).around(robot).around(checkLogMessageRule).around(getGatewayRule());
+        this.chain = RuleChain.outerRule(new MethodExecutionTrace()).around(getGatewayRule()).around(checkLogMessageRule).around(robot);
```
dpwspoon: Does the order matter?

test.util/src/main/java/org/kaazing/test/util/LoggingTestRule.java
```diff
@@ -0,0 +1,65 @@
...
+    private String filterPattern;
+    private long timeOutMilis = 2000; // default value
+
+    private static final long SLEEP_STEP = 50; // how much to sleep between retries
+
+    @Override
+    public Statement apply(Statement base, Description description) {
+        return new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
```
dpwspoon: Please document what race you are working around here in a comment.  Presumably, the gateway killing itself prior to logging every necessary item?

test.util/src/main/java/org/kaazing/test/util/LoggingTestRule.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+package org.kaazing.test.util;
+
+import java.util.List;
+
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+public class LoggingTestRule implements TestRule {
```
dpwspoon: nit: I would suggest closer naming to the [shipped/public Rules](http://junit.org/junit4/javadoc/4.12/org/junit/rules/ExpectedException.html):

LoggingTestRule -> ExceptedLog? ExpectedLogRule? LoggingRule? LogVerifier (no need for Test)
setExpectedPatterns -> expectPatterns? 
setForbiddenPatterns -> forbiddenPatterns? forbidPatterns? never?

service/http.proxy/src/test/java/org/kaazing/gateway/service/http/proxy/HttpProxyLoggingWarnIT.java
StCostea: Yes. We need to let the gateway process all events. If we have the checkLogMessageRule  around the gateway, the gateway will be stopped before doing the check. No matter how long you wait, nothing will change as the gateway is already stopped.

test.util/src/main/java/org/kaazing/test/util/LoggingTestRule.java
StCostea: I have updated the names.

test.util/src/main/java/org/kaazing/test/util/LoggingTestRule.java
StCostea: The gateway rule/statement kills the gateway before the logs are created. I will add a comment.

test.util/src/main/java/org/kaazing/test/util/LoggingRule.java
```diff
@@ -0,0 +1,81 @@
...
+ * expected log lines before stopping the gateway. 
+ *
+ */
+public class LoggingRule implements TestRule {
+    private List<String> expectedPatterns;
+    private List<String> forbiddenPatterns;
+    private String filterPattern;
+    private long timeOutMilis = 2000; // default value
+
+    private static final long SLEEP_STEP = 50; // how much to sleep between retries
```
cmebarrow: Rename to indicate the unit, e.g. SLEEP_TIME_MILLIS

test.util/src/main/java/org/kaazing/test/util/LoggingRule.java
```diff
@@ -0,0 +1,81 @@
...
+        this.forbiddenPatterns = forbiddenPatterns;
+        return this;
+    }
+
+    public LoggingRule filterPattern(String filterPattern) {
+        this.filterPattern = filterPattern;
+        return this;
+    }
+    
+    public LoggingRule timeOutInMilis(long timeOutMilis) {
```
cmebarrow: I would suggest adding a time unit parameter to facilitate use and readability:
```
public LoggingRule timeOut(long time, TimeUnit timeUnit)
```
That way usage can be like `loggingRule.timeOut(5, SECONDS)` which is more readable than loggingRule.timeOutMillis(5000)

test.util/src/main/java/org/kaazing/test/util/LoggingRule.java
```diff
@@ -0,0 +1,81 @@
...
+ * 
+ * In case a GatewayRule is used, make sure that the GatewayRule is around the LoggingRule, to allow the LoggingRule to wait for the
+ * expected log lines before stopping the gateway. 
+ *
+ */
+public class LoggingRule implements TestRule {
+    private List<String> expectedPatterns;
+    private List<String> forbiddenPatterns;
+    private String filterPattern;
+    private long timeOutMilis = 2000; // default value
```
cmebarrow: Usual abbreviation of milliseconds is "millis" (with two L's)

cmebarrow:
@StCostea Once this is merged are you planning to make similar changes to Enterprise Gateway to use the new rule? That would be highly desirable I would think.

StCostea:
@cmebarrow I started to check EE (but did not finish) and did not find any situation where this is required. But yes, we do need to update EE also. After the merge for this PR I will do it.

As for the comments, sorry about that, the previous PR that I closed had the issue connected. I will connect this one also.

StCostea:
Sorry, it seems it was the wrong issue. :(

cmebarrow:
@StCostea Looks good. Please resolve the build failure then I or @dpwspoon can merge. Build failure was `[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:2.11:check (validate) on project test.util: Failed during checkstyle execution: There are 4 checkstyle errors. -`

StCostea:
I don't see the error in Travis. I will try to rebuild locally.

PR #896, opened 2017-04-13
Update login module for a token use case

robinzimmermann:


samples/security/TokenLoginModule.java
```diff
@@ -0,0 +1,256 @@
...
+                sharedState.forEach((k, v) -> debug(String.format("Commit:   %s=%s", k, v)));
+            }
+        }
+
+        return true;
+    }
+
+    @Override
+    public boolean abort() throws LoginException {
+        // Clear state here, if any was created, particularly anything stored in
```
danibusu: If this login module succeeded, but the overall process entered the abort phase, maybe we should also remove the token from the expiring state map? I'm thinking that the user won't be able to login again until the token expires otherwise, even though it was a valid token.

samples/security/TokenLoginModule.java
```diff
@@ -0,0 +1,256 @@
...
+    //
+    private void processToken(String tokenData) throws JSONException, LoginException {
+
+        JSONObject json = new JSONObject(tokenData);
+
+        // Validate that the token hasn't expired.
+        ZonedDateTime expires = ZonedDateTime.parse(json.getString("tokenExpires"));
+        ZonedDateTime now = ZonedDateTime.now(expires.getZone());
+        if (expires.isBefore(now)) {
+            throw new LoginException("Token has expired");
```
danibusu: Shouldn't we challenge to provide this information to the client?

samples/security/TokenLoginModule.java
```diff
@@ -0,0 +1,256 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.kaazing.gateway.server.spi.security.AuthenticationToken;
+import org.kaazing.gateway.server.spi.security.AuthenticationTokenCallback;
+import org.kaazing.gateway.server.spi.security.ExpiringState;
+import org.kaazing.gateway.server.spi.security.LoginModuleOptions;
+import org.kaazing.gateway.server.spi.security.LoginResult;
+import org.kaazing.gateway.server.spi.security.LoginResultCallback;
+
+// This is an example custom login module that reads a token to extract the username. If
```
robinzimmermann: I interpreted some statements from @dpwspoon that Javadoc style comments should be reserved for an API rather than an implementation.

samples/security/TokenLoginModule.java
```diff
@@ -0,0 +1,256 @@
...
+            debug(String.format("Login: Got an Exception: %s", e.getMessage()));
+            throw new LoginException(e.getMessage());
+        }
+
+    }
+
+    @Override
+    public boolean commit() throws LoginException {
+        // Save the username into shared state for downstream login modules to authorize.
+        debug(String.format("Commit: Saving username to shared state: %s", username));
```
robinzimmermann: You are spot on @danibusu! I realized this myself earlier this morning and I will resubmit with this fixed.

samples/security/TokenLoginModule.java
robinzimmermann: Yes. With the realization that the shared state should be set in `login()` rather than `commit()`, then it is correct to clear that shared state here in `abort()`. I will include that in the next commit.

samples/security/TokenLoginModule.java
robinzimmermann: No. The client was expected to provide a valid token and they failed by providing an expired token. Therefore authorization fails, resulting in a 403. The client can then choose to retry if they want.

We only challenge if they failed to provide a token at all.

robinzimmermann:
Reviewers: I have fixed the shared state usage.

I also provided some instructions on how you can easily run the login module here:

https://github.com/kaazing/roadmap/issues/1105#issuecomment-294003163

samples/security/TokenLoginModule.java
danibusu: As far as I know, Javadoc is also used for implementations. If is a Kaazing practice, then I can live with that.

samples/security/TokenLoginModule.java
danibusu: I was thinking that the reason for failure might help, but you're right. No need to do that.

samples/security/TokenLoginModule.java
```diff
@@ -0,0 +1,250 @@
...
+    private String username;
+
+    // Flag to enable or disable debugging output.
+    private boolean debug = false;
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public void initialize(Subject subject, CallbackHandler callbackHandler, Map<String, ?> sharedState, Map<String, ?> options) {
+
+        this.debug = ((String) options.get("debug")).equalsIgnoreCase("true");
```
danibusu: This should be `"true".equalsIgnoreCase((String) options.get("debug"))` to avoid a `NullPointerException` is the option is not set.

samples/security/TokenLoginModule.java
```diff
@@ -0,0 +1,250 @@
...
+                loginResult.challenge();
+                throw new LoginException("No tokens provided");
+            }
+            debug(String.format("Login: Obtained AuthenticationToken: %s", authToken));
+
+            // Ensure that the token data uses the "Token" authentication scheme.
+            // Note: Compare to "Token" rather than "Application Token" because the scheme
+            // is just the scheme type which is Basic, Negotiate, or Token.
+            String authScheme = authToken.getScheme();
+            if (!authScheme.equals("Token")) {
```
danibusu: Same here, use `!"Token".equals(authScheme)`

samples/security/TokenLoginModule.java
```diff
@@ -0,0 +1,250 @@
...
+    @Override
+    public boolean commit() throws LoginException {
+
+        // This method associates relevant Principals and Credentials with the Subject.
+        // This login module does not perform authorization, but extracts the username
+        // from a token for downstream login modules to authorize. Therefore there is
+        // nothing to commit here.
+
+        debug("Commit: Nothing to commit");
+
```
danibusu: Maybe add a commented out example of adding a principal, like you did in the logout() method?

samples/security/TokenLoginModule.java
robinzimmermann: You generally don't want to convey to the client what went wrong because it leaks information. If an attacker tries a username and password, for example, and gets an error that says "Invalid username" then they try a different username. If they get back "Invalid password", then they know they got the username right and now just need to focus on the password.

Generally you just want to tell the client it failed, and you can log more granular information on the server side, if desired.

samples/security/TokenLoginModule.java
robinzimmermann: Thanks, I never knew why people wrote it that way, and now I know!

samples/security/TokenLoginModule.java
robinzimmermann: Will do.

samples/security/TokenLoginModule.java
robinzimmermann: It's a tough call. The reason for the example in the `logout()` method is to show the if statement that checks if the `Subject` is read-only. People usually don't know to do that check.

If I put an example in the `login()` method I feel like it's encouraging people to use the `Subject`, which they shouldn't really be doing in this kind of a login module.

Either way is justified, but right now I feel more comfortable not putting an example in `login()`.

PR #898, opened 2017-04-13
Change HttpLoginSecurityFilter to return 401 when the client responds by sending a different scheme

stanculescu:
@dpwspoon can you please review my PR? 
Another fix I was thinking is calling the method HttpLoginSecurityFilter.loginMissingToken when the schemes are different.


transport/wsn/src/test/scripts/org/kaazing/gateway/transport/wsn/auth/shouldNotFailDueToMatchingAuthScheme.rpt
```diff
@@ -61,14 +61,15 @@ read [0x82 0x00]
 # Masked Request
 write [0x81 0xFE 0x00 0x8C 0xB1 0x34 0x9C 0x39 0xF6 0x71 0xC8 0x19 0x9E 0x51 0xFF 0x51 0xDE 0x75 0xE9 0x4D 0xD9 0x14 0xD4 0x6D 0xE5 0x64 0xB3 0x08 0x9F 0x05 0x91 0x33 0xF0 0x41 0xE8 0x51 0xDE 0x46 0xF5 0x43 0xD0 0x40 0xF5 0x56 0xDF 0x0E 0xBC 0x7B 0xD0 0x47 0xF5 0x5A 0x91 0x55 0xF1 0x00 0xDD 0x7B 0xF2 0x5D 0xDD 0x56 0xDB 0x77 0xC7 0x56 0xCB 0x6C 0x8C 0x39 0x96 0x6E 0xD4 0x56 0xCF 0x56 0xD2 0x5F 0xF9 0x4D 0x9C 0x64 0xEE 0x56 0xC5 0x5B 0xFF 0x56 0xDD 0x0E 0xBC 0x34 0xBB 0x67 0xF9 0x5A 0x9C 0x63 0xF9 0x5B 0xE2 0x5B 0xFF 0x52 0xD4 0x40 0xB1 0x7C 0xC9 0x40 0xF9 0x57 0xC2 0x5D 0xF3 0x57 0xC2 0x0E 0xBC 0x41 0x9C 0x5F 0xFD 0x58 0xCB 0x5D 0xF2 0x5E 0x9C 0x5C 0xE8 0x4D 0xC1 0x19 0xEE 0x5C 0xC7 0x55 0xF0 0x50 0xD5 0x55 0xE8 0x5C 0xBC 0x3E 0x91 0x33]
```
dpwspoon: please rename .rpt (maybe "shouldChallengeDueToMismatchingAuthSchemes.rpt" or similiar)

transport/wsn/src/test/scripts/org/kaazing/gateway/transport/wsn/auth/shouldFailDueToMismatchedAuthSchemes.rpt
```diff
@@ -63,14 +63,15 @@ read [0x82 0x00]
 # Masked Request
 write [0x81 0xFE 0x00 0x8C 0xB1 0x34 0x9C 0x39 0xF6 0x71 0xC8 0x19 0x9E 0x51 0xFF 0x51 0xDE 0x75 0xE9 0x4D 0xD9 0x14 0xD4 0x6D 0xE5 0x64 0xB3 0x08 0x9F 0x05 0x91 0x33 0xF0 0x41 0xE8 0x51 0xDE 0x46 0xF5 0x43 0xD0 0x40 0xF5 0x56 0xDF 0x0E 0xBC 0x7B 0xD0 0x47 0xF5 0x5A 0x91 0x55 0xF1 0x00 0xDD 0x7B 0xF2 0x5D 0xDD 0x56 0xDB 0x77 0xC7 0x56 0xCB 0x6C 0x8C 0x39 0x96 0x6E 0xD4 0x56 0xCF 0x56 0xD2 0x5F 0xF9 0x4D 0x9C 0x64 0xEE 0x56 0xC5 0x5B 0xFF 0x56 0xDD 0x0E 0xBC 0x34 0xBB 0x67 0xF9 0x5A 0x9C 0x63 0xF9 0x5B 0xE2 0x5B 0xFF 0x52 0xD4 0x40 0xB1 0x7C 0xC9 0x40 0xF9 0x57 0xC2 0x5D 0xF3 0x57 0xC2 0x0E 0xBC 0x41 0x9C 0x5F 0xFD 0x58 0xCB 0x5D 0xF2 0x5E 0x9C 0x5C 0xE8 0x4D 0xC1 0x19 0xEE 0x5C 0xC7 0x55 0xF0 0x50 0xD5 0x55 0xE8 0x5C 0xBC 0x3E 0x91 0x33]
 
-# 400 Response
```
dpwspoon: please rename script

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpLoginSecurityFilter.java
```diff
@@ -288,24 +288,34 @@ protected boolean login(NextFilter nextFilter,
...
-            rolesAreSufficient == false) {
+        if ((authTokenIsMissing(authToken) &&
+            rolesAreSufficient == false) ) {
             return loginMissingToken(nextFilter, session, httpRequest, authToken, additionalCallbacks, realms, realmIndex, loginContexts);
         }
 
         DefaultLoginResult loginResult = null;
+        ResultAwareLoginContext loginContext = null;
+
+        if (clientChallengeScheme != null && clientChallengeScheme.equals(expectedChallengeScheme) == false) {
```
dpwspoon: Note, I believe [RFC 7235](https://tools.ietf.org/html/rfc7235#section-2.1) states scheme is case-insensitive 

```
It uses a case-
   insensitive token as a means to identify the authentication scheme,
   followed by additional information necessary for achieving




Fielding & Reschke           Standards Track                    [Page 3]
 
RFC 7235                 HTTP/1.1 Authentication               June 2014


   authentication via that scheme.  The latter can be either a comma-
   separated list of parameters or a single sequence of characters
   capable of holding base64-encoded information.
```

so maybe equalsIgnoreCase?

Please add a k3po spec test for this.

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpSubjectSecurityFilter.java
```diff
@@ -357,32 +356,17 @@ void securityMessageReceived(NextFilter nextFilter, IoSession session, Object me
 
         DefaultAuthenticationToken authToken = (DefaultAuthenticationToken) tokenExtractor.extract(httpRequest, realm);
 
-        // If the client request provided authentication data which has
```
dpwspoon: You have correctly changed this logic to be a 401 instead of 400. 
AND you have moved this logic into the the HttpLoginSecurityFilter.login().  

The HttpLoginSecurityFilter.login() is in a scheduled task with suspends incoming data.  Yet, the logic you are checking does not require suspended reads nor a scheduled task.  So it would be better if the logic you implemented remained in this thread, rather than being scheduled.  

If there are practical reasons why it needs to be in that method that I'm not aware of, please mention them

dpwspoon: I see, by moving it you might have fixed an issue for the multiple realm case where subsequent realms don't match the challenge scheme and need a 401??

dpwspoon:
> Another fix I was thinking is calling the method HttpLoginSecurityFilter.loginMissingToken when the schemes are different.

If that works, that is ok.

@stanculescu this looks ok, to me but it is hard to tell without tests.

Please add or check that at a minimum we have tests for:

1. scheme are case insensitive 
2. 401 for pure HttpAcceptor on missing scheme (maybe we already have)
3. 401 for pure HttpAcceptor on scheme is not of correct type
4. 401 on multiple http realms where the second scheme differs 

These should all run in the http.transport module.

If "(4) 401 on multiple http realms where the second scheme differs" works before your change, consider moving the logic back out of the scheduled Thread or combining your logic with what is invoking it.  If it does not work, then consider optimizing the initial case by having generic logic somewhere where both use cases check (don't perform the check twice!).  This optimization does not need to be done, but if you can implement it in a clean way do it, though I'd rather have it implemented cleanly than optimized for a slight speed up on initial connect (1 time hit).

transport/http/src/main/java/org/kaazing/gateway/transport/http/bridge/filter/HttpLoginSecurityFilter.java
```diff
@@ -288,24 +288,34 @@ protected boolean login(NextFilter nextFilter,
...
             return loginMissingToken(nextFilter, session, httpRequest, authToken, additionalCallbacks, realms, realmIndex, loginContexts);
         }
 
         DefaultLoginResult loginResult = null;
+        ResultAwareLoginContext loginContext = null;
+
+        if (clientChallengeScheme != null && clientChallengeScheme.equalsIgnoreCase(expectedChallengeScheme) == false) {
+            String challenge = sendChallengeResponse(nextFilter, session, httpRequest, new DefaultLoginResult(), realms, realmIndex, loginContexts);
+            if (loggerEnabled()) {
+                log(String.format("Login module login failed; Issued another challenge '%s'", challenge));
```
dpwspoon: Is this the best message?  Seems to me like it is sending the challenge because the scheme no longer matches.

dpwspoon:
@stanculescu, thank you.  One nit on the comment, and then my apologies.  I wasn't clear enough in the comments above, it would be better if the tests were in the transport.http module.  This is what I was trying to say by adding tests on the pure HttpAcceptor.  This is better because changes made in transport.http (like you are doing) are tested in the same module.  Instead, you added the tests to ws transport.  Functionally, I don't see any issues with your implementations or tests (other than minor nit on the log).  But I would still like to aim at getting the tests in the correct (Better) location.  Thanks!

@danibusu will you please review after the last update (feel free to merge too!)  Thanks

stanculescu:
@dpwspoon thanks for the review, I moved the tests in transport.http as you suggested. 
The test "4. 401 on multiple http realms where the second scheme differs" didn't work before my change. As an optimization I was thinking to call a method that perform the check out of the scheduled Thread for the first realm and inside the HttpLoginSecurityFilter.login for the realms that haven't been logged yet successfully by adding a new parameter in login method signature "realmStartAt" and comparing with the current index of realm. If realmStartAt >  realmIndex do nothing, otherwise call the check method. But I don't know if this is a good solution or just makes the code harder to read.

dpwspoon:
> As an optimization ...

I don't think it is worth doing the optimization here as it adds complexity and if the strategy of going to a scheduled thread doesn't perform that well then we already have a bigger issue with the whole login process.  Additionally, if we were spending the time to optimize it, it would be better to measure before and after.  I only originally brought it up because I noticed the change, but the change makes perfect sense as we now support multiple realms.

Please don't implement it, but if we were to do it, a perfectly reasonable approach would be to have the check in both places without special casing the second check.  Meaning the scheme gets checked twice for the first realm, but the second check always succeed.  I think this is better than the index approach because the index is trying to couple code in 2 different spots that are not programmatically coupled.  Thus, it seems like the index approach is asking for trouble in refactoring as there is nothing implicitly tieing the code together. Moreover, this is a security related part of the product so we should be even less inclined to add special cases for it.  Also String.equalsIgnore case vs index !== 0 seems to be the performance change in this alternative approach.

transport/http/src/test/scripts/org/kaazing/gateway/transport/http/security/auth/shouldNotChallengeDueToMismatchingAuthSchemes.rpt
```diff
@@ -0,0 +1,45 @@
+#
+# Copyright 2007-2016, Kaazing Corporation. All rights reserved.
+#
```
danibusu: This script and `shouldChallengeDueToMismatchingAuthSchemes` are the same. Leave just one testing the full flow: no authorization header, then a header with an invalid scheme and finally a valid request.

transport/http/src/test/java/org/kaazing/gateway/transport/http/security/auth/MismatchedAuthSchemeSendsExtended401AuthTestIT.java
```diff
@@ -0,0 +1,189 @@
...
+        realms = new HttpRealmInfo[1];
+        realms[0] = new DefaultHttpRealmInfo(REALM_NAME, APPLICATION_TOKEN_CHALLENGE_SCHEME, REALM_NAME, EMPTY_STRING_ARRAY,
+                EMPTY_STRING_ARRAY, EMPTY_STRING_ARRAY, loginContextFactoryMock, null);
+
+        acceptor.getAcceptOptions().put("http.realms", realms);
+    }
+
+    @Test
+    @Specification("shouldChallengeDueToMismatchingAuthSchemes")
+    public void shouldChallengeDueToMismatchingAuthSchemes2()
```
danibusu: After merging with `shouldNotChallengeDueToMismatchingAuthSchemes2`, make sure to remove the 2 from the name.

dpwspoon:
LGTM @danibusu please merge if you are happy with the changes!  Thanks @stanculescu / @danibusu

PR #901, opened 2017-04-13
 Applied keepAliveTimeout to all layers properly.  No need for HTTPPersistanceFilter for HTTPXE when HTTP is directly over HTTP on the upper HTTP

dpwspoon:
…sistanceFilter for HTTPXE when HTTP is directly over HTTP on the upper HTTP

### DIFF
--- a/server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultAcceptOptionsContext.java
+++ b/server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultAcceptOptionsContext.java
@@ -164,7 +164,12 @@ public class DefaultAcceptOptionsContext extends DefaultOptionsContext implement
         result.put("ws[ws/draft-7x].ws[ws/draft-7x].maxMessageSize", wsMaxMessageSize);
 
         int httpKeepaliveTimeout = getHttpKeepaliveTimeout(httpKeepaliveTimeoutStr);
+
         result.put("http[http/1.1].keepAliveTimeout", httpKeepaliveTimeout);
+        result.put("http[x-kaazing-handshake].keepAliveTimeout", httpKeepaliveTimeout);
+        result.put("http[httpxe/1.1].keepAliveTimeout", httpKeepaliveTimeout);
+        result.put("http[httpxe/1.1].http[http/1.1].keepAliveTimeout", httpKeepaliveTimeout);
+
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/kaazing/gateway/server/context/resolve/DefaultAcceptOptionsContext.java
```diff
@@ -164,7 +164,12 @@ public void addBind(String scheme, String hostPort) {
         result.put("ws[ws/draft-7x].ws[ws/draft-7x].maxMessageSize", wsMaxMessageSize);
 
         int httpKeepaliveTimeout = getHttpKeepaliveTimeout(httpKeepaliveTimeoutStr);
+
         result.put("http[http/1.1].keepAliveTimeout", httpKeepaliveTimeout);
+        result.put("http[x-kaazing-handshake].keepAliveTimeout", httpKeepaliveTimeout);
+        result.put("http[httpxe/1.1].keepAliveTimeout", httpKeepaliveTimeout);
```
jitsni: If you remove the Persistence filter, then this option won't matter, right ?

transport/http/src/main/java/org/kaazing/gateway/transport/http/HttpAcceptor.java
```diff
@@ -168,7 +169,8 @@ public HttpAcceptor() {
                                                                          PROTOCOL_HTTP,
                                                                          HOST_HEADER,
                                                                          ELEVATE_EMULATED_REQUEST,
-                                                                         CONDITIONAL_WRAPPED_RESPONSE)));
+                                                                         CONDITIONAL_WRAPPED_RESPONSE,
+                                                                         PERSISTENCE)));
```
jitsni: Do you want to remove this filter in the httpxe case ?

jitsni:
Looks like the travis build is broken (don't know whether it is related)

dpwspoon:
I'm not sure which approach to take, I'm going to think it over.  Either way, I'll add a comment and that will restart travis.  I'm leaning towards keeping the filter off of HTTPXE and then commenting it out in the options, saying why it isn't needed for HTTPXE

dpwspoon:
Ping @jitsni I removed the special case of not having the persistence filter in httpxe, this is ready for review

PR #902, opened 2017-04-18
Adds message for MacOS Java version in gateway.start script

zahariaca:
Added a check for MacOS, and changed the message to reflect the different situation on MacOS


zahariaca:
No need for PR anymore.

PR #905, opened 2017-04-21
fix redirect for TLS secured URLs

danibusu:
- fix was done in WsnAcceptor
- updated expection message in DefaultHttpSession
- wrote one test (if  you think the approachis ok,, I can write more for ws)


dpwspoon:
Looks ok to me, please add the other tests and make sure wse works as well, judging by your code change http probably already works.

danibusu:
@dpwspoon Added some more tests. Let me know if I missed something. Thanks!

service/http.balancer/src/test/java/org/kaazing/gateway/service/http/balancer/SecureBalancerIT.java
```diff
@@ -0,0 +1,151 @@
...
+        // Initialize gateway truststore 
+        trustStore = KeyStore.getInstance("JKS");
+        FileInputStream tis = new FileInputStream("target/truststore/truststore.db");
+        trustStore.load(tis, null);
+        tis.close();
+    }
+
+    @Test
+    @Specification("wsx.balancer.request")
+    @ScriptProperty({"hostHeader 'localhost:8002'", "redirectAddress 'https://localhost:8003/echo'"})
```
dpwspoon, line 119: WSX and WSN are go forward strategy, they should all redirect with a location header that has "http/https".

Ideally, legacy balancer request would also redirect to "http/https" (change all wss tests to https).

Was the WS/WSS introduced here for legacy balancer bytes: https://github.com/kaazing/gateway/pull/908/files?

I suspect most client implementations don't actually care about the scheme for the balancer legacy style, but if we recently introduced the code change than I would suggest we undo it so it is http/https.  If it has always been that way then it is ok.

dpwspoon:
LGTM, thanks @danibusu, please see question

service/http.balancer/src/test/java/org/kaazing/gateway/service/http/balancer/SecureBalancerIT.java
danibusu: It has always been that way. The scripts were already written and used in the non-secure tests, I just figured I could reuse them for the secure ones. So, I just updated them to use properties, so that I could change the expected value of the location header.

dpwspoon: Thanks!

PR #909, opened 2017-04-25
Add property CLUSTER_BYPASS_AWS_CHECK in InternalSystemProperty

stanculescu:


util/src/main/java/org/kaazing/gateway/util/InternalSystemProperty.java
```diff
@@ -133,6 +133,8 @@
 
     UPDATE_CHECK_SERVICE_URL("org.kaazing.gateway.server.UPDATE_CHECK_SERVICE_URL"),
 
+    CLUSTER_BYPASS_AWS_CHECK("com.kaazing.cluster.bypass.aws.check", "false"),
```
dpwspoon: Thanks, It should be "com.kaazing.gateway" prefix so please make it "com.kaazing.gateway.cluster.bypass.aws.check"

PR #924, opened 2017-05-26
Tickets 1103 http redirect accept validation

DoruM:
Added validation in http.redirect. This is similar to http.proxy check if accept URI path ends in "/".

### DIFF
--- a/service/http.redirect/src/main/java/org/kaazing/gateway/service/http/redirect/HttpRedirectService.java
+++ b/service/http.redirect/src/main/java/org/kaazing/gateway/service/http/redirect/HttpRedirectService.java
@@ -15,6 +15,7 @@
  */
 package org.kaazing.gateway.service.http.redirect;
 
+import java.util.Collection;
 import java.util.Properties;
 
 import javax.annotation.Resource;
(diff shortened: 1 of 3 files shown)


service/http.redirect/src/main/java/org/kaazing/gateway/service/http/redirect/HttpRedirectService.java
```diff
@@ -47,8 +48,21 @@ public void setConfiguration(Properties configuration) {
 
     @Override
     public void init(ServiceContext serviceContext) throws Exception {
+        logger.info("Entering http.redirect init");
         EarlyAccessFeatures.HTTP_REDIRECT.assertEnabled(configuration, serviceContext.getLogger());
 
+        Collection<String> acceptURIs = serviceContext.getAccepts();
+        for (String acceptURI : acceptURIs) {
+            logger.info("acceptURI=" + acceptURI);
```
dpwspoon, line 56: Please do not log any of this, especially at info level: https://docs.google.com/document/d/10IYGNlvqRIfqfseQ36M4uIA8P-dgbZkQ__wbHaGiIZw/edit#heading=h.ate0rhjaq47f.  

Trace is okay if you really feel like it is needed, but 

1) The exception will already be thrown and logged, so need to put "Throwing Exception"
2) The accept URIs for all the services are already logged at startup, so no need for that.  
3) Entering http.redirect init will be in the stack trace of any exception thrown here...

